PR #4, opened 2025-01-03
cargo vet

gknopf-aranya:


gknopf-aranya:
@mooreoak, @elagergren-spideroak, @jdygert-spok the cargo vet PRs have failed to pass cicd a few times after previously vetting all the dependencies. I'm concerned about how much work it will be to keep up with vetting dependency versions and how that could affect the stability of the cicd pipeline.

jdygert-spok:
> I'm concerned about how much work it will be to keep up with vetting dependency versions and how that could affect the stability of the cicd pipeline.

To be clear, since we have the lockfile, that should only happen if someone does `cargo update` or adds new dependencies.

mooreoak:
> @mooreoak, @elagergren-spideroak, @jdygert-spok the cargo vet PRs have failed to pass cicd a few times after previously vetting all the dependencies. I'm concerned about how much work it will be to keep up with vetting dependency versions and how that could affect the stability of the cicd pipeline.

I think it indicates that not properly managing our dependencies if it is too much work. I am open to be wrong here but it seems like it *must* be manageable if we are to have trust in our supply chain.

gknopf-aranya:
Now that I think about it, this has been happening more frequently on this PR because of merging in changes that were pushed from the closed-source repositories. Going forward, each new PR will be responsible for vetting the dependencies it introduces so the work/responsibility will be spread out a bit more across the development team.

mooreoak (approved):

apetkov-so (approved):

nikki-aranya (approved):
My two cents: looking at our dependency tree and the commit history on this, I don’t think there’s any avoiding some of the churn. libc and wasm are going to be updated a lot due to API nonsense, tokio is used by a lot of the ecosystem and gets constant attention, thiserror and anyhow are extremely useful (even if I prefer snafu).

A bunch of this (e.g. itertools) is only included due to dev-dependencies like criterion. Some obvious wins I can see:
* we can ditch clap for something like argp which not only reduces dependencies but has a measurable impact on binary size
* we can use the `#[bench]` attribute instead of criterion which requires nightly but is already relegated to dev-only so we’ll already have it installed, plus it removes a massive number of dependencies
* we don’t *have* to use tokio, especially on embedded devices where there’s barely more than one core and we can’t use its full potential anyways, we can use a slimmer runtime like smol, rust’s built-in async capabilities, or just strip it out entirely.

We can and should rely on those third party audits, they’re closer to the scale of the crates we currently use and cover a lot of the obvious ground, but there’s a lot we can do to be more mindful about our dependencies and anything we can do to reduce our overall attack surface is a good thing. Us using "popular" crates means they’re more likely to be targetted in any kind of attack specifically due to them being used by a lot of applications.

apetkov-so (approved):

mooreoak (approved):

jdygert-spok (approved):

PR #5, opened 2025-01-03
feat(crypto)!: make engine and rng shared

jdygert-spok:


ericlagergren:
I haven't thought about it much or looked at the PR in much depth, but I think it lgtm.

crates/aranya-crypto/src/bearssl.rs
```diff
@@ -1056,7 +1056,7 @@ hmac_impl!(HmacSha384, "HMAC-SHA384", Sha384);
 hmac_impl!(HmacSha512, "HMAC-SHA512", Sha512);
 
 /// A `HMAC_DRBG`-based CSPRNG.
-pub struct HmacDrbg(br_hmac_drbg_context);
+pub struct HmacDrbg(spin::Mutex<br_hmac_drbg_context>);
```
jdygert-spok, line 1059: Had to add a mutex here.

elagergren-spideroak:
I haven't thought about it much or looked at the PR in much depth, but I think it lgtm.

elagergren-spideroak:
fyi https://github.com/aranya-project/aranya-core/pull/34

elagergren-spideroak:
> > fyi #34
> 
> Yeah that'll be fun to resolve :)
> 
> Any thoughts on having mutexes on some rngs?

It's fine. We only need a mutex for user space CSPRNGs, which we only use when absolutely forced to (FIPS or the `trng` feature). User space CSPRNGs should hopefully be fast enough to not suffer (much) contention. And in a multi-threaded environment we should be using thread-local CSPRNGs anyway.

elagergren-spideroak:
https://github.com/aranya-project/aranya-core/pull/47

Issue #15, opened 2025-01-06
Publish multiple commands from action


apetkov-so:
This is needed to enable `map`. E.g.:

```
action do_something() {
    map F[] as f {
        // Each iteration will publish a new `DoSomething` command.
        publish DoSomething { f: f }
    }
}
```

The VM currently exists after processing a command, preventing subsequent commands from being published.

PR #16, opened 2025-01-07
#15: Publish multiple commands

apetkov-so:
- The VM now `Yield`s to the  runtime when it publishes a command. The runtime is responsible for sealing the command, updating the facts database, then resuming the VM so it can process the next command.
- FFI `call`s now take an immutable self.


apetkov-so:
Add test that publishes multiple types of commands (per @gknopf-aranya)
```
map F[] as f {
  publish A {}
  publish B {}
}
```

crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -562,27 +515,123 @@ impl<E: aranya_crypto::Engine> Policy for VmPolicy<E> {
...
+                                error!("{e}");
+                                EngineError::Write
+                            })?;
+
+                        // Resume action after last Publish
+                        exit_reason = rs.run().map_err(|e| {
+                            error!("{e}");
+                            EngineError::InternalError
+                        })?;
+                    }
```
ssweetney-spok, line 634: **Copying my comment from the other PR:**

The runstate's context should be updated with the head_id of the newly added command before resuming so that code like:

```
action publish_chain_of_commands() {
     publish Foo {}
     perspective::head_id() // Should return the command ID of the Foo command that was just published
     publish Bar {}
     perspective::head_id() // Should return the command ID of the Bar command that was just published
}
```
works as expected.

In hindsight, I should've added a test for this...

apetkov-so: Aren't we already sort-of testing for this in `should_create_client_with_ffi_and_publish_chain_of_commands`?
```
    let mut expected_parent_id = eff1.command.into_id();

    for eff in [eff2, eff3] {
        // Observe that the actual 'parent_id' of the command that created this effect
        // is equal to the expected 'parent_id'
        let actual_parent_id = retrieve_id("parent_id", eff)?;
        assert_eq!(expected_parent_id, actual_parent_id);

        // Update the expected 'parent_id' with the 'command_id' of the command that created the
        // current effect so that it can be used in the next loop iteration
        let current_command_id = retrieve_id("command_id", eff)?;
        expected_parent_id = current_command_id;
    }
```

apetkov-so: Ah, so the _policy_ would get the wrong command id, not the runtime.

apetkov-so: Pushed commit to update the RS context with the new command id.

ssweetney-spok (dismissed):
Looks good!

ssweetney-spok (dismissed):

crates/aranya-runtime/src/vm_policy/io.rs
```diff
@@ -42,11 +47,11 @@ where
 
 /// Implements the `MachineIO` interface for [VmPolicy](super::VmPolicy).
 pub struct VmPolicyIO<'o, P, S, E, FFI> {
-    facts: &'o mut P,
-    sink: &'o mut S,
+    facts: Rc<RefCell<&'o mut P>>,
+    sink: Rc<RefCell<&'o mut S>>,
```
jdygert-spok, line 51:
```suggestion
    facts: &'o RefCell<&'o mut P>,
    sink: &'o RefCell<&'o mut S>,
```

crates/aranya-afc-util/src/ffi.rs
```diff
@@ -30,13 +31,15 @@ macro_rules! error {
 /// An [`FfiModule`][aranya_policy_vm::ffi::FfiModule] for AFC.
 #[derive(Clone)]
 pub struct Ffi<S> {
-    store: S,
+    store: RefCell<S>,
```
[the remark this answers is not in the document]
apetkov-so: `spin::Mutex` doesn't implement `Clone`. We can drop the `Clone` trait from `struct Ffi`, if that's acceptable.

crates/aranya-afc-util/src/ffi.rs
```diff
@@ -92,7 +93,8 @@ function create_bidi_channel(
 
         let our_sk = &self
             .store
-            .safe_borrow()?
+            .try_lock()
```
jdygert-spok, line 96: You want `lock`, `try_lock` in this case means it won't block if the mutex is already locked. Spin mutexes don't do the poison panic thing that std mutexes do.

crates/aranya-afc-util/src/ffi.rs
```diff
@@ -6,12 +6,12 @@
...
 
 use alloc::vec::Vec;
-use core::{cell::RefCell, result::Result};
+use core::result::Result;
 
 use aranya_crypto::{
     self,
     afc::{BidiChannel, BidiSecrets, UniChannel, UniSecrets},
-    aranya_buggy::{Bug, BugExt, SafeBorrow},
+    buggy::Bug,
```
jdygert-spok, line 14: Import separately

crates/aranya-afc-util/src/ffi.rs
```diff
@@ -28,15 +30,17 @@ macro_rules! error {
 }
 
 /// An [`FfiModule`][aranya_policy_vm::ffi::FfiModule] for AFC.
-#[derive(Clone)]
+// #[derive(Clone)]
```
jdygert-spok, line 33:
```suggestion
```

jdygert-spok (approved):

ssweetney-spok (dismissed):

crates/aranya-policy-vm/src/machine.rs
```diff
@@ -795,34 +841,29 @@ where
...
             }
             Instruction::Serialize => {
+                let command_struct: Struct = self.ipop()?;
                 let CommandContext::Seal(SealContext { name, .. }) = self.ctx else {
-                    return Err(MachineError::from_position(
-                        MachineErrorType::InvalidInstruction,
-                        self.pc,
-                        self.machine.codemap.as_ref(),
-                    ));
+                    return Err(self.err(MachineErrorType::BadState("Serialize: no seal context")));
```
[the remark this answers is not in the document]
apetkov-so: So `BadState` is a valid error type, we just want a different message? How about "Expected seal context"?

jdygert-spok (approved):

crates/aranya-model/src/tests/basic-policy.md
```diff
@@ -288,4 +288,33 @@ command StoreSessionData {
...
+
+    policy {
+        finish {
+            emit Relationship{parent_id: envelope.parent_id, command_id: envelope.command_id}
+        }
+    }
+}
+
+// Publishes multiple `Link` commands
+action publish_multiple_commands() {
```
gknopf-aranya, line 315: Does publish multiple commands require all the commands to be the same, or could they be different commands?

crates/aranya-policy-vm/tests/bits/ffi_envelope.rs
```diff
@@ -0,0 +1,77 @@
...
+        payload: Vec<u8>,
+    ) -> Result<Envelope, MachineError> {
+        let CommandContext::Seal(ctx) = ctx else {
+            bug!("envelope::seal called outside seal context");
+        };
+
+        let parent_id: Id = ctx.head_id;
+        let author_id = self.user;
+
+        // let data = postcard::to_allocvec(&HashedFields {
```
gknopf-aranya, line 49: Uncomment or remove

crates/aranya-policy-vm/tests/bits/ffi_envelope.rs
```diff
@@ -0,0 +1,77 @@
...
+
+        let command_id = Id::default(); // CommandId::hash_for_testing_only(&data);
+
+        Ok(Envelope {
+            parent_id,
+            author_id: author_id.into(),
+            command_id,
+            payload,
+            // TODO(chip): use an actual signature
+            signature: b"LOL".to_vec(),
```
gknopf-aranya, line 64: Is this todo needed on this PR?

crates/aranya-runtime/src/vm_policy/io.rs
```diff
@@ -133,26 +149,29 @@ where
...
     ) {
         let fields: Vec<_> = fields.into_iter().collect();
-        self.sink.consume(VmEffect {
-            name,
-            fields,
-            command: command.into(),
-            recalled,
-        });
+        self.sink
+            .borrow_mut() // TODO use non-panicking borrow
```
gknopf-aranya, line 153: @jdygert-spok is there a way return an error instead of panicking here?

crates/aranya-model/src/tests/basic-policy.md
apetkov-so: They can be different.

crates/aranya-runtime/src/vm_policy/io.rs
apetkov-so: I'll add an `assume()`. It's not expected to fail, and the method has no return value anyway.

crates/aranya-policy-vm/tests/bits/ffi_envelope.rs
jdygert-spok: I think this entire file is unused now after a change of approach.

crates/aranya-policy-vm/tests/bits/ffi_envelope.rs
apetkov-so: Actually, the whole file can be deleted - it was a copy. The `publish_multiple_command` test lives in `aranya-model` which uses the original `runtime::testing::TestFfiEnvelope`.

crates/aranya-policy-vm/tests/bits/ffi_envelope.rs
```diff
@@ -0,0 +1,77 @@
+extern crate alloc;
+
+use alloc::vec::Vec;
+use core::convert::Infallible;
+
+use aranya_crypto::{Id, UserId};
+use aranya_policy_vm::{ffi::ffi, CommandContext, MachineError};
+use buggy::bug;
+
+// use crate::CommandId;
```
smckeon-so: Similar with this, if we're not using it we can remove it.

smckeon-so (commented):
Looks great, just a couple of comments/questions.

crates/aranya-model/src/tests/mod.rs
```diff
@@ -1414,8 +1414,10 @@ fn should_create_clients_with_args() {
     assert_eq!(effects, expected);
 }
 
+/// If there are no facts when writing a fact perspective, we skip writing it and just return its prior.
+/// When starting a perspective from the middle of a segment with no facts, we previously would grab the wrong fact index, by taking that prior's prior to apply 0 fact updates to it.
 #[test]
-fn test_storage_fact() {
+fn test_storage_fact_index() {
```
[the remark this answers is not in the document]
apetkov-so: Ok

crates/aranya-policy-vm/tests/bits/ffi_envelope.rs
apetkov-so: This file has been deleted.

jdygert-spok (approved):

crates/aranya-model/src/tests/basic-policy.md
gknopf-aranya: @apetkov-so, @jdygert-spok, @chip-so do you think using different commands would improve test coverage? I don't have enough understanding of the implementation to know if it would be beneficial.

crates/aranya-policy-vm/tests/bits/ffi_envelope.rs
gknopf-aranya: Once the unused file is removed, I can approve the PR

smckeon-so (approved):
🦀

gknopf-aranya (approved):

PR #41, opened 2025-01-07
release-plz

gknopf-aranya:
Automates crates.io publishing and release PR creation.

Quickstart:
https://release-plz.dev/docs/github/quickstart


jdygert-spok (dismissed):

djohnson-spideroak (approved):

jdygert-spok (dismissed):

.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,56 @@
...
+name: Release-plz
+
+permissions:
+  pull-requests: write
+  contents: write
+
+on:
+  push:
+    branches:
+      - main
```
elagergren-spideroak, line 15: ok, so whenever we push to main this workflow will publish all crates that have new version numbers?

elagergren-spideroak: eg. if we update `foo` and change its version from v0.1.0 to v0.2.0 this action will publish it. If we update `bar` but do not change its version, this action will not publish it.

gknopf-aranya: Yes, that's the expected behavior as I understand it

nikki-aranya (approved):
I don't see anything obvious, so LGTM

release-plz.toml
```diff
@@ -0,0 +1,2 @@
+[workspace]
+changelog_update = false
```
apetkov-so, line 2: Why don't we want to update changelogs?

gknopf-aranya: `release-plz` creates a changelog for each crate. But we only want to use a single changelog for GitHub releases. We're setting all the crates to the same version, so it doesn't make sense to track versions and changelogs for each crate.

nikki-aranya (approved):

release-plz.toml
elagergren-spideroak: > We're setting all the crates to the same version, so it doesn't make sense to track versions and changelogs for each crate.

Except we're not doing that for `aranya-core`, so it does make sense to do that.

jdygert-spok: My original thought was that a changelog in the github release for each crate would be sufficient and we didn't need a `CHANGELOG.md` in each crate. But thinking some more, not really sure what we want. Is a GH release per crate excessive? Also, I don't think we can have just the GH release changelog without the file changelog via release-plz.

elagergren-spideroak: I don't think a GH release per crate release is excessive. Worst case is we try it, figure out that it's excessive, and then do something else.

elagergren-spideroak (approved):

gknopf-aranya:
@jdygert-spok, @nikki-aranya could you please take another look?

smckeon-so (approved):

PR #44, opened 2025-01-07
crypto: get rid of ThreadRng

### LINKED ISSUE #43: crypto: get rid of `rand::ThreadRng` usage
At compile time, `Rng` selects one of:

- A thread-local (ish) user space fast key erasure ChaCha8 CSPRNG if `trng` is enabled, or
- [ThreadRng] if `std` is enabled, or
- [getrandom] if `getrandom` is enabled, or
- `extern "C" fn crypto_getrandom(...)` otherwise

I'm not exactly a fan of user space CSPRNGs. The ChaCha8 `trng` CSPRNG is unavoidable, but [ThreadRng] isn't. We should probably get rid of it.

[ThreadRng]: https://docs.rs/rand/latest/rand/rngs/struct.ThreadRng.html
[getrandom]: https://docs.rs/getrandom/latest/getrandom/

elagergren-spideroak:
Fixes #43

### DIFF
--- a/crates/aranya-crypto-core/Cargo.toml
+++ b/crates/aranya-crypto-core/Cargo.toml
@@ -16,7 +16,6 @@ workspace = true
 [features]
 default = [
 	"getrandom",
-	#"trng",
 ]
 
 # Enable allocations.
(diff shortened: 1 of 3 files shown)


ycarmy (dismissed):

jdygert-spok (approved):

jdygert-spok (approved):

jdygert-spok:
Should be done on https://github.com/aranya-project/crypto now.

PR #45, opened 2025-01-07
crypto: use ChaCha8 instead of AES-CTR for `trng` CSPRNG

### LINKED ISSUE #40: crypto: use ChaCha8 instead of AES-CTR for the `trng` CSPRNG
Instead of DIYing a CSPRNG construction, we've been choosing to follow what Go does.

See https://github.com/golang/go/commit/f705cf8f96bef3f61ba65e62b0b9123d718d0d12

elagergren-spideroak:
Fixes #40


ycarmy (approved):
lgtm!

PR #68, opened 2025-01-21
Migrate Errors to `thiserror`

nikki-aranya:
Migrates the majority of errors to `thiserror`, save for the places where it doesn't make sense (mostly more complex ErrorKind-type errors and macros). Solves #71.


nikki-aranya:
The main thing from this initial commit is not changing any functionality, enforcing lowercase strings since that seems to be the overwhelming majority, and prefixing the derive trait as `thiserror::Error` and sticking it at the very end.

Open questions remain: 
- [ ] In [aead.rs](https://github.com/aranya-project/aranya-core/pull/68/files#diff-b0ef691e87fc5a4c0abc4354f1f33949497cb758fb1967cf7ac47274b9d9db31R110-R125) and [safe.rs](https://github.com/aranya-project/aranya-core/pull/68/files#diff-94f43898a265f9c33c559796b291b995aeb2dcf82ee63ad24333bbab54172becR1012-R1020), we use as_str. Do we want to enforce that for all errors, do we want to remove those couple 3 instances I found? 
- [ ] In my notes, I have that `Error(Repr)`] in [fs_keystore](https://github.com/aranya-project/aranya-core/blob/74832552cad4404a9aaea805eca0bc8fd4a5a8a9/crates/aranya-crypto/src/keystore/fs_keystore/error.rs#L23) and [AFC buf.rs](hhttps://github.com/aranya-project/aranya-core/blob/74832552cad4404a9aaea805eca0bc8fd4a5a8a9/crates/aranya-fast-channels/src/buf.rs#L16) could probably be refactored into a more normal Error type (or ErrorKind).
- [ ] This is maybe more of a thing for the MVP spec, but seeing all these enums, do we want `#[non_exhaustive]` to not have to commit to API guarantees? 
- [ ] In [aranya-policy-vm](https://github.com/aranya-project/aranya-core/pull/68/files#diff-17af57b7a23fd569614391b4c43397fb9bd8b4ed683f4a2ebab579c98c21dc5fR148-R155), do we want to rename source to info and err_type to source? `MachineError` is more of a wrapper providing context than anything else, `MachineErrorKind` could easily be its own standalone error.
- [ ] In [aranya-policy-module](https://github.com/aranya-project/aranya-core/blob/74832552cad4404a9aaea805eca0bc8fd4a5a8a9/crates/aranya-policy-module/src/data.rs#L17C1-L24C7) and [aranya-policy-vm](https://github.com/aranya-project/aranya-core/pull/68/files#diff-17af57b7a23fd569614391b4c43397fb9bd8b4ed683f4a2ebab579c98c21dc5fR34-R41), do we want to refactor these want/got errors with actual type enums? I asked Chip and he brought up [VType](https://github.com/aranya-project/aranya-core/blob/74832552cad4404a9aaea805eca0bc8fd4a5a8a9/crates/aranya-policy-ast/src/ast.rs#L75C1-L96C2) as a possible solution, but we should look into it further.
- [ ] We have `Errno` wrappers in a handful of places and it seems like they could probably be unified into a single type. Do we want to do that? 
- [ ] We have alllll these types (see the commit), do we want to refactor and unify these into a more broadened Error set, possibly in its own crate in aranya-core?

elagergren-spideroak:
@nikki-aranya No need to update `aranya-crypto-core`. The code is being moved, see https://github.com/aranya-project/crypto/pull/1

jdygert-spok:
> In [aead.rs](https://github.com/aranya-project/aranya-core/pull/68/files#diff-b0ef691e87fc5a4c0abc4354f1f33949497cb758fb1967cf7ac47274b9d9db31R110-R125) and [safe.rs](https://github.com/aranya-project/aranya-core/pull/68/files#diff-94f43898a265f9c33c559796b291b995aeb2dcf82ee63ad24333bbab54172becR1012-R1020), we use as_str. Do we want to enforce that for all errors, do we want to remove those couple 3 instances I found?

AFAIK this was done for APS (now fast channel)'s standalone C API which doesn't exist anymore (for now). I'm in no hurry to remove it. We did switch to a different way with aranya's C API so we might not need it, but it is pretty convenient to have the static descriptions.

> In my notes, I have that `Error(Repr)` in [fs_keystore](https://github.com/aranya-project/aranya-core/blob/74832552cad4404a9aaea805eca0bc8fd4a5a8a9/crates/aranya-crypto/src/keystore/fs_keystore/error.rs#L23) and [AFC buf.rs](https://github.com/aranya-project/aranya-core/blob/74832552cad4404a9aaea805eca0bc8fd4a5a8a9/crates/aranya-fast-channels/src/buf.rs#L16) could probably be refactored into a more normal Error type (or ErrorKind).

Looks like the buf one might be intended to keep the repr private. For the keystore there is some context of having a way to do no-alloc custom errors in traits and legacy of pre-error_in_core, but looking at it again it does seem kinda funky.

> This is maybe more of a thing for the MVP spec, but seeing all these enums, do we want `#[non_exhaustive]` to not have to commit to API guarantees?

I've been pondering this, I think so.

> In [aranya-policy-vm](https://github.com/aranya-project/aranya-core/pull/68/files#diff-17af57b7a23fd569614391b4c43397fb9bd8b4ed683f4a2ebab579c98c21dc5fR148-R155), do we want to rename source to info and err_type to source? `MachineError` is more of a wrapper providing context than anything else, `MachineErrorKind` could easily be its own standalone error.

I have been considering renaming them to something like `MachineError` and `LocatedMachineError`, and changing the "internal" methods to return the unlocated one and just add the location at the public methods. Same thing for the compiler. Those would clean up the error handling a lot I think.

> In [aranya-policy-module](https://github.com/aranya-project/aranya-core/blob/74832552cad4404a9aaea805eca0bc8fd4a5a8a9/crates/aranya-policy-module/src/data.rs#L17C1-L24C7) and [aranya-policy-vm](https://github.com/aranya-project/aranya-core/pull/68/files#diff-17af57b7a23fd569614391b4c43397fb9bd8b4ed683f4a2ebab579c98c21dc5fR34-R41), do we want to refactor these want/got errors with actual type enums? I asked Chip and he brought up [VType](https://github.com/aranya-project/aranya-core/blob/74832552cad4404a9aaea805eca0bc8fd4a5a8a9/crates/aranya-policy-ast/src/ast.rs#L75C1-L96C2) as a possible solution, but we should look into it further.

IIRC there were some cases that meant it would need a separate enum type, so I didn't want to block the improvement on adding that. But it would be nice to have.

> We have `Errno` wrappers in a handful of places and it seems like they could probably be unified into a single type. Do we want to do that?

Yeah, it'd be nice to pull some more things into `aranya-libc` and organize it a bit more. And/or figure out if we can make rustix more broadly available or find a replacement. I think VxWorks was the reason we couldn't use it in some places? But other future platforms could be an issue too.

> We have alllll these types (see the initial commit), do we want to refactor and unify these into a more broadened Error set, possibly in its own crate in aranya-core?

Thinking no, but I'd be open to hear more on what you envision.

>  We use `anyhow` in a handful of test files. This is probably more for a test-focused PR but I've brought up in the past that we should use `#[bench]` for any benchmarking/tests so we can remove criterion and all the dependencies it pulls in that we have to audit, as well as anyhow would be really nice.

`anyhow` is so minimal anyway, and is used in `aranya` and other integrations. I don't see a reason to remove it from tests/bins. It's very convenient.

crates/aranya-fast-channels/src/error.rs
```diff
@@ -12,106 +12,65 @@ use crate::{buf::AllocError, errno::Errno, header::HeaderError, state::ChannelId
...
-
-#[cfg(any(feature = "sdlib", feature = "posix"))]
-impl From<crate::shm::Error> for Error {
-    fn from(value: crate::shm::Error) -> Self {
-        Self::SharedMem(value)
-    }
-}
-
 #[cfg(any(feature = "sdlib", feature = "posix"))]
 impl From<crate::shm::Corrupted> for Error {
```
jdygert-spok, line 75: rm

crates/aranya-crypto-core/src/hpke.rs
```diff
@@ -303,97 +296,33 @@ impl Display for AeadId {
...
+    #[error(transparent)]
+    Import(#[from] ImportError),
     /// A key could not be exported.
-    Export(ExportError),
+    #[error(transparent)]
+    Export(#[from] ExportError),
     /// The encryption context has been used to send the maximum
     /// number of messages.
+    #[error("message limit reached")]
     MessageLimitReached,
```
jdygert-spok, line 322: Might be nice to put the unit struct in this variant.

```rust
MessageLimitReached(#[from] MessageLimitReached),
```

crates/aranya-crypto-core/src/import.rs
```diff
@@ -1,94 +1,44 @@
...
 
 use aranya_buggy::Bug;
 use generic_array::{ArrayLength, GenericArray};
 
 use crate::signer::PkError;
 
 /// A slice could not be converted to a fixed-size buffer.
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Debug, Eq, PartialEq, thiserror::Error)]
+#[error("invalid size data: got {got}, want {start}..{end}", start = self.want.start, end = self.want.end)]
```
jdygert-spok, line 12: `{want:?}` will do here

elagergren-spideroak:
> @nikki-aranya No need to update `aranya-crypto-core`. The code is being moved, see [aranya-project/crypto#1](https://github.com/aranya-project/crypto/pull/1)

Also, the other thing about `aranya-crypto-core` is that I have a couple open issues to flatten out and simplify some errors. Eg, going from `enum SealError { ... }` to `struct SealError(())`.

elagergren-spideroak:
> > @nikki-aranya No need to update `aranya-crypto-core`. The code is being moved, see [aranya-project/crypto#1](https://github.com/aranya-project/crypto/pull/1)
> 
> Also, the other thing about `aranya-crypto-core` is that I have a couple open issues to flatten out and simplify some errors. Eg, going from `enum SealError { ... }` to `struct SealError(())`.

To be clear, this only applies to `aranya-crypto-core`. `aranya-crypto` is fair game.

elagergren-spideroak:
> > In [aead.rs](https://github.com/aranya-project/aranya-core/pull/68/files#diff-b0ef691e87fc5a4c0abc4354f1f33949497cb758fb1967cf7ac47274b9d9db31R110-R125) and [safe.rs](https://github.com/aranya-project/aranya-core/pull/68/files#diff-94f43898a265f9c33c559796b291b995aeb2dcf82ee63ad24333bbab54172becR1012-R1020), we use as_str. Do we want to enforce that for all errors, do we want to remove those couple 3 instances I found?
> 
> AFAIK this was done for APS (now fast channel)'s standalone C API which doesn't exist anymore (for now). I'm in no hurry to remove it. We did switch to a different way with aranya's C API so we might not need it, but it is pretty convenient to have the static descriptions.

Yes, it was primarily for AFC (née APS). I /think/ there was another reason that necessitated a static string, but I don't remember exactly what it is. I don't care whether we keep or remove the `as_str` methods. But we should probably do that on a separate PR.

> > In my notes, I have that `Error(Repr)` in [fs_keystore](https://github.com/aranya-project/aranya-core/blob/74832552cad4404a9aaea805eca0bc8fd4a5a8a9/crates/aranya-crypto/src/keystore/fs_keystore/error.rs#L23) and [AFC buf.rs](https://github.com/aranya-project/aranya-core/blob/74832552cad4404a9aaea805eca0bc8fd4a5a8a9/crates/aranya-fast-channels/src/buf.rs#L16) could probably be refactored into a more normal Error type (or ErrorKind).
> 
> Looks like the buf one might be intended to keep the repr private. For the keystore there is some context of having a way to do no-alloc custom errors in traits and legacy of pre-error_in_core, but looking at it again it does seem kinda funky.

Pretty much, yeah.

`AllocError(Repr)` is just to keep `Repr` private. It's really just the same thing as `AllocError` or `AllocError(())`, but knowing the underlying error can be useful and it's ~cheap.

`keystore` was me experimenting with an error type similar to `io::Error`, but no-std (and no-alloc?). I've been trying to find a nice pattern for errors that's not `enum FooError { ... }` (since that gets unwieldily very quickly and leaks a lot of implementation details), but is also no-std and no-alloc.

> > This is maybe more of a thing for the MVP spec, but seeing all these enums, do we want `#[non_exhaustive]` to not have to commit to API guarantees?
> 
> I've been pondering this, I think so.

Agreed.

> > In [aranya-policy-vm](https://github.com/aranya-project/aranya-core/pull/68/files#diff-17af57b7a23fd569614391b4c43397fb9bd8b4ed683f4a2ebab579c98c21dc5fR148-R155), do we want to rename source to info and err_type to source? `MachineError` is more of a wrapper providing context than anything else, `MachineErrorKind` could easily be its own standalone error.
> 
> I have been considering renaming them to something like `MachineError` and `LocatedMachineError`, and changing the "internal" methods to return the unlocated one and just add the location at the public methods. Same thing for the compiler. Those would clean up the error handling a lot I think.

Yes, they should be renamed, but on a separate PR.

> > In [aranya-policy-module](https://github.com/aranya-project/aranya-core/blob/74832552cad4404a9aaea805eca0bc8fd4a5a8a9/crates/aranya-policy-module/src/data.rs#L17C1-L24C7) and [aranya-policy-vm](https://github.com/aranya-project/aranya-core/pull/68/files#diff-17af57b7a23fd569614391b4c43397fb9bd8b4ed683f4a2ebab579c98c21dc5fR34-R41), do we want to refactor these want/got errors with actual type enums? I asked Chip and he brought up [VType](https://github.com/aranya-project/aranya-core/blob/74832552cad4404a9aaea805eca0bc8fd4a5a8a9/crates/aranya-policy-ast/src/ast.rs#L75C1-L96C2) as a possible solution, but we should look into it further.
> 
> IIRC there were some cases that meant it would need a separate enum type, so I didn't want to block the improvement on adding that. But it would be nice to have.

Also, as mentioned on the PR that added that variant, we should try and make it no-alloc, or at least `Cow` or something. It's tricky, but probably doable. Allocating in errors isn't a big deal, but if it's feasible then we generally try to make our code no-alloc.

> > We have `Errno` wrappers in a handful of places and it seems like they could probably be unified into a single type. Do we want to do that?
> 
> Yeah, it'd be nice to pull some more things into `aranya-libc` and organize it a bit more. And/or figure out if we can make rustix more broadly available or find a replacement. I think VxWorks was the reason we couldn't use it in some places? But other future platforms could be an issue too.

Yeah, Rustix didn't support VxWorks and we didn't want to keep two nearly identical implementations of `IoManager` (`libc` vs `rustix`) around in `aranya_runtime::storage::linear`, so we dropped Rustix. I then created `aranya-libc` and consolidated the `libc` code there because I needed it for another crate (the C API?).

I'm reticent to start using Rustix again because I'm really not convinced that we're not going to get more oddball platforms that Rustix doesn't support. At the same time, I don't exactly want to maintain our own homebrew version of Rustix. We could probably have `aranya-libc` delegate to Rustix on supported platforms. I don't know if that's more or less work, though.

We should definitely consolidate libc/errno/whatever stuff inside `aranya-libc`, though. I've written like four different errno impls now and it's a bit silly.

> > We have alllll these types (see the initial commit), do we want to refactor and unify these into a more broadened Error set, possibly in its own crate in aranya-core?
> 
> Thinking no, but I'd be open to hear more on what you envision.

Ditto.

> > We use `anyhow` in a handful of test files. This is probably more for a test-focused PR but I've brought up in the past that we should use `#[bench]` for any benchmarking/tests so we can remove criterion and all the dependencies it pulls in that we have to audit, as well as anyhow would be really nice.
> 
> `anyhow` is so minimal anyway, and is used in `aranya` and other integrations. I don't see a reason to remove it from tests/bins. It's very convenient.

I would love to get rid of Criterion, but I think built-in benchmarking is still unstable. Running benchmarks with `+nightly` isn't a /huge/ deal, maybe. If I could just get proper nested benchmark (and testing, tbh) support ala https://pkg.go.dev/testing I would be so happy.

nikki-aranya:
> > > In my notes, I have that `Error(Repr)` in [fs_keystore](https://github.com/aranya-project/aranya-core/blob/74832552cad4404a9aaea805eca0bc8fd4a5a8a9/crates/aranya-crypto/src/keystore/fs_keystore/error.rs#L23) and [AFC buf.rs](https://github.com/aranya-project/aranya-core/blob/74832552cad4404a9aaea805eca0bc8fd4a5a8a9/crates/aranya-fast-channels/src/buf.rs#L16) could probably be refactored into a more normal Error type (or ErrorKind).
> > 
> > 
> > Looks like the buf one might be intended to keep the repr private. For the keystore there is some context of having a way to do no-alloc custom errors in traits and legacy of pre-error_in_core, but looking at it again it does seem kinda funky.
> 
> Pretty much, yeah.
> 
> `AllocError(Repr)` is just to keep `Repr` private. It's really just the same thing as `AllocError` or `AllocError(())`, but knowing the underlying error can be useful and it's ~cheap.
> 
> `keystore` was me experimenting with an error type similar to `io::Error`, but no-std (and no-alloc?). I've been trying to find a nice pattern for errors that's not `enum FooError { ... }` (since that gets unwieldily very quickly and leaks a lot of implementation details), but is also no-std and no-alloc.

That makes a lot more sense. Seems like the kind of thing we'd put in a crate ala buggy, if we ever find a better pattern. For now I'll disregard that whole variant.

> > > In [aranya-policy-module](https://github.com/aranya-project/aranya-core/blob/74832552cad4404a9aaea805eca0bc8fd4a5a8a9/crates/aranya-policy-module/src/data.rs#L17C1-L24C7) and [aranya-policy-vm](https://github.com/aranya-project/aranya-core/pull/68/files#diff-17af57b7a23fd569614391b4c43397fb9bd8b4ed683f4a2ebab579c98c21dc5fR34-R41), do we want to refactor these want/got errors with actual type enums? I asked Chip and he brought up [VType](https://github.com/aranya-project/aranya-core/blob/74832552cad4404a9aaea805eca0bc8fd4a5a8a9/crates/aranya-policy-ast/src/ast.rs#L75C1-L96C2) as a possible solution, but we should look into it further.
> > 
> > 
> > IIRC there were some cases that meant it would need a separate enum type, so I didn't want to block the improvement on adding that. But it would be nice to have.
> 
> Also, as mentioned on the PR that added that variant, we should try and make it no-alloc, or at least `Cow` or something. It's tricky, but probably doable. Allocating in errors isn't a big deal, but if it's feasible then we generally try to make our code no-alloc.

Yeah, no_alloc is always a pain, and you're right, I didn't think about that this is just an error so it's less critical that we migrate off a &'static str.

> > > We have `Errno` wrappers in a handful of places and it seems like they could probably be unified into a single type. Do we want to do that?
> > 
> > 
> > Yeah, it'd be nice to pull some more things into `aranya-libc` and organize it a bit more. And/or figure out if we can make rustix more broadly available or find a replacement. I think VxWorks was the reason we couldn't use it in some places? But other future platforms could be an issue too.
> 
> Yeah, Rustix didn't support VxWorks and we didn't want to keep two nearly identical implementations of `IoManager` (`libc` vs `rustix`) around in `aranya_runtime::storage::linear`, so we dropped Rustix. I then created `aranya-libc` and consolidated the `libc` code there because I needed it for another crate (the C API?).
> 
> I'm reticent to start using Rustix again because I'm really not convinced that we're not going to get more oddball platforms that Rustix doesn't support. At the same time, I don't exactly want to maintain our own homebrew version of Rustix. We could probably have `aranya-libc` delegate to Rustix on supported platforms. I don't know if that's more or less work, though.
> 
> We should definitely consolidate libc/errno/whatever stuff inside `aranya-libc`, though. I've written like four different errno impls now and it's a bit silly.

Y'all would know that better than I do about any FFI-adjacent stuff, especially on more obscure platforms.

> > > We have alllll these types (see the initial commit), do we want to refactor and unify these into a more broadened Error set, possibly in its own crate in aranya-core?
> > 
> > 
> > Thinking no, but I'd be open to hear more on what you envision.
> 
> Ditto.

I didn't really have any plans there, I just wanted to throw it out as food for thought.

> > > We use `anyhow` in a handful of test files. This is probably more for a test-focused PR but I've brought up in the past that we should use `#[bench]` for any benchmarking/tests so we can remove criterion and all the dependencies it pulls in that we have to audit, as well as anyhow would be really nice.
> > 
> > 
> > `anyhow` is so minimal anyway, and is used in `aranya` and other integrations. I don't see a reason to remove it from tests/bins. It's very convenient.
> 
> I would love to get rid of Criterion, but I think built-in benchmarking is still unstable. Running benchmarks with `+nightly` isn't a /huge/ deal, maybe. If I could just get proper nested benchmark (and testing, tbh) support ala https://pkg.go.dev/testing I would be so happy.

It's mostly just a nice-to-have, since I saw how many dependencies criterion pulls in for cargo vet because that includes dev-dependencies. I'd have to spend more time looking at how our tests actually work before I could be confident on how to approach that change.

crates/aranya-quic-syncer/examples/quic_syncer.rs
```diff
@@ -152,17 +134,12 @@ async fn run(options: Opt) -> Result<()> {
                 TestActions::Init(0),
                 sink.lock().await.deref_mut(),
             )
-            .map_err(|e| SyncError {
-                error_msg: e.to_string(),
-            })?;
+            .map_err(|e| anyhow!("SyncError: {e}"))?;
```
jdygert-spok, line 137: `map_err` not needed

crates/aranya-quic-syncer/examples/quic_syncer.rs
```diff
@@ -186,9 +163,7 @@ async fn run(options: Opt) -> Result<()> {
                 .lock()
                 .await
                 .action(storage_id, sink.lock().await.deref_mut(), action)
-                .map_err(|e| SyncError {
-                    error_msg: e.to_string(),
-                })?;
+                .map_err(|e| anyhow!("SyncError: {e}"))?;
```
jdygert-spok, line 166: ditto

crates/aranya-quic-syncer/examples/quic_syncer.rs
```diff
@@ -152,17 +134,12 @@ async fn run(options: Opt) -> Result<()> {
...
+            .map_err(|e| anyhow!("SyncError: {e}"))?;
         println!("Storage id: {}", storage_id)
     } else if let Some(id) = options.storage_id {
         storage_id = id;
     } else {
-        return Err(SyncError {
-            error_msg: "storage id is missing".to_string(),
-        }
-        .into());
+        return Err(anyhow!("SyncError: storage id is missing"));
```
jdygert-spok, line 142:
```suggestion
        bail!("storage id is missing");
```

jdygert-spok (commented):
nits:

jdygert-spok (approved):

gknopf-aranya:
@nikki-aranya one of the comments has unchecked checkboxes. Were those all resolved?

gknopf-aranya (approved):

nikki-aranya:
> @nikki-aranya one of the comments has unchecked checkboxes. Were those all resolved?

@gknopf-aranya Took care of those, needed to create a couple issues, also commented below them with what we decided, should be good to merge.

smckeon-so (approved):
🎊

benz-spideroak (approved):

gknopf-aranya:
@elagergren-spideroak or @ycarmy just needs review from one of you

ycarmy (approved):

PR #70, opened 2025-01-22
Fix git-unchanged check and fix lockfile

jdygert-spok:
`cargo-make` uses `sh` by default, which accepts bash-isms on our Mac dev machines, but not in CI. Since the problem was in an `if`, it just got always skipped from `[[ not found`.


jdygert-spok:
Updated

PR #75, opened 2025-01-23
69: update cache after commit

benz-spideroak:


crates/aranya-runtime/src/client.rs
```diff
@@ -142,18 +142,26 @@ where
...
+        storage_id: GraphId,
+        commands: &[impl Command],
+        request_heads: &mut PeerCache,
+    ) -> Result<(), ClientError> {
+        let storage = self.provider.get_storage(storage_id)?;
+        for command in commands {
+            if let Some(loc) = storage.get_location(command.address()?)? {
+                request_heads.add_command(storage, command.address()?, loc)?;
+            }
+        }
```
jdygert-spok, line 161: nit: pull out `let addr = command.address()?;`

crates/aranya-runtime/src/client.rs
```diff
@@ -142,18 +142,26 @@ where
...
 
+    pub fn update_heads(
+        &mut self,
+        storage_id: GraphId,
+        commands: &[impl Command],
+        request_heads: &mut PeerCache,
+    ) -> Result<(), ClientError> {
+        let storage = self.provider.get_storage(storage_id)?;
+        for command in commands {
+            if let Some(loc) = storage.get_location(command.address()?)? {
```
jdygert-spok, line 158: Is there a case where this being `None` isn't an error?

jdygert-spok: IOW should we treat a missing command as an error instead of silently ignoring it.

benz-spideroak: One of the times we update heads is when we get a sample from the peers graph. These commands may not be in our graph. I wasn't using `update_heads` in that case. I updated these code to also call `update_heads` then. Let me know if you think the errors are important and I should separate them.

jdygert-spok: Oh I see, I think this makes sense then.

crates/aranya-runtime/src/client.rs
```diff
@@ -142,18 +142,26 @@ where
...
-            request_heads,
-        )?;
+        let count = trx.add_commands(commands, &mut self.provider, &mut self.engine, sink)?;
         Ok(count)
     }
 
+    pub fn update_heads(
+        &mut self,
+        storage_id: GraphId,
+        commands: &[Address],
```
jdygert-spok, line 153:
```suggestion
        addrs: impl IntoIterator<Item = Address>,
```

jdygert-spok (approved):

gknopf-aranya:
@benz-spideroak could you "Update branch"?

crates/aranya-runtime/src/testing/testdata/large_sync.test
```diff
@@ -126,7 +126,7 @@
       "graph": 0,
       "client": 0,
       "from": 1,
-      "must_send": 1,
+      "must_send": 100,
```
gknopf-aranya, line 129: Curious why this was changed from 1 to 100. Seems like a significantly different number so wanting to understand the rationale.

PR #85, opened 2025-01-27
Remove limit on session command

jdygert-spok:
This is more consistent since we don't currently limit the command size anywhere else. We may want to handle session commands differently anyways. Limiting will later be handled by #84.


elagergren-spideroak (approved):

elagergren-spideroak:
it's also in the syncer iirc, grep for the constant

jdygert-spok:
> it's also in the syncer iirc, grep for the constant

Sort of, it only sets the overall sync message size limit:

```rust
pub const MAX_SYNC_MESSAGE_SIZE: usize = 1024 + MAX_COMMAND_LENGTH * COMMAND_RESPONSE_MAX;
```

djohnson-spideroak (approved):

gknopf-aranya (approved):

PR #97, opened 2025-02-11
Add `@generated` tag to code generators

jdygert-spok:
See <https://generated.at>.

I have not removed any `#[rustfmt::skip]` because then `cargo +stable fmt` would still change the generated files.

Thanks @nikki-aranya for finding this.


nikki-aranya (approved):
I had assumed the `//! This code is @generated by `policy-ifgen`. DO NOT EDIT.` header was also generated by the tool but seemingly not? Seems good if so.

jdygert-spok:
> I had assumed the `//! This code is @generated by `policy-ifgen`. DO NOT EDIT.` header was also generated by the tool but seemingly not? Seems good if so.

I'm not sure what you mean, I added them into the `quote!` calls of our generators so they are included in the code they generate.

nikki-aranya:
Ah okay, I hadn't looked closely enough at all the uses, it uses prettyplease/quote. Carry on.

elagergren-spideroak (approved):

PR #99, opened 2025-02-13
More type checking

### LINKED ISSUE #14: Check types for more things
With a functioning type checker, we can check more things at compile time:

- fields of structs and fact literals
- arguments to function calls
- match test expressions versus arm expressions
- check that an if expression is a boolean

As noted, marking `Typeish` as `#[must_use]` lights up all the places that need type checking. So that's a good place to start.

chip-so:
Closes #14

### DIFF
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -149,6 +149,12 @@ impl<'a> CompileState<'a> {
         // ensure key identifiers are unique
         let mut identifiers = BTreeSet::new();
         for key in fact.key.iter() {
+            if !key.is_hashable() {
+                return Err(self.err(CompileErrorType::InvalidType(format!(
+                    "Fact `{}` key field `{}` is not int, bool, string, or id",
+                    fact.identifier, key.identifier
+                ))));
+            }
             if !identifiers.insert(key.identifier.as_str()) {
                 return Err(self.err(CompileErrorType::AlreadyDefined(key.identifier.to_owned())));
             }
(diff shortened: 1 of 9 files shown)


crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1151,7 +1227,23 @@ impl<'a> CompileState<'a> {
...
+                                    n
+                                ))));
+                            }
+                        }
+                        Typeish::Type(ot) => {
+                            return Err(self.err(CompileErrorType::InvalidType(format!(
+                                "Cannot publish `{ot}`, must be a command struct"
+                            ))))
+                        }
+                        _ => {}
```
ssweetney-spok, line 1245: Does this allow for something like?

```
    action foo() {
        publish None
    }

**Review by apetkov-so — DISMISSED** [review]

File: crates/aranya-policy-compiler/src/compile.rs (hunk 1 above)
Comment (ssweetney-spok) [reply crates/aranya-policy-compiler/src/compile.rs:1245]: > It is an indeterminate type and thus is allowed anywhere

Should it be allowed here if we're expecting only Command structs to be published?

I could be wrong but I don't think any valid programs would be rejected if expressions with an indeterminate type weren't allowed.

I thought about something like

```
action foo() {
    // Assume Foo is a Command struct
    let cmd_struct = Foo {}
    let bytes = serialize(cmd_struct)
    let deserialized_foo = deserialize(bytes) // indeterminate type
    
    publish deserialized_foo
}
```

but using serialize/deserialize in an action results in a compilation error

ssweetney-spok: Ah, I see; `if` expressions do make things tricky. This makes a lot of sense.

ssweetney-spok (approved):
LGTM

ssweetney-spok:
Btw it looks like [the RUSTSEC advisory](https://rustsec.org/advisories/RUSTSEC-2025-0007) that caused the security check to fail has been withdrawn so it should pass after re-running it

crates/aranya-policy-compiler/src/compile.rs
jdygert-spok, line 1: There's some `ok_or` that should be `ok_or_else`, and I think some more unnecessary clones like Steve pointed out.

apetkov-so (dismissed):

jdygert-spok (approved):

jdygert-spok (approved):

elagergren-spideroak (dismissed):

jdygert-spok (approved):

apetkov-so (approved):

PR #107, opened 2025-02-24
101: Initial version of quic channels. Still missing the crypto parts.

benz-spideroak:


crates/aranya-quic-channels/src/lib.rs
jdygert-spok, line 1: I think this is on the right track. Some thoughts:

- How will labels be used?
- What happens if both sides open a channel?
- Can probably remove the stream mspc channel and instead use `stream.receive()` directly.
- Nice to have: for the streaming API, expose a type that implements something like `AsyncBufRead`/`AsyncWrite`

benz-spideroak: > How will labels be used?

They'll be used to identify channels between peers. That allows for multiple channels between peers. A label will be passed in when creating a channel and used to create the channel in the policy. 

> What happens if both sides open a channel?

That will be determined by the policy. It's fine if the labels are different. Each channel will have a separate connection which may result in multiple connections between a set of peers. 

If the labels are the same I'm guessing the first channel in the braid will be active and the second will be disregarded.

benz-spideroak: > Can probably remove the stream mspc channel and instead use stream.receive() directly.

If I do that I probably need to remove try_recv_stream. Since stream.receive() is async I can't have a non async method. I'm OK with doing that. I just wanted to note it.

crates/aranya-quic-channels/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+        select! {
+            Some(conn) = server.accept() => {
+                match handle_connection(conn).await {
+                    Ok(Some(channel)) => {
+                        if client
+                            .lock()
+                            .await
+                            .new_channels
+                            .push(channel).is_err() {
+                            error!("Channel full. Unable to insert channel");
```
gknopf-aranya, line 65: Include the error returned from `push()` in the trace:
```suggestion
error!("Unable to insert channel: {:?}", e);
```

crates/aranya-quic-channels/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+    while let Ok(Some(stream)) = conn.accept_receive_stream().await {
+        tokio::spawn(handle_message(stream, sender.clone()));
+    }
+}
+
+async fn handle_message(mut stream: ReceiveStream, sender: mpsc::Sender<Bytes>) {
+    let mut data = Vec::new();
+    loop {
+        match stream.receive().await {
+            Ok(Some(req)) => {
```
gknopf-aranya, line 112: Does `req` stand for request? This isn't necessarily request data, it could be any incoming data received by a `ReceiveStream`.

crates/aranya-quic-channels/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+                data.extend_from_slice(&req);
+            }
+            Ok(None) => {
```
gknopf-aranya, line 115: If `stream.receive()` returns `Ok(None)`, does that mean the stream has been closed and will not receive any more data?

crates/aranya-quic-channels/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+            Err(_) => {
+                debug!("error receiving from stream");
+                break;
+            }
+        }
+    }
+}
+
+#[derive(Eq, Hash, PartialEq, Clone, Copy, Debug)]
+/// Identifies a unique channel between two peers.
```
gknopf-aranya, line 130: We should differentiate from AFC in the docs:
```suggestion
/// Identifies a unique AQC channel between two peers.
```

crates/aranya-quic-channels/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+pub struct AqcChannelID {
+    channel_id: u64,
```
gknopf-aranya, line 132: Missing docs for public field

crates/aranya-quic-channels/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+                    match r {
+                        Err(_) => {
+                            debug!("error receiving from stream");
+                        }
+                        Ok(Some(req)) => {
+                let len = req.len();
+                target[..len].copy_from_slice(&req);
+                return Some(len);
+                        }
+                        // The stream has been closed.
```
gknopf-aranya, line 186: We should add `debug!` tracing to indicate when a stream has been closed.

crates/aranya-quic-channels/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+            message_sender,
+        )
+    }
+
+    /// Receive the next available data from a channel. If the channel has been
+    /// closed, return None.
+    ///
+    /// This method will block until data is available to return.
+    /// The data is not guaranteed to be complete, and may need to be called
+    /// multiple times to receive all data from a message.
```
gknopf-aranya, line 172: How would the caller know when all the data has been received?

crates/aranya-quic-channels/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+
+    /// Close the given channel if it's open. If the channel is already closed, do nothing.
+    pub fn close(&mut self) {
+        const ERROR_CODE: u32 = 0;
+        self.send_stream.connection().close(ERROR_CODE.into());
+    }
+}
+
+/// The maximum number of channels that haven't been received.
+const MAXIMUM_UNRECEIVED_CHANNELS: usize = 10;
```
gknopf-aranya, line 250: This seems like a really low number. Especially for environments where peer connectivity is not guaranteed.

crates/aranya-quic-channels/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+
+/// An AQC client. Used to create and receive channels.
+#[derive(Debug)]
+pub struct AqcClient {
+    quic_client: Client,
```
gknopf-aranya, line 255: Pub field needs docs

crates/aranya-quic-channels/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+    /// Holds channels that have created, but not yet been received.
+    pub new_channels: HVec<AqcChannel, MAXIMUM_UNRECEIVED_CHANNELS>,
+}
+
+impl AqcClient {
+    /// Create an Aqc client with the given certificate chain.
+    pub fn new<T: provider::tls::Provider>(cert: T) -> Result<AqcClient, AqcError> {
+        let quic_client = Client::builder()
+            .with_tls(cert)?
+            .with_io("0.0.0.0:0")?
```
gknopf-aranya, line 265: I don't think we want to hard-code this to `0.0.0.0:0`

crates/aranya-quic-channels/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+
+    /// Receive the next available channel. If no channel is available, return None.
+    /// This method will return a channel created by a peer that hasn't been received yet.
+    pub fn receive_channel(&mut self) -> Option<AqcChannel> {
+        self.new_channels.pop()
+    }
+
+    /// Create a new channel to the given address.
+    pub async fn create_channel(&mut self, addr: SocketAddr) -> Result<AqcChannel, AqcError> {
+        // TODO: Create the channel in the graph.
```
gknopf-aranya, line 281: @ycarmy, @djohnson-spideroak we can start making policy updates to support adding AQC channels to the graph

gknopf-aranya: Created an issue for making those updates to the policy here:
https://github.com/aranya-project/aranya/issues/140

crates/aranya-quic-channels/src/lib.rs
gknopf-aranya, line 1: I don't see any code that handles sending/receiving the ephemeral session command for setting up the AQC channel in the graph prior to sending encrypted data via the channel.

benz-spideroak: That requires the crypto to produce the key we want to send.

crates/aranya-quic-channels/src/lib.rs
benz-spideroak: That field isn't pub.

crates/aranya-quic-channels/src/lib.rs
benz-spideroak: When the stream is closed.

crates/aranya-quic-channels/src/lib.rs
benz-spideroak: Why would we want to know that? Closing streams is expected and if closed without an error means things are working as expected.

crates/aranya-quic-channels/src/lib.rs
benz-spideroak: Yes.

crates/aranya-quic-channels/src/lib.rs
jdygert-spok: Yeah it sometimes matters for firewall reasons.

crates/aranya-quic-channels/src/lib.rs
gknopf-aranya: Having a way to tell from the logs how many AQC channels, streams, peers there are seems useful for debugging.

crates/aranya-quic-channels/src/lib.rs
gknopf-aranya: Let's mention that in the docs here

crates/aranya-quic-channels/src/lib.rs
gknopf-aranya: The code could be mocked up with a `Vec<u8>` for now until it's replaced with the crypto keys later.

crates/aranya-quic-channels/tests/test.rs
```diff
@@ -0,0 +1,137 @@
...
+        assert_eq!(total_len, big_data.len());
+        assert_eq!(&target[..total_len], &big_data[..]);
+    } else {
+        panic!("channel is not available");
+    }
+
+    Ok(())
+}
+
+fn get_server(cert: String, key: String) -> Result<Server> {
```
gknopf-aranya, line 114: `get_server` implies that a server is being returned from an existing object. This is allocating a new server:
```suggestion
fn new_server(cert: String, key: String) -> Result<Server> {
```

crates/aranya-quic-channels/tests/test.rs
```diff
@@ -0,0 +1,137 @@
...
+    protocol::{TestActions, TestEngine, TestSink},
+    storage::memory::MemStorageProvider,
+    ClientState,
+};
+use bytes::Bytes;
+use s2n_quic::{provider::congestion_controller::Bbr, Server};
+use tokio::sync::Mutex as TMutex;
+
+#[test_log::test(tokio::test)]
+async fn test_channels() -> Result<()> {
```
gknopf-aranya, line 16: To differentiate from other tests (e.g. afc), we should include aqc in the name of the test:
```suggestion
async fn test_aqc_channels() -> Result<()> {
```
or
```suggestion
async fn test_quic_channels() -> Result<()> {
```

crates/aranya-quic-channels/tests/test.rs
```diff
@@ -0,0 +1,137 @@
...
+fn spawn_channel_listener(
+    aqc_client: Arc<TMutex<AqcClient>>,
+    server: Server,
+) -> Result<SocketAddr> {
+    let server_addr = server.local_addr()?;
+    tokio::spawn(run_channels(aqc_client, server));
+    Ok(server_addr)
+}
+
+fn make_client() -> Arc<TMutex<ClientState<TestEngine, MemStorageProvider>>> {
```
gknopf-aranya, line 132: Our codebase generally uses `new` in the method name when creating new objects.
```suggestion
fn new_client() -> Arc<TMutex<ClientState<TestEngine, MemStorageProvider>>> {
```

crates/aranya-quic-channels/tests/test.rs
```diff
@@ -0,0 +1,137 @@
...
+    let _ = spawn_channel_listener(aqc_client1.clone(), server1)?;
+    tokio::time::sleep(Duration::from_millis(100)).await;
+
+    let server2 = get_server(cert.clone(), key)?;
+    let aqc_client2 = Arc::new(TMutex::new(AqcClient::new(&*cert)?));
+    let addr2 = spawn_channel_listener(aqc_client2.clone(), server2)?;
+    let mut channel1 = aqc_client1.lock().await.create_channel(addr2).await?;
+    tokio::time::sleep(Duration::from_millis(100)).await;
+
+    if let Some(mut channel2) = aqc_client2.lock().await.receive_channel() {
```
gknopf-aranya, line 43: I'd like to see a test with at least 3 aqc peers that opens multiple simultaneous channels/streams between each peer and sends data on each of those streams.

crates/aranya-quic-channels/src/lib.rs
gknopf-aranya, line 1: @benz-spideroak, @djohnson-spideroak, @elagergren-spideroak is there a reason for implementing AQC in the `aranya-core` repo when AFC was implemented in the `aranya` repo?
https://github.com/aranya-project/aranya/blob/main/crates/aranya-client/src/afc.rs

I'd prefer to implement similar features at similar places in the architecture unless we have a good reason for implementing them in a different place. Otherwise, this could cause some confusion and additional work to integrate the `aranya-core` implementation into the `aranya` repo.

For AFC, the primary code that was implemented in the `aranya-core` repo previously was `aranya-crypto::afc` and the `aranya-afc-util` FFI.

For AQC I'm suggesting:
- Implement AQC in the `aranya` repo at `aranya-client/src/aqc.rs`.
- Implement AQC crypto in the `aranya-core` repo
- Implement AQC FFIs in the `aranya-core` repo

jdygert-spok: There is also `aranya-fast-channels` in `aranya-core`, which implements the shared memory, seal/open client, and headers.

I think we should err on the side of implementing in `aranya-core` for more reusability. But some of what's here may end up in `aranya`. We have mentioned possibly keeping networking out of `aranya-core` :shrug:

jdygert-spok: > How will labels be used?

I mean where will they be in the API? Can they be added now?

> If the labels are the same I'm guessing the first channel in the braid will be active and the second will be disregarded.

I believe channels are set up off-graph like AFC, so there will be no braid. If we choose either the first or second for priority, in either case the two sides will end up with different channels. I guess we'll either want to allow a quic-connection per side for a given channel, or somehow combine them?

gknopf-aranya: I think whichever approach we use should be consistent for AFC and AQC. Don't want to spend time refactoring AFC to match AQC right now since it's going behind an experimental flag. We could refactor both to use a similar approach later if that's desired.

crates/aranya-quic-channels/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+    Io(#[from] std::io::Error),
+    /// An internal AQC error.
+    #[error("internal error: {0}")]
+    Internal(#[from] anyhow::Error),
+}
+
+/// Runs a server listening for quic channel requests from other peers.
+pub async fn run_channels(client: Arc<TMutex<AqcClient>>, mut server: Server) {
+    loop {
+        select! {
```
nikki-aranya, line 56: Is this select actually doing anything? There only seems to be one case, and I'm hesitant to keep using tokio::select unless we have to due to how it adds cancellation complexity

crates/aranya-quic-channels/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+
+impl AqcChannel {
+    /// Create a new channel with the given send stream.
+    ///
+    /// Returns the channel and the senders for the stream and message channels.
+    pub fn new(
+        send_stream: SendStream,
+        receive_stream: ReceiveStream,
+    ) -> (Self, mpsc::Sender<Bytes>) {
+        let (message_sender, message_receiver) = mpsc::channel(1);
```
nikki-aranya, line 156: Is an mpsc really the best mechanism if we're only allowing a buffer size of 1? I know it's easy to reach for and we use it often in this codebase, but I don't think it's always the correct one to choose

mooreoak (approved):

gknopf-aranya:
This PR needs to be moved to the `aranya` repo.

PR #116, opened 2025-02-28
Add struct composition

### LINKED ISSUE #104: (Policy) Add Struct Composition
Add the struct composition operator to allow fields to be pulled from other structs when creating a struct.

See [Policy Lang v2 spec](https://github.com/aranya-project/aranya-docs/blob/main/docs/policy-v2.md#struct-composition)

ssweetney-spok:
Closes #104

~~Note that the struct composition operator is not allowed when declaring global values. It's also not allowed in a command's attribute block.~~

Struct composition *does* work when declaring global values and in a command's attribute block

# Implementation

During the compilation step, transform something like

```
let b = Bar {
        ...x,
        c: false,
}
```

into

```
let b = Bar {
        a: x.a,
        b: x.b,
        c: false,
}
```


ssweetney-spok:
Opening this up for feedback now but I'm waiting until #80 is merged into main and then I'll update the base branch

crates/aranya-policy-ast/src/ast.rs
```diff
@@ -179,6 +179,8 @@ pub struct NamedStruct {
     pub identifier: String,
     /// The fields, which are pairs of identifiers and expressions
     pub fields: Vec<(String, Expression)>,
+    /// sources is a list of identifiers
```
apetkov-so, line 182: Maybe mention struct composition, to make it clearer how these are used.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2045,6 +2049,99 @@ impl<'a> CompileState<'a> {
...
+                    .map(|field_defns| (ident, field_defns, type_name)),
+                Typeish::Type(_) => Err(self.err(CompileErrorType::InvalidType(format!(
+                    "Expected `{ident}` to be a struct"
+                )))),
+                Typeish::Indeterminate => unreachable!(),
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let mut seen = BTreeMap::new();
+        for (source_struct_name, field_defns, source_struct_type_name) in source_field_defns {
```
apetkov-so, line 2092: Maybe the nested loops can be replaced with `std::iter::zip`?

crates/aranya-policy-compiler/src/compile/error.rs
```diff
@@ -60,6 +60,12 @@ pub enum CompileErrorType {
     /// A validation step failed
     #[error("validation failed")]
     Validation,
+    /// Source structs in struct composition have overlapping fields
+    #[error("Struct {0} and Struct {1} have at least 1 field with the same name")]
+    DuplicateSourceFields(String, String),
+    /// The source struct is not a subset of the base struct
+    #[error("Struct {0} must be a subset of Struct {1}")]
+    SourceStructTooManyFields(String, String),
```
apetkov-so, line 68: This variant name doesn't match the description. Maybe rename it to something like `SourceStructNotSubsetOfBase`?

apetkov-so (dismissed):

jdygert-spok:
> Note that the struct composition operator is not allowed when declaring global values. It's also not allowed in a command's attribute block.

I think we could support these, do we want to?

crates/aranya-policy-lang/src/lang/parse/policy.pest
```diff
@@ -61,16 +61,20 @@ struct_def = _{ "{" ~ field_definition_list? ~ "}" }
 // An effect struct definition is similar but supports the additional
 // "dynamic" keyword.
 effect_struct_def = _{ "{" ~ effect_field_definition_list? ~ "}" }
+// Struct composition is shorthand for including the fields of some other
+// struct(s) when defining a struct literal
+struct_composition = { "..." ~ identifier }
```
jdygert-spok, line 66: Do we want to only allow identifiers or should we accept any expression?

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2045,6 +2049,99 @@ impl<'a> CompileState<'a> {
...
+                            "Unknown identifier `{src_ident}`"
+                        )))
+                    })
+                    .map(|ident_type| (src_ident, ident_type))
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let source_field_defns: Vec<_> = source_types
+            .into_iter()
+            .filter(|(_, src_type)| matches!(*src_type, Typeish::Type(_))) // Ignore identifiers with an indeterminate type
```
jdygert-spok, line 2075: It seems like this just entirely ignores sources with indeterminate type, we should probably error out. Or we could support them but that seems like much more effort when we want to eventually remove indeterminate.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2045,6 +2049,99 @@ impl<'a> CompileState<'a> {
...
+                        }
+                        e.insert(source_struct_type_name.clone());
+                    }
+                    Entry::Occupied(other) => {
+                        let (struct_1, struct_2) =
+                            (source_struct_type_name.to_string(), other.get().to_string());
+                        return Err(
+                            self.err(CompileErrorType::DuplicateSourceFields(struct_1, struct_2))
+                        );
+                    }
```
jdygert-spok, line 2139: I wonder if we should allow overlap when we set the field anyway:
```
struct SrcA { x: int, a: int }
struct SrcB { x: int, b: int }
struct Foo { x: int, a: int, b: int }

let foo = Foo {
    x: 1,
    ...src_a,
    ...src_b,
}
```

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2045,6 +2049,99 @@ impl<'a> CompileState<'a> {
...
+                    .get(&type_name)
+                    .assume("identifier with a struct type has that struct already defined")
+                    .map_err(|err| self.err(err.into()))
+                    .map(|field_defns| (ident, field_defns, type_name)),
+                Typeish::Type(_) => Err(self.err(CompileErrorType::InvalidType(format!(
+                    "Expected `{ident}` to be a struct"
+                )))),
+                Typeish::Indeterminate => unreachable!(),
+            })
+            .collect::<Result<Vec<_>, _>>()?;
```
jdygert-spok, line 2089: nit: I think in this case having these separate from the for loop below makes this harder to follow.

crates/aranya-policy-lang/src/lang/parse/policy.pest
ssweetney-spok: Allowing other types of expressions seems reasonable. Something like:

```
let b = Bar {
        ... x substruct Foo,
        c: false,
}
```

could be useful

crates/aranya-policy-compiler/src/compile.rs
ssweetney-spok: Yea, this in this case there isn't any ambiguity about which source struct should take precedence so it should work here. Thoughts @apetkov-so @chip-so ?

crates/aranya-policy-compiler/src/compile.rs
ssweetney-spok: When I initially wrote this, I didn't want to error out because I was thinking of [Chip's comment](https://github.com/aranya-project/aranya-core/pull/99#discussion_r1966304073) so I did want to support indeterminate types but as I got further in the implementation I realized that would've been a bit harder. Like you said, we could support them but I don't think it's easily doable with my current approach but I could move the source transformation into the VM

ssweetney-spok:
> > Note that the struct composition operator is not allowed when declaring global values. It's also not allowed in a command's attribute block.
> 
> I think we could support these, do we want to?

I couldn't tell from reading the spec but I assumed based on the usage of `expression_value` that global let statements and attribute blocks only allowed for "simple" expressions that could be evaluated to `aranya_policy_module::Value`s at compile time

crates/aranya-policy-lang/src/lang/parse/policy.pest
```diff
@@ -61,16 +61,20 @@ struct_def = _{ "{" ~ field_definition_list? ~ "}" }
...
 // A struct literal field is an identifier and an expression separated
 // by a colon. The colon marks this as defining the value of the field
 // instead of its type.
 struct_literal_field = { identifier ~ ":" ~ expression }
-// A struct literal field list is a list of struct literal fields.
-struct_literal_field_list = _{ struct_literal_field ~ ("," ~ struct_literal_field)* ~ ","? }
+struct_data = _{ (struct_literal_field | struct_composition) }
+// A struct data list is a list of struct literal fields and/or struct compositions.
+struct_data_list = _{ struct_data ~ ("," ~ struct_data)* ~ ","? }
```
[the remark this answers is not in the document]
jdygert-spok: Yeah rust requires it last, with preceding fields overriding the spread ones. But python and javascript both allow it anywhere, and have the later ones take priority in merging...

jdygert-spok:
There is an inefficiency in field access that we are aware of, but this makes it possible with much less code. For a struct with `n` fields, just spreading the struct into itself will make `n` clones of the struct and thus do `n^2` field clones. Is this something we want to worry about here or handle later with a general solution?

crates/aranya-policy-lang/src/lang/parse/policy.pest
mooreoak: I would go with only identifiers for now and see how that feels.

crates/aranya-policy-lang/src/lang/parse/policy.pest
```diff
@@ -61,16 +61,22 @@ struct_def = _{ "{" ~ field_definition_list? ~ "}" }
...
+struct_composition_list =  _{ struct_composition ~ ("," ~ struct_composition)* }
 // A struct literal field is an identifier and an expression separated
 // by a colon. The colon marks this as defining the value of the field
 // instead of its type.
 struct_literal_field = { identifier ~ ":" ~ expression }
 // A struct literal field list is a list of struct literal fields.
-struct_literal_field_list = _{ struct_literal_field ~ ("," ~ struct_literal_field)* ~ ","? }
+struct_literal_field_list = _{ struct_literal_field ~ ("," ~ struct_literal_field)* }
+// A struct data list is a list of struct literal fields followed by struct compositions.
+struct_data_list = _{ ((struct_literal_field_list ~ "," ~ struct_composition_list) | struct_literal_field_list | struct_composition_list) ~ ","? }
```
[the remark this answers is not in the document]
ssweetney-spok: Yea, field-value pairs can't come after struct composition(s). It seemed like the consensus we had in [this convo](https://github.com/aranya-project/aranya-core/pull/116#discussion_r1982045249). Good point about the spec. I can create an issue to to update it

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -1789,6 +1836,139 @@ fn test_type_errors() -> anyhow::Result<()> {
...
+#[test]
+fn test_struct_composition_global_let_and_command_attributes() -> anyhow::Result<()> {
+    let policy_str = r#"
+        struct Foo {
+            x int,
+            y int
+        }
+
+        let foo = Foo { x: 10, y: 20 }
+        let foo2 = Foo { x: 1000, ...foo }
```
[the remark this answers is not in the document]
ssweetney-spok: Ah, I missed that requirement. I was thinking it worked like Rust's struct update syntax.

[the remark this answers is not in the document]
ssweetney-spok: FWIW, it seems like this approach was *easier* since checking to see to see if a struct A is a subset of struct B is easier than checking if struct A is a strict subset of struct B 🤷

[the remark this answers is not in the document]
ssweetney-spok: Yea, this all sounds reasonable.

> I think it is still good to prevent overlapping fields in the composed structs.

Agreed. This is already done

> I think it should be an error to compose when all fields are explicitly defined

I can see the reasoning behind why this should error. I haven't implemented that but this should be an easy fix

apetkov-so (approved):

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2212,26 +2212,166 @@ impl<'a> CompileState<'a> {
...
+
+        let field_idents: HashSet<_> = strukt
+            .fields
+            .iter()
+            .map(|(field_name, _)| field_name)
+            .collect();
+
+        if strukt_defns
+            .iter()
+            .all(|defn| field_idents.contains(&defn.identifier))
```
jdygert-spok, line 2370: I guess this could just be `strukt.fields.len() == strukt_defns.len()` ?

ssweetney-spok: Ahh, yea this can be simplified

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2226,27 +2232,144 @@ impl<'a> CompileState<'a> {
...
+            let src_type = self
+                .identifier_types
+                .get(src_var_name)
+                .map_err(|err| self.err(err))?;
+
+            let src_struct_type_name = match src_type {
+                Typeish::Type(VType::Struct(type_name)) => type_name,
+                Typeish::Type(_) | Typeish::Indeterminate => {
+                    return Err(self.err(CompileErrorType::InvalidType(format!(
+                        "Expected `{src_var_name}` to be a struct"
```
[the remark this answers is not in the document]
ssweetney-spok: > Telling the developer "Expected definitely_a_struct to be a struct" is a very confusing message.

Ah, yea. That's a good point.

> Something like "Cannot compile struct composition because type resolution failed for var" is still not helpful, at least it doesn't point them in the wrong direction.

That is definitely less confusing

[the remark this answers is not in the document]
jdygert-spok: Yeah `None` will no longer be indeterminate after #321, and #323 gave `deserialize` the proper type as well.

The remaining cases of indeterminate will be `todo()` and stubbed ffi.

It's unfortunate that if someone uses struct composition with the output of an ffi call, then you can't compile with stubbed ffi, which is currently needed for policy-ifgen. (#330).

crates/aranya-policy-compiler/src/compile/types.rs
```diff
@@ -125,7 +125,9 @@ impl IdentifierTypeStack {
         if let Some(v) = self.globals.get(name) {
             return Ok(v.clone());
         }
-        Err(CompileErrorType::NotDefined(name.to_string()))
+        Err(CompileErrorType::NotDefined(format!(
+            "Unknown identifier `{name}`"
```
apetkov-so, line 129: nit: I'm not sure the message adds much value - the type already provides enough context.

ssweetney-spok: I think I added additional text because this error variant is used for both undefined types and undefined identifiers.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2251,15 +2250,15 @@ impl<'a> CompileState<'a> {
         // Panic when running a module without setup.
         self.append_instruction(Instruction::Exit(ExitReason::Panic));
 
+        for struct_def in &self.policy.structs {
```
apetkov-so, line 2253: Moving struct definitions here allows using composition in global vars, right?

ssweetney-spok: Yea it does. I didn't have that initially but updated it based on [Chip's feedback](https://github.com/aranya-project/aranya-core/pull/116#issuecomment-2701823690).

[Related test](https://github.com/aranya-project/aranya-core/pull/116/files)

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2360,27 +2359,151 @@ impl<'a> CompileState<'a> {
...
-            Expression::EnumReference(e) => self.enum_value(e).ok(),
-            _ => None,
+            Expression::Int(v) => Ok(Value::Int(*v)),
+            Expression::Bool(v) => Ok(Value::Bool(*v)),
+            Expression::String(v) => Ok(Value::String(v.clone())),
+            Expression::NamedStruct(struct_ast) => {
+                let Some(struct_def) = self.m.struct_defs.get(&struct_ast.identifier).cloned()
+                else {
+                    return Err(self.err(CompileErrorType::NotDefined(format!(
+                        "Struct `{}` not defined",
```
apetkov-so, line 2371: Most `NotDefined` error instances just specify the identifier, without additional messages. I don't feel very strongly about it, but it's probably worth keeping them consistent. Exception: add the type before the identifier, e.g. "struct/enum {ident}".

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2384,27 +2384,150 @@ impl<'a> CompileState<'a> {
...
+                // Known type, but not a struct
+                Typeish::Known(other_type) => {
+                    return Err(self.err(CompileErrorType::InvalidType(format!(
+                        "Expected `{src_var_name}` to be a struct, but it's a(n) {other_type}",
+                    ))));
+                }
+                Typeish::Never => {
+                    // Never-typed expressions panic at runtime, making this code
+                    // unreachable. Skip to continue type checking.
+                    continue;
```
ssweetney-spok, line 2476: Does this comment make sense or is it confusing? I'm re-reading it and I can see it being the latter.

When I wrote "this code", I was referring to the struct composition operation in the policy code

cc: @apetkov-so @chip-so @jdygert-spok

PR #119, opened 2025-03-04
Close #110: Add match expressions.

### LINKED ISSUE #110: Add match expressions
match now works in expression context, using block expressions for its arms. While this looks similar to the existing match statement, it might require a different approach.

c.f. [Policy Lang v2 spec](https://git.spideroak-inc.com/spideroak-inc/flow3-docs/blob/main/src/policy-v2.md)

apetkov-so:
Allows `match` to be use as an expression. E.g.:
```
let x = match n {
    1 => {
      :"one"
    }
    _ => "other"
  }
```

### DIFF
--- a/crates/aranya-policy-ast/src/ast.rs
+++ b/crates/aranya-policy-ast/src/ast.rs
@@ -389,13 +389,13 @@ pub struct MatchExpression {
     pub arms: Vec<AstNode<MatchExpressionArm>>,
 }
 
-/// A container for one of two possible values
+/// A container for a statement or expression
 #[derive(Debug, Clone, PartialEq)]
-pub enum Either<A, B> {
-    /// option A
-    First(A),
-    /// option B
-    Second(B),
+pub enum LanguageContext<A, B> {
+    /// statement
+    Statement(A),
+    /// expression
+    Expression(B),
 }
 
 /// Match arm expression
--- a/crates/aranya-policy-ast/src/ast.rs
+++ b/crates/aranya-policy-ast/src/ast.rs
@@ -386,7 +386,7 @@ pub struct MatchExpression {
     /// Value to match against
     pub expression: Expression,
     /// Match arms
-    pub arms: Vec<AstNode<MatchArmExpression>>,
+    pub arms: Vec<AstNode<MatchExpressionArm>>,
 }
 
 /// A container for one of two possible values
@@ -400,7 +400,7 @@ pub enum Either<A, B> {
 
 /// Match arm expression
 #[derive(Debug, Clone, PartialEq)]
-pub struct MatchArmExpression {
+pub struct MatchExpressionArm {
     /// value to match against the match expression
     pub pattern: MatchPattern,
     /// Expression
--- a/crates/aranya-policy-ast/src/ast.rs
+++ b/crates/aranya-policy-ast/src/ast.rs
@@ -384,7 +384,7 @@ pub struct MatchStatement {
 #[derive(Debug, Clone, PartialEq)]
 pub struct MatchExpression {
     /// Value to match against
-    pub expression: Expression,
+    pub scrutinee: Expression,
     /// Match arms
     pub arms: Vec<AstNode<MatchExpressionArm>>,
 }
--- a/crates/aranya-policy-ast/src/ast.rs
+++ b/crates/aranya-policy-ast/src/ast.rs
@@ -317,6 +317,8 @@ pub enum Expression {
     Is(Box<Expression>, bool),
     /// A block expression
     Block(Vec<AstNode<Statement>>, Box<Expression>),
+    /// Match expression
+    MatchExpression(Box<Expression>, Vec<AstNode<MatchArmExpression>>),
 }
 
 /// Encapsulates both [FunctionDefinition] and [FinishFunctionDefinition] for the purpose
@@ -376,6 +378,24 @@ pub struct MatchStatement {
     pub expression: Expression,
     /// All of the potential match arms
     pub arms: Vec<MatchArm>,
+}
+
+/// Match statement expression
+#[derive(Debug, Clone, PartialEq)]
+pub struct MatchExpression {
+    /// Value to match against
+    pub expression: Expression,
+    /// Match arms
+    pub arms: Vec<MatchArmExpression>,
+}
+
+/// Match arm expression
+#[derive(Debug, Clone, PartialEq)]
+pub struct MatchArmExpression {
+    /// value to match against the match expression
+    pub pattern: MatchPattern,
+    /// Expression
+    pub expression: Expression,
 }
 
 /// Test a series of conditions and execute the statements for the first true condition.
--- a/crates/aranya-policy-ast/src/ast.rs
+++ b/crates/aranya-policy-ast/src/ast.rs
@@ -318,7 +318,7 @@ pub enum Expression {
     /// A block expression
     Block(Vec<AstNode<Statement>>, Box<Expression>),
     /// Match expression
-    MatchExpression(Box<Expression>, Vec<AstNode<MatchArmExpression>>),
+    Match(Box<MatchExpression>),
 }
 
 /// Encapsulates both [FunctionDefinition] and [FinishFunctionDefinition] for the purpose
@@ -386,7 +386,16 @@ pub struct MatchExpression {
     /// Value to match against
     pub expression: Expression,
     /// Match arms
-    pub arms: Vec<MatchArmExpression>,
+    pub arms: Vec<AstNode<MatchArmExpression>>,
+}
+
+/// A container for one of two possible values
+#[derive(Debug, Clone, PartialEq)]
+pub enum Either<A, B> {
+    /// option A
+    First(A),
+    /// option B
+    Second(B),
 }
 
 /// Match arm expression
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -1011,8 +1011,10 @@ impl<'a> CompileState<'a> {
                 subexpression_type
             }
             Expression::Match(e) => {
-                let expr_type = self.compile_match_statement_or_expression(Either::Second(e), 0)?;
-                expr_type.unwrap_or(Typeish::Indeterminate)
+                let expr_type = self
+                    .compile_match_statement_or_expression(Either::Second(e), 0)?
+                    .assume("match expression must return a type")?;
+                expr_type
             }
         };
 
@@ -1875,6 +1877,8 @@ impl<'a> CompileState<'a> {
         Ok(())
     }
 
+    /// Compile a match statement or expression
+    /// Returns the type of the `match` is an expression, or `None` if it's a statement.
     fn compile_match_statement_or_expression(
         &mut self,
         s: Either<&MatchStatement, &MatchExpression>,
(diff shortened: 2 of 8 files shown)


crates/aranya-policy-lang/src/lang/parse/policy.pest
```diff
@@ -152,16 +152,30 @@ at_least = { "at_least" ~ int_literal ~ fact_literal}
 at_most = { "at_most" ~ int_literal ~ fact_literal }
 // count facts, returning true if the number of facts found equals the expected number
 exactly = { "exactly" ~ int_literal ~ fact_literal }
+// Similar to the `match` statement, but resolves to a value. e.g. `let x = match { ... }`
+match_expression = { "match" ~ expression ~ "{" ~ match_arm_expr* ~ "}" }
+match_arm_expr = { (match_arm_expression | match_default) ~ "=>" ~ block_expression }
```
jdygert-spok, line 157: I think we should allow any expression in the arm body.

```suggestion
match_arm_expr = { (match_arm_expression | match_default) ~ "=>" ~ expression }
```

My only concern is parsing ambiguity but I _think_ we are currently fine.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -958,6 +958,36 @@ impl<'a> CompileState<'a> {
...
+                        Some(ref t) => {
+                            if !t.is_equal(&etype) {
+                                return Err(self.err(CompileErrorType::InvalidType(format!(
+                                    "match arm expression type mismatch; expected {}, got {}",
+                                    t, etype
+                                ))));
+                            }
+                        }
+                    }
+                }
```
jdygert-spok, line 977: This is a little more permissive than it should be due to indeterminate type, I think this will pass:

```
match x {
    0 => None
    1 => ""
    2 => 42
}
```

apetkov-so: Added test for this, it doesn't pass. The error for this case is "expected indeterminate, got string". The only other expression variant that returns `Indeterminate` is `Deserialize`, which has a TODO (for @chip-so) use the command info to return the correct type.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1827,6 +1757,137 @@ impl<'a> CompileState<'a> {
...
+                    self.append_instruction(Instruction::Jump(Target::Unresolved(
+                        arm_label.clone(),
+                    )));
+
+                    // Ensure this is the last case, and also that it's not the only case.
+                    if pattern != patterns.last().expect("last arm") {
+                        return Err(self.err(CompileErrorType::Unknown(String::from(
+                            "Default match case must be last.",
+                        ))));
+                    }
```
ssweetney-spok, line 1827: I know this was already added from compiling match statements but I just noticed that this seems like an error that should be caught while parsing.

jdygert-spok: Hmm I feel like it's not really a syntactic error. Especially if you think of `_` as just a pattern which happens to catch everything.

ssweetney-spok: I see what you mean. To me, it seems like allowing for match patterns after the default, `_`, breaks the rules of the language which feels like a syntactic error

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -1226,6 +1234,55 @@ fn test_match_arm_should_be_limited_to_literals() -> anyhow::Result<()> {
...
+        ),
+        (
+            // expression type doesn't match expected type
+            r#"function f(n int) bool {
+                return match n {
+                    0 => 1
+                    _ => 0
+                }
+            }"#,
+            CompileErrorType::InvalidType("Return value of `f()` must be bool".to_string()),
```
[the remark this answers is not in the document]
apetkov-so: It ensures the `match` expression evaluates to the correct type.

crates/aranya-policy-compiler/src/compile.rs
apetkov-so: This test now passes, i.e. the expression evaluates to `Indeterminate`. I guess that's fine, until we can resolve all expressions to a definite type.

crates/aranya-policy-lang/src/lang/parse.rs
```diff
@@ -611,6 +606,80 @@ impl<'a> ChunkParser<'a> {
...
+                        .into_inner()
+                        .map(|token| {
+                            let expr = self.parse_expression(token.to_owned())?;
+                            // Ensure expression values are all literals
+                            if !matches!(
+                                expr,
+                                Expression::Int(_)
+                                    | Expression::String(_)
+                                    | Expression::Bool(_)
+                                    | Expression::EnumReference(_)
```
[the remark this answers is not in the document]
apetkov-so: The compiler was already doing that. I just removed the check from the parser.

crates/aranya-policy-lang/src/lang/parse/policy.pest
```diff
@@ -152,16 +152,30 @@ at_least = { "at_least" ~ int_literal ~ fact_literal}
...
 // don't fit into the pratt parser.
-internal_function = _{ query | exists | count_up_to | at_least | at_most | exactly | if_e | serialize | deserialize }
+internal_function = _{
+    query |
+    exists |
+    count_up_to |
+    at_least |
+    at_most |
+    exactly |
+    match_expression |
```
apetkov-so, line 174: Hmm, I think I did something wrong here. match is not an internal function...

apetkov-so: Moved `match_expression` to `atom`.

crates/aranya-policy-lang/src/lang/parse/policy.pest
apetkov-so: It is intentional, yes. I've added a [ticket](https://github.com/aranya-project/aranya-docs/issues/87) to update the spec.

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -1226,6 +1234,57 @@ fn test_match_arm_should_be_limited_to_literals() -> anyhow::Result<()> {
...
+    let valid_cases = vec![
+        // match expression type is indeterminate
+        r#"function f(n int) int {
+            return match n {
+                0 => None
+                1 => "1"
+                _ => 0
+            }
+        }"#,
+    ];
```
ssweetney-spok, line 1281: Something like this could be a compile time error if we kept track of the number of different types that each arm expression evaluates to. 

`let arm_types = [Typeish::Indeterminate, Typeish::Type(String), Typeish::Type(Int)]`

Then we'd know this should be an error since there's 2 non-indeterminate types in this match expression

ssweetney-spok (dismissed):

crates/aranya-policy-compiler/src/tests.rs
ssweetney-spok: Good question, in that case there isn't enough information to determine if the compiler should error so it'll have to be checked at runtime. It's probably not worth handling right now. Like you said, eventually we'll try to resolve all expressions to a type

jdygert-spok: I have been wondering if `unify_pair` should pick the known type instead of indeterminate when only one is known. I think that's better in many cases, but maybe would cause an issue somewhere?

crates/aranya-policy-ast/src/ast.rs
```diff
@@ -378,6 +380,33 @@ pub struct MatchStatement {
     pub arms: Vec<MatchArm>,
 }
 
+/// Match statement expression
+#[derive(Debug, Clone, PartialEq)]
+pub struct MatchExpression {
+    /// Value to match against
+    pub expression: Expression,
```
jdygert-spok, line 387: I found out this is called a "scrutinee", if we want to be fancy

crates/aranya-policy-lang/src/lang/parse/policy.pest
```diff
@@ -174,6 +187,7 @@ atom = _{
     bool_literal |
     optional_literal |
     named_struct_literal |
+    match_expression |
```
jdygert-spok, line 190: Smells fishy to have match expressions and if expressions in different groups. Not sure internal_function and atom are really right TBH. But it's all arbitrary :shrug: not a blocker IMO

crates/aranya-policy-ast/src/ast.rs
apetkov-so: Ah, I [see](https://doc.rust-lang.org/reference/expressions/match-expr.html). I thought it was called a predicate.

elagergren-spideroak: I prefer "matchee" :D

crates/aranya-policy-lang/src/lang/parse/policy.pest
apetkov-so: I like it. It looks readable for simple expressions, and more complex ones can be wrapped in block expressions (#163).

jdygert-spok: I think I prefer always braces but that's my Rust showing.

apetkov-so: If you insist... `let x = if y { :x+1 } else { :0 }`

crates/aranya-policy-lang/src/lang/parse/policy.pest
```diff
@@ -152,16 +152,29 @@ at_least = { "at_least" ~ int_literal ~ fact_literal}
 at_most = { "at_most" ~ int_literal ~ fact_literal }
 // count facts, returning true if the number of facts found equals the expected number
 exactly = { "exactly" ~ int_literal ~ fact_literal }
+// Similar to the `match` statement, but resolves to a value. e.g. `let x = match { ... }`
+match_expression = { "match" ~ expression ~ "{" ~ match_expression_arm* ~ "}" }
+match_expression_arm = { (match_arm_expression | match_default) ~ "=>" ~ expression }
```
ssweetney-spok, line 157: Nit:Can we make it so that arms ending in a non-block expression must have a comma after the expression if there's another arm afterwards?

Otherwise you can write

```
        action foo(a int) {
                let x = match a {
                    1 => "one" 2 => false
                }
            }
```

Not a blocker so this can be resolved in a separate PR if needed

ssweetney-spok (approved):

crates/aranya-policy-lang/src/lang/parse/policy.pest
jdygert-spok: It's consistent with the rest of the language, I believe this is also legal. 

```
let z = { let x = 1 let y = 2 : x + y }
```

(I'd personally lean toward always mandatory commas and semicolons)

jdygert-spok (approved):

crates/aranya-policy-lang/src/lang/parse/policy.pest
ssweetney-spok: Huh, I never realized that. Maybe this is something that's better solved by a formatter?

apetkov-so: Ok. https://github.com/aranya-project/aranya-core/issues/185

PR #120, opened 2025-03-04
Add struct subselection

### LINKED ISSUE #117: Add Struct Subselection
Add the struct subselection operator, which converts one struct into another, smaller struct.

See [Policy Lang V2 spec](https://github.com/aranya-project/aranya-docs/blob/main/docs/policy-v2.md#struct-subselection)

ssweetney-spok:
Resolves #117

## Changes
- Adds new Instructions for getting and setting multiple fields in a struct, see https://github.com/aranya-project/aranya-core/pull/120#discussion_r1987959649
- New `aranya_policy_ast::Expression:` variant -> `aranya_policy_ast::Expression::Substruct`

### DIFF
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -819,6 +819,34 @@ impl<'a> CompileState<'a> {
                     })
                     .map_err(|e| self.err(e.into()))?
             }
+            Expression::Substruct(lhs, sub) => {
+                if !self.m.struct_defs.contains_key(sub) {
+                    return Err(self.err(CompileErrorType::NotDefined(format!(
+                        "Struct `{}` not defined",
+                        sub
+                    ))));
+                }
+
+                match self.compile_expression(lhs)? {
+                    Typeish::Type(VType::Struct(lhs_struct_name)) => {
+                        if !self.m.struct_defs.contains_key(&lhs_struct_name) {
+                            return Err(self.err(CompileErrorType::NotDefined(format!(
+                                "Struct `{lhs_struct_name}` is not defined",
+                            ))));
+                        }
+                    }
+                    Typeish::Type(_) => {
+                        return Err(self.err(CompileErrorType::InvalidType(
+                            "Expression to the left of the substruct operator is not a struct"
+                                .to_string(),
+                        )));
+                    }
+                    _ => {}
+                }
+
+                self.append_instruction(Instruction::Substruct(sub.clone()));
+                Typeish::Type(VType::Struct(sub.clone()))
+            }
             Expression::Add(a, b) | Expression::Subtract(a, b) => {
                 let left_type = self.compile_expression(a)?;
                 let right_type = self.compile_expression(b)?;
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -851,17 +851,6 @@ impl<'a> CompileState<'a> {
                                 lhs_struct_name,
                             )));
                         }
-
-                        // Check that the struct type on the RHS is a strict subset of the struct expression on the LHS
-                        if lhs_field_defns
-                            .iter()
-                            .all(|field_def| sub_field_defns.contains(field_def))
-                        {
-                            return Err(self.err(CompileErrorType::InvalidSubstruct(
-                                sub.clone(),
-                                lhs_struct_name,
-                            )));
-                        }
                     }
                     Typeish::Indeterminate => {}
                     Typeish::Type(_) => {
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -841,9 +841,21 @@ impl<'a> CompileState<'a> {
                             ))));
                         };
 
+                        // Check that the struct type on the RHS is a subset of the struct expression on the LHS
                         if !sub_field_defns
                             .iter()
                             .all(|field_def| lhs_field_defns.contains(field_def))
+                        {
+                            return Err(self.err(CompileErrorType::InvalidSubstruct(
+                                sub.clone(),
+                                lhs_struct_name,
+                            )));
+                        }
+
+                        // Check that the struct type on the RHS is a strict subset of the struct expression on the LHS
+                        if lhs_field_defns
+                            .iter()
+                            .all(|field_def| sub_field_defns.contains(field_def))
                         {
                             return Err(self.err(CompileErrorType::InvalidSubstruct(
                                 sub.clone(),
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -857,7 +857,6 @@ impl<'a> CompileState<'a> {
                     }
                 }
 
-                // self.append_instruction(Instruction::Substruct(sub.clone()));
                 let field_count = sub_field_defns.len();
                 for field in sub_field_defns {
                     self.append_instruction(Instruction::Const(Value::String(field.identifier)));
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -5,6 +5,7 @@ mod types;
 use std::{
     collections::{btree_map::Entry, BTreeMap, BTreeSet},
     fmt,
+    num::NonZeroUsize,
     ops::Range,
 };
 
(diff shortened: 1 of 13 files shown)


crates/aranya-policy-vm/src/machine.rs
```diff
@@ -676,6 +676,44 @@ where
...
+                    .struct_defs
+                    .get(&sub_struct)
+                    .ok_or_else(|| missing_struct_def(&sub_struct))?;
+
+                if !sub_struct_field_defns
+                    .iter()
+                    .all(|field_def| s_field_defs.contains(field_def))
+                {
+                    return Err(self.err(MachineErrorType::InvalidSubstruct));
+                }
```
jdygert-spok, line 704: I think this is something we can and should check at compile time.

jdygert-spok, line 704: I think this is something we can and should check at compile time.

ssweetney-spok:
Opening this up for feedback now but I'm waiting until PR #99 is merged into main and then I'll update the base branch

crates/aranya-policy-vm/src/machine.rs
ssweetney-spok: > We should be able to do the operation with gets and sets based on the struct definitions that the compiler already has

Yea, I was considering that at first. It seems like I'd have to clone the source struct n-1 (where n is equal to the number of fields in the substruct) times so that have it at the top of the stack for `StructGet` instructions

ssweetney-spok: These are nice. We could also use them for the [`as` operator](https://github.com/aranya-project/aranya-docs/blob/main/docs/policy-v2.md#isomorphic-struct-conversion).

ssweetney-spok: It's an interesting idea.

> My main issue with this is it turns stack access for anything below the top of the stack into a linear-time operation, which erases one of the main advantages of a stack machine. But if the depth we peek into the stack is limited, this might still work fine.

That's a good point. I suppose we could add a note/lint about keeping frequently accessed fields near the beginning of a Struct's defintion

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -822,6 +823,55 @@ impl<'a> CompileState<'a> {
...
+                    Typeish::Indeterminate => {}
+                    Typeish::Type(_) => {
+                        return Err(self.err(CompileErrorType::InvalidType(
+                            "Expression to the left of the substruct operator is not a struct"
+                                .to_string(),
+                        )));
+                    }
+                }
+
+                let field_count = sub_field_defns.len();
```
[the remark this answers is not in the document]
ssweetney-spok: I put it here because the for loop takes ownership of `sub_field_defns` so I can't use `sub_field_defns` after the for loop. The for loop doesn't need to take ownership but I wanted to avoid having to make a clone for each field defintion

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -822,6 +823,55 @@ impl<'a> CompileState<'a> {
...
+                for field in sub_field_defns {
+                    self.append_instruction(Instruction::Const(Value::String(field.identifier)));
+                }
+
+                if let Some(field_count) = NonZeroUsize::new(field_count) {
```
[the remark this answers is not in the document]
ssweetney-spok: > But separately, I don't understand the use of NonZeroUsize here. Isn't this equivalent to saying field_count != 0?

Yea, it's essentially the same thing. I defined the `MSet` and `MGet` variants like

```
    /// Add multiple members to the struct
    MStructSet(NonZeroUsize),
    /// Get multiple members from the struct
    MStructGet(NonZeroUsize),
```

It felt like a nice way to document that you should use them with positive integers since 0 would be a no-op. I can replace them with a regular usize though

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -822,6 +823,55 @@ impl<'a> CompileState<'a> {
...
+                    Typeish::Type(VType::Struct(lhs_struct_name)) => {
+                        let Some(lhs_field_defns) = self.m.struct_defs.get(&lhs_struct_name) else {
+                            return Err(self.err(CompileErrorType::NotDefined(format!(
+                                "Struct `{lhs_struct_name}` is not defined",
+                            ))));
+                        };
+
+                        if !sub_field_defns
+                            .iter()
+                            .all(|field_def| lhs_field_defns.contains(field_def))
```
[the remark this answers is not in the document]
ssweetney-spok: Yea lol, the recent discussion on this PR made me take another look at the code and then I realized 😆 

> Question still stands, though.

We could relax the definition but then the `as` operator becomes a less powerful version of the `substruct` operator 🤔

apetkov-so (approved):

crates/aranya-policy-vm/src/machine.rs
```diff
@@ -716,6 +716,54 @@ where
...
+                        return Err(self.err(MachineErrorType::InvalidStructMember(field_name)));
+                    };
+
+                    if !field_val.fits_type(&field_defn.field_type) {
+                        return Err(self.err(MachineErrorType::InvalidStructMember(field_name)));
+                    }
+
+                    target.fields.insert(field_name, field_val);
+                }
+                self.ipush(target)?;
```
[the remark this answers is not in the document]
ssweetney-spok: >You can do ipeek() to get a &mut Struct. That way you don't have to pop and push the target struct.

Oh nice, I didn't know about `ipeek()`. I did try it but the borrow checker doesn't like that since `self` is borrowed mutably there and on the next line `self.machine.struct_defs.get...`. I could do some refactoring to wrap `stack` and `machine` fields on the `RunState` struct with `RefCell` which should work

> Separately I wonder if it might be more efficient to allow getting references further down the stack...

Yea, that's sounds like it could be. I think there might be an issue with having references to stack items while simultaneously pushing/popping items to/from the stack?

crates/aranya-policy-compiler/src/compile.rs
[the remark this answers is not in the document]
ssweetney-spok: Are we in agreement that the subselection operator to work for 2 structs with identical fields?

cc: @apetkov-so @mooreoak @ycarmy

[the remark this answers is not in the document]
mooreoak: @ssweetney-spok  can you give an example of what you mean?

mooreoak (commented):
Is there a spec for this? I think I understand what it is supposed to be used for but I think I missed the discussion about the idea.

crates/aranya-policy-compiler/src/compile.rs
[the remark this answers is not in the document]
ssweetney-spok:
```
struct Foo {
    a int,
    b string,
    c bool,
}

struct Bar {
    a int,
    b string,
    c bool,
}

action test_action(x Bar) {
    let f = x substruct Foo
}
```

Should that work?

The spec for [struct subselction](https://github.com/aranya-project/aranya-docs/blob/main/docs/policy-v2.md#struct-subselection). In other words, should we drop the mention of "strict".

> A struct A whose fields are a *strict* subset of the fields of struct B can be assigned from struct B with the struct subselection operator substruct.

to

> A struct A whose fields are a subset of the fields of struct B can be assigned from struct B with the struct subselection operator substruct.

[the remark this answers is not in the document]
apetkov-so: I think the above should work. It may be of questionable use, but it's not an error.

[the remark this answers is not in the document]
ssweetney-spok: Cool, I've reverted the strict subset check. Here's [the PR](https://github.com/aranya-project/aranya-docs/pull/90) for updating the spec

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -822,6 +823,56 @@ impl<'a> CompileState<'a> {
                     })
                     .map_err(|e| self.err(e.into()))?
             }
+            Expression::Substruct(lhs, sub) => {
+                self.append_instruction(Instruction::StructNew(sub.clone()));
+
+                let Some(sub_field_defns) = self.m.struct_defs.get(sub).cloned() else {
+                    return Err(self.err(CompileErrorType::NotDefined(format!(
```
gknopf-aranya, line 830: The error code here and elsewhere seems like it could be simplified by defining these error types in an enum ahead of time. E..g.:
`return ErrorStructSubNotDefined;`

gknopf-aranya: I see that there is an enum defined, but it takes a lot of code like `Err` and `self.err` just to use it. Maybe a macro could help?

jdygert-spok: I've mentioned previously a possible option: inner functions return an unlocated error, and wrapping with location occurs only in the public API functions. Probably a significant refactor.

crates/aranya-policy-vm/tests/vm.rs
```diff
@@ -2322,3 +2322,152 @@ fn test_block_expression() -> anyhow::Result<()> {
...
+                KVPair::new("x", Value::Int(30)),
+                KVPair::new("y", Value::Bool(false)),
+            ]
+        )
+    );
+    Ok(())
+}
+
+#[test]
+fn test_substruct_errors() -> anyhow::Result<()> {
```
gknopf-aranya, line 2383: Separating these cases into different tests for each error would allow the test names to be more descriptive.
E.g. `test_substruct_too_small`, `test_substruct_missing_field`, etc.

apetkov-so (approved):

crates/aranya-policy-vm/tests/vm.rs
ssweetney-spok: That's a good point. I saw Aleko do this on some other PRs and I do prefer it since it reduces repetition. I could include the case name in the [panic message of `assert_eq`](https://doc.rust-lang.org/std/macro.assert_eq.html) which would do something similar. I'll do that in another PR

PR #159, opened 2025-03-13
(aranya-capi-codegen) Fix `Option<&T>` for fn args

### LINKED ISSUE #164: capi-codegen: support Option<&T> as a function arg
Note: supporting `Option<&Option<&T>>` is not required

ssweetney-spok:
Resolves #164


jdygert-spok:
Please add some "tests" to `crates/aranya-capi-codegen-test/src/defs.rs`, and check how nested option is handled (I think it will be rejected eariler?)

elagergren-spideroak:
what's broken?

ssweetney-spok:
> what's broken?

I'm trying to add a function that includes to the C client API and the corresponding Rust function has a signature like
```Rust
pub unsafe fn sync_now(
    client: &mut Client,
    team: &TeamId,
    addr: Addr,
    config: Option<&SyncPeerConfig>,
) -> Result<(), imp::Error>
```

The generated Rust code for that function looks like
```Rust
fn __tramp_aranya_sync_now(
    client: *mut AranyaClient,
    team: *const AranyaTeamId,
    addr: AranyaAddr,
    config: *const AranyaSyncPeerConfig,
) -> ::core::result::Result<
    ::core::result::Result<(), __imports::imp::Error>,
    __capi::InvalidArg<'static>,
> {
    #[allow(clippy::let_with_type_underscore)]
    let client: &mut _ = {
        let client = __capi::try_as_mut!(client);
        __capi::to_inner_mut!(client)
    };
    #[allow(clippy::let_with_type_underscore)]
    let team: &_ = {
        let team = __capi::try_as_ref!(team);
        __capi::to_inner_ref!(team)
    };
    #[allow(clippy::let_with_type_underscore)]
    let addr: _ = __capi::to_inner!(addr);
    let config: ::core::option::Option<&AranyaSyncPeerConfig> = __capi::try_as_opt!(
        config
    );
    #[allow(clippy::blocks_in_conditions)] #[allow(clippy::match_single_binding)]
    #[allow(unused_braces)]
    match unsafe { crate::api::defs::sync_now(client, team, addr, config) } {
        __pattern => ::core::result::Result::Ok(__pattern),
    }
}
```

Which results in this error

```
mismatched types
expected enum `Option<&defs::SyncPeerConfig>`
   found enum `Option<&__AranyaSyncPeerConfigFfiWrapper<defs::SyncPeerConfig, __AranyaDurationFfiWrapper<defs::Duration>, generated::AranyaSyncWhen>>`
```

ssweetney-spok:
So it looks like the issue is even more general since `&&T` doesn't work either. I could be wrong but it looks like it would take a non-trivial amount of work to support an arbitrary number of nested references 🤔

jdygert-spok (dismissed):
LGTM, would like at least a quick look from @elagergren-spideroak as well.

crates/aranya-capi-codegen/src/ast/expand.rs
```diff
@@ -683,14 +683,22 @@ impl Ast {
...
-                                let #name = #conv;
-                                #newtype
-                            };
-                        })
+
+                        let token_stream = match input.ty {
+                            Type::Option(_) => quote! {
+                                #[allow(clippy::let_with_type_underscore)]
+                                let #name: #ty =  #conv.map(|#name| #newtype);
+                            },
```
elagergren-spideroak, line 691: I think we can handle `Option` more "generically" and get rid of this case by adding the following to `internal/conv/newtype.rs`:

```rust
// SAFETY: `T` is `NewType`.
unsafe impl<T: NewType> NewType for Option<&T> {
    type Inner = Option<&T::Inner>;
}
```

The block should then expand to

```rust
let config: ::core::option::Option<&AranyaSyncPeerConfig> = {
    let config = __capi::try_as_opt!(config);
    __capi::to_inner_ref!(config)
};
```

which is

```rust
let config: ::core::option::Option<&AranyaSyncPeerConfig> = {
    let config: Option<&__AranyaSyncPeerConfigFfiWrapper<...>> = if config.is_null() {
        None
    } else {
        Some(...)
    };
    <&Hide<...> as ToInnerKind>::kind(&config).to_inner_ref(&config)
};
```

NB: This is all untested.

elagergren-spideroak: But to be clear: I'm fine with the current approach (for now) since it seemingly works and unblocks us.

gknopf-aranya:
@ssweetney-spok could you link this PR to the relevant GitHub issue so we can track it in the upcoming sprint?

ssweetney-spok:
> @ssweetney-spok could you link this PR to the relevant GitHub issue so we can track it in the upcoming sprint?

@gknopf-aranya Gotcha, I just created an [issue](https://github.com/aranya-project/aranya-core/issues/164)

crates/aranya-capi-codegen/src/ast/expand.rs
ssweetney-spok: I did try this but the generated Rust code stayed the same. I tried understanding how the `NewType` trait works, I'm not sure implementing the `NewType` trait alone will be sufficient. It seems like we'd still need to changes to "crates/aranya-capi-codegen/src/ast/expand.rs"

jdygert-spok: You do need `Type::Option(_) => Some(parse_quote!(#capi::to_inner!(#ident))),` in `unpack_newtype_glue`, but can get rid of the highlighted change.

On another note, I wonder if we can get rid of all the `to_inner_*!` variants and just use `to_inner!`

ssweetney-spok: Gotcha, that makes sense

ssweetney-spok: Oh I see what you mean. Yea, it does look like things could be simplified

jdygert-spok (dismissed):

jdygert-spok (approved):

elagergren-spideroak (approved):
thank you!

Issue #112, opened 2025-02-26
runtime: Make SyncRequester/SyncResponder Use A FSM


nikki-aranya:
We currently store the current state as a variable which means we have to manually check that we're in the state we expected. We should instead roll a proper Finite State Machine to make this less of a burden, as well as improve the actual state flow so that we don't have to manually "reset" it if we get some data we don't expect.

My first thought would be to use Zero-Sized Type Markers to achieve this, since invalid states are unrepresentable, with an optional enum to hide this complexity from an end user, i.e.
```rs
pub enum SyncResponder {
    New(SyncResponderInner<New>),
    ...
}

// Marker types
struct New;
struct Start;
struct Idle;

struct SyncResponderInner<State = New> {
    ...
    state: PhantomData<State>,
}

impl SyncResponderInner<Idle> {
    // These functions are only able to be called from a SyncResponderInner in this state,
    // enforced by the type system at compile time, since they're distinct types.
    // Functions should consume self in order to mutate it to a new state.
    pub fn poll(mut self) -> SyncResponderInner<Start> {
        // do stuff
        SyncResponderInner {
            state: PhantomData,
        }
    }
}
```

mooreoak:
I would be interested to see a more fleshed out example showing how to handle unexpected state transitions in input.

nikki-aranya:
> I would be interested to see a more fleshed out example showing how to handle unexpected state transitions in input.

Having written that short example of what I was thinking, I can already foresee some problems even in this new approach. I'd have to more completely model an entire sync sequence and how it should handle each function in every state, but the complexity is large enough that I don't think we can currently guarantee it behaves like we want it to.

For example, SyncResponder::push() getting called before SyncResponder::receive so we don't actually have a storage_id. We currently claim that situation is a `bug!()`, but with the current APIs there's nothing stopping an end user from doing so (it's all pub). Ideally, in tackling this we produce a diagram of the state flow that can be used in documentation.

nikki-aranya:
Part of this issue should be that we need to build out a proper polling loop; we only ever call poll once currently, so that should be part of modeling our sync procedure

PR #166, opened 2025-03-19
111: add max bytes and improve syncer api

benz-spideroak:


nikki-aranya (dismissed):
This was a less intrusive change into requester/responder than I thought, looks good

crates/aranya-runtime/src/sync/responder.rs
```diff
@@ -466,16 +467,35 @@ impl<A: Serialize + Clone> SyncResponder<A> {
 
             for command in &found {
                 let mut policy_length = 0;
+                let policy = if let Some(p) = command.policy() {
```
djohnson-spideroak, line 470: nit, could avoid policy_length being `mut` by returning a tuple in this `if let`?

```rust
let (policy, policy_len) = if let ... {
    (p, p.len())
} else {
...
}
```

crates/aranya-quic-syncer/src/lib.rs
```diff
@@ -408,7 +417,8 @@ where
             response_syncer.receive(SyncRequestMessage::SyncRequest {
                 session_id,
                 storage_id,
-                max_bytes: 0,
+                max_bytes: u64::MAX,
+                // max_bytes: subscription.remaining_bytes,
```
djohnson-spideroak, line 421: rm

crates/aranya-runtime/src/sync/responder.rs
```diff
@@ -466,16 +467,35 @@ impl<A: Serialize + Clone> SyncResponder<A> {
...
+                        bytes_to_add
+                            .try_into()
+                            .expect("bytes_to_add conversion to u64 failed"),
+                    )
+                    .assume("total_bytes mustn't overflow")?;
+                if total_bytes > self.remaining_bytes {
+                    break;
+                }
+
+                if policy_length > 0 {
```
jdygert-spok, line 493: nit: just extend unconditionally, it will be a no-op when empty.

jdygert-spok (dismissed):
LGTM, just some open nits

crates/aranya-quic-syncer/src/lib.rs
gknopf-aranya: @benz-spideroak can you update the PR with this feedback?

gknopf-aranya:
Overall lgtm. I wasn't able to find the rustdocs for the new max bytes parameter.
Let's make sure the new syncer API is documented so it's easy to update aranya code to use the new API.

jdygert-spok (approved):

djohnson-spideroak (dismissed):

nikki-aranya (dismissed):

crates/aranya-runtime/src/sync/responder.rs
```diff
@@ -465,17 +466,28 @@ impl<A: Serialize + Clone> SyncResponder<A> {
...
+                total_bytes = total_bytes
+                    .checked_add(
+                        bytes_to_add
+                            .try_into()
+                            .expect("bytes_to_add conversion to u64 failed"),
+                    )
+                    .assume("total_bytes mustn't overflow")?;
+                if total_bytes > self.remaining_bytes {
+                    break;
                 }
```
ssweetney-spok, line 485: Since the index is incremented above, `index = index.checked_add(1).assume("index + 1 mustn't overflow")?;`, would the command that caused the loop the terminate be skipped the next time `get_commands` is called?

crates/aranya-runtime/src/sync/responder.rs
```diff
@@ -465,17 +466,28 @@ impl<A: Serialize + Clone> SyncResponder<A> {
...
-                    command_data
-                        .extend_from_slice(policy)
-                        .ok()
-                        .assume("command_data is too large")?;
+                let policy = command.policy().unwrap_or_default();
+                let policy_length = policy.len();
+                let bytes = command.bytes();
+                let bytes_to_add = policy_length
+                    .checked_add(bytes.len())
+                    .and_then(|sum| sum.checked_add(size_of::<CommandMeta>()))
```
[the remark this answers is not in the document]
benz-spideroak: > This value will usually be bigger than the actual size of the serialized data right? 

It depends on the size of the command. 

> That's OK, I'm just wondering how precise we want/need to be

I'm not understanding what you're saying. We should be exact. `total_bytes` includes the length of the command, the policy, and command meta which is everything we add.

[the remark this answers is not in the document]
ssweetney-spok: > I'm not understanding what you're saying. We should be exact. total_bytes includes the length of the command, the policy, and command meta which is everything we add.

I meant that the sum of `total_bytes` is the number of bytes as they exist in memory. It looks that data is being serialized with postcard before being sent. I'm not 100% sure but I don't think the size of the serialized data is 1:1 with the in-memory size.

As an [example](https://postcard.jamesmunns.com/wire-format#maximum-encoded-length), a `u16` takes up 2 bytes in memory but may require 3 to encode it

crates/aranya-runtime/src/sync/responder.rs
```diff
@@ -465,17 +466,29 @@ impl<A: Serialize + Clone> SyncResponder<A> {
...
+                    .checked_add(
+                        bytes_to_add
+                            .try_into()
+                            .expect("bytes_to_add conversion to u64 failed"),
+                    )
+                    .assume("total_bytes mustn't overflow")?;
+                if total_bytes > self.remaining_bytes {
+                    all_commands_added = false;
+                    break;
                 }
```
ssweetney-spok, line 486: This fixed the issue with skipping but now I think you can get "stuck" if max_bytes isn't big enough to fit all commands in a segment? 

Say you start at a segment that has 44 commands and the 42nd command causes the condition above to be true. The index isn't incremented so it won't be skipped but now you'll start at the beginning of the same segment the next time `get_commands` is called.

benz-spideroak: That is possible. You'll only get stuck if max bytes is low enough that it's impossible to transfer a segment in one request. That can already happen if there are more commands in a segment than COMMAND_RESPONSE_MAX. See the comment on line 506.

ssweetney-spok: > That can already happen if there are more commands in a segment...

Yep, both problems seem like they could be solved by #112

crates/aranya-quic-syncer/src/lib.rs
```diff
@@ -197,22 +198,24 @@ where
...
+        peer_addr: SocketAddr,
         storage_id: GraphId,
+        max_bytes: u64,
     ) -> Result<usize, QuicSyncError> {
+        let client_state = self.client_state.clone();
+        let mut client = client_state.lock().await;
         let mut buffer = vec![0u8; MAX_SYNC_MESSAGE_SIZE];
         let mut received = 0;
-        let heads = self.remote_heads.entry(syncer.server_addr()).or_default();
+        let heads = self.remote_heads.entry(peer_addr).or_default();
```
gknopf-aranya, line 209: Is `peer_addr` the peer's client address or server address?

crates/aranya-quic-syncer/examples/quic_syncer.rs
```diff
@@ -56,18 +55,15 @@ fn main() {
...
     storage_id: GraphId,
     server_addr: SocketAddr,
 ) where
     EN: Engine,
     SP: StorageProvider,
     S: Sink<<EN as Engine>::Effect>,
 {
-    let sync_requester = SyncRequester::new(storage_id, &mut Rng::new(), server_addr);
-    let fut = syncer.sync(client, sync_requester, sink, storage_id);
+    let fut = syncer.sync(server_addr, storage_id, u64::MAX);
```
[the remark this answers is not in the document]
benz-spideroak: Why would that be helpful? This isn't a magic number. It's the largest possible number. If we want to add a default or use this value somewhere else then we can add a constant.

crates/aranya-quic-syncer/src/lib.rs
benz-spideroak: Server address.

gknopf-aranya (approved):

PR #172, opened 2025-03-20
runtime: check for multiple finalize commands in braid

### LINKED ISSUE #108: graph: finalization in weave
Finalization is one feature that contributes to Aranya's ability to scale to large graphs. A brief description of finalization was included in graph.md in the aranya-docs repo, snippit below:

### Finalization

If an event `A` has been finalized in some set of commands `S`, then all the commands which appear in the weave of `S` prior to `A` must be ancestors of `A`.
```alloy
pred finalized[final: one Command, commands: some Command] {
  weave[commands]
  all st: FactDB | {
    eventually st.event = final implies {
      before historically ancestor[st.event, final]
    }
  }
}
```

@elagergren-spideroak @mooreoak @jdygert-spok @chip-so @ycarmy what are your thoughts on a spec for this? Or do we know the functionality enough that we can keep this issue as implementation.

jdygert-spok:
Diff is noisy due to moving code but the braid algorithm didn't change, I just check to make sure there are never two finalize commands in the strand heap. Closes #108.


jdygert-spok:
Passes existing tests but needs tests with finalize commands.

crates/aranya-runtime/src/client.rs
```diff
@@ -309,60 +384,53 @@ pub fn braid<S: Storage>(
...
-        'location: for location in prior {
-            for other in &strands {
-                trace!("checking {}", other.next);
-                if (location.same_segment(other.next) && location.command <= other.next.command)
-                    || storage.is_ancestor(location, &other.segment)?
-                {
-                    trace!("found ancestor");
-                    continue 'location;
+        pub fn push(&mut self, strand: Strand<S>) -> Result<(), ParallelFinalize> {
+            if matches!(strand.key.0, Priority::Finalize) {
```
gknopf-aranya, line 405: We could use some documentation on why this check is performed here

jdygert-spok: Added some docs on the struct, field, method, and on `ClientError::ParallelFinalize`

crates/aranya-runtime/src/client.rs
```diff
@@ -28,6 +26,15 @@ pub enum ClientError {
     NotAuthorized,
     #[error("session deserialize error: {0}")]
     SessionDeserialize(#[from] postcard::Error),
+    /// Attempted to braid two parallel finalize commands together.
+    ///
+    /// Policy must be designed such that two parallel finalize commands are never produced.
```
gknopf-aranya, line 31: Does this mean all the finalize commands must be generated by the same device?

crates/aranya-runtime/src/client.rs
```diff
@@ -28,6 +26,15 @@ pub enum ClientError {
...
+    ///
+    /// Currently, this is practically an unrecoverable error. You must wipe all graphs containing
+    /// the "bad" finalize command and resync from the "good" clients. Otherwise, your network will
```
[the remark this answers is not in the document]
gknopf-aranya: If there is no path to recovery planned, we should at least pass an error up through the MVP API to notify the application that an unrecoverable error occurred.

gknopf-aranya (dismissed):
lgtm. Are we planning to merge this PR first before testing the finalize command in policy, or do we have another way to verify these changes will work for adding finalization commands to the policy?

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -916,4 +933,44 @@ mod test {
...
+    fn test_finalize_success() {
+        let mut gb = graph! {
+            ClientState::new(SeqEngine, MemStorageProvider::new());
+            "a";
+            commit;
+            "a" < "b" "c" "d" "e" "f" "g";
+            "d" < "h" "i" "j";
+            "e" < finalize "fff1";
+        };
+        gb.finalize(mkid("e"), mkid("fff1")).unwrap();
```
ssweetney-spok, line 947: Duplicate command?

ssweetney-spok (commented):
It's my first time looking at the DSL for building a graph. Very clever stuff

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -897,4 +933,44 @@ mod test {
...
+        let mut gb = graph! {
+            ClientState::new(SeqEngine, MemStorageProvider::new());
+            "a";
+            commit;
+            "a" < "b" "c" "d" "e" "f" "g";
+            "d" < "h" "i" "j";
+            "e" < finalize "fff1";
+            "i" < finalize "fff2";
+        };
+        let err = gb.commit().expect_err("merge should fail");
```
gknopf-aranya, line 973: Do we want to fail during the merge even if no `gb.finalize()` happens later?

crates/aranya-runtime/src/client/transaction.rs
jdygert-spok: Oops yeah.

crates/aranya-runtime/src/client/transaction.rs
jdygert-spok: Not sure what you're asking. There's two finalize commands in the graph next to each other. I should add a test for one after the other though, which is valid.

gknopf-aranya: I wasn't sure if finalization happens right when a finalize command is merged into the graph or if it happens when the `gb.finalize()` method is invoked on the graph. If it happens when the finalization commands are merged into the graph, then ignore this comment.

jdygert-spok:
> lgtm. Are we planning to merge this PR first before testing the finalize command in policy, or do we have another way to verify these changes will work for adding finalization commands to the policy?

Finalization is just a priority value which affects braiding, so transaction level tests are all we really need here.

crates/aranya-runtime/src/client/transaction.rs
jdygert-spok: The dsl above creates a transaction with two finalize commands in branches, then when we `gb.commit()` it tries to braid them.

ssweetney-spok (approved):
🚀

crates/aranya-runtime/src/client.rs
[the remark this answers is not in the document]
jdygert-spok: Right now it's just "don't do that". Note there's no way to know which of the commands is the "good" one without external information.

gknopf-aranya (dismissed):

djohnson-spideroak (approved):

gknopf-aranya (approved):

ssweetney-spok (approved):

crates/aranya-runtime/src/client/braiding.rs
```diff
@@ -0,0 +1,261 @@
...
+            }
+            Some(strand)
+        }
+
+        /// Pops last strand when only one strand remains.
+        pub fn lone(&mut self) -> Option<Strand<S>> {
+            if self.heap.len() != 1 {
+                return None;
+            }
+            self.heap.pop()
```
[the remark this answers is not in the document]
jdygert-spok: Not currently since it is finished at that point, but good to add anyway for later possible changes.

Issue #175, opened 2025-03-20
rustfmt: Add Comments + Style Edition


nikki-aranya:
As in [Aranya](https://github.com/aranya-project/aranya/pull/139), we should add style_edition = "2024", wrap_comments = true, comment_width = 100 (with overrides for capi crates at 65), and format_code_in_doc_comments = true.

PR #177, opened 2025-03-21
Enforce `wrap_comments` on policy crates

nikki-aranya:
This adds a rustfmt override to all policy crates and includes all formatting changes. The goal is to open separate PRs addressing the other crates, and at the end a single commit that unifies all rustfmt.toml back into root, once we've approved all changes.

### DIFF
--- a/crates/aranya-policy-ast/rustfmt.toml
+++ b/crates/aranya-policy-ast/rustfmt.toml
@@ -1,0 +1,7 @@
+unstable_features = true
+group_imports = "StdExternalCrate"
+imports_granularity = "Crate"
+format_generated_files = false
+wrap_comments = true
+comment_width = 100
+format_code_in_doc_comments = true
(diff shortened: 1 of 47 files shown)


djohnson-spideroak (approved):
LGTM, I dont have strong opinions on when to wrap comments, but seems fine set at 100.

nikki-aranya:
100 is the same as our/the default code wrapping length so it fits a lot better

elagergren-spideroak (changes requested):
So, I'm probably fine with `wrap_comments`, but is this something we really need to spend time on before MVP?

djohnson-spideroak:
> So, I'm probably fine with `wrap_comments`, but is this something we really need to spend time on before MVP?

Agreed, I think we should finish out the PR you currently have open and then focus on the MVP tasks. The AFC feature flag is more important than changing the style of comments here.

elagergren-spideroak:
> > So, I'm probably fine with `wrap_comments`, but is this something we really need to spend time on before MVP?
> 
> Agreed, I think we should finish out the PR you currently have open and then focus on the MVP tasks. The AFC feature flag is more important than changing the style of comments here.

I don't even think we should finish this PR if it'll cause merge conflicts.

nikki-aranya:
This PR is now very outdated; I'll revisit when we have more bandwidth. See issue https://github.com/aranya-project/aranya-core/issues/175 for more details.

PR #180, opened 2025-03-24
aranya-capi-codegen: Apply `#[cfg()]` Attributes To FFI Wrapper

nikki-aranya:
Since we're making AFC experimental, some types are now conditionally defined, which causes errors in our FFI wrapper since the type path doesn't exist. This should fix that by copying the same `#[cfg()]` flags over.

### DIFF
--- a/crates/aranya-capi-codegen/src/ast/expand.rs
+++ b/crates/aranya-capi-codegen/src/ast/expand.rs
@@ -1646,12 +1646,14 @@ fn ffi_wrapper(ctx: &Ctx, strukt: &Struct, underlying: &Path) -> TokenStream {
 
         #[repr(transparent)]
         #[derive(Debug)]
+        #(#attrs)*
         pub struct #wrapper<#generics> {
             pub inner: #inner,
             #(#fields : ::core::marker::PhantomData<#fields>),*
         }
 
         #[automatically_derived]
+        #(#attrs)*
         impl<#generics> #capi::InitDefault for #wrapper<#generics>
         where
             #inner: #capi::InitDefault,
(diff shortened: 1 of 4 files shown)


crates/aranya-capi-codegen/src/ast/expand.rs
```diff
@@ -315,6 +315,12 @@ impl Ast {
         let underlying = ctx.defs.join(old.clone());
         strukt.ident = strukt.ident.with_prefix(&ctx.ty_prefix);
         let name = &strukt.ident;
+        let attrs = strukt
```
elagergren-spideroak, line 318: `ffi_wrapper` should do this, it's already passed the struct.

nikki-aranya: I need a way to ignore #[cfg(cbindgen)], I think. That's why I cloned it and it's all the way above anything that adds to attrs.

elagergren-spideroak:
```rust
if cfg.path().is_ident("cfg") {
}

Comment (elagergren-spideroak) [reply crates/aranya-capi-codegen/src/ast/expand.rs:318]:
```rust
if cfg.path().is_ident("cfg") {
    if cfg.parse_args::<Ident>().is_some_and(|ident| ident == "cbindgen") { /* skip */ }
}
```

crates/aranya-capi-codegen/src/ast/expand.rs
```diff
@@ -1623,6 +1623,11 @@ fn ffi_wrapper(ctx: &Ctx, strukt: &Struct, underlying: &Path) -> TokenStream {
     let fields = (0..fields.len())
         .map(|i| format_ident!("_{i}"))
         .collect::<Vec<_>>();
+    let attrs = strukt
+        .attrs
+        .iter()
+        .filter(|attr| attr.path().is_ident("cfg"))
```
elagergren-spideroak, line 1629: do we also need `cfg_attr`?

jdygert-spok:
Test on your `aranya` PR with `patch` and/or add header generation to the test crate.

crates/aranya-capi-codegen/src/ast/expand.rs
nikki-aranya: We don't actually use it anywhere but I guess it couldn't hurt

gknopf-aranya (approved):

elagergren-spideroak (approved):

PR #192, opened 2025-03-31
Close #186: Implement struct field insertion.

### LINKED ISSUE #186: Add Struct Definition Field Insertion
Define struct in terms of other structs. E.g.
```
struct Foo {
    a int
}
struct Bar {
    +Foo
    b string
}
```
See [spec](https://github.com/aranya-project/aranya-docs/blob/main/docs/policy-v2.md#struct-definition-field-insertion)

apetkov-so:
Allow defining structs in terms of other structs.
```
struct Foo { x int }
struct Bar { +Foo, y int } // => struct Bar { x int, y int }
```

### DIFF
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -2190,7 +2190,6 @@ impl<'a> CompileState<'a> {
         // define the structs provided by FFI schema
         for ffi_mod in self.ffi_modules {
             for s in ffi_mod.structs {
-                // TODO(apetkov) implement field insertion for FFI structs
                 let fields: Vec<StructItem<FieldDefinition>> = s
                     .fields
                     .iter()
@@ -2213,7 +2212,6 @@ impl<'a> CompileState<'a> {
         for fact in &self.policy.facts {
             let FactDefinition { key, value, .. } = &fact.inner;
 
-            // TODO(apetkov) implement field insertion for facts
             let fields: Vec<StructItem<FieldDefinition>> = key
                 .iter()
                 .chain(value.iter())
@@ -2227,17 +2225,7 @@ impl<'a> CompileState<'a> {
 
         // Define command structs before compiling functions
         for command in &self.policy.commands {
-            self.define_struct(
-                &command.identifier,
-                &command
-                    .fields
-                    .iter()
-                    .map(|si| match si {
-                        StructItem::Field(fd) => StructItem::Field(fd.clone()),
-                        StructItem::StructRef(s) => StructItem::StructRef(s.clone()),
-                    })
-                    .collect::<Vec<_>>(),
-            )?;
+            self.define_struct(&command.identifier, &command.fields)?;
         }
 
         // Define the finish function signatures before compiling them, so that they can be
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -187,8 +187,8 @@ impl<'a> CompileState<'a> {
     /// Insert a struct definition while preventing duplicates of the struct name and fields
     pub fn define_struct(
         &mut self,
-        identifier: Identifier,
-        items: &[StructItem],
+        identifier: &str,
+        items: &[StructItem<FieldDefinition>],
     ) -> Result<(), CompileError> {
         if self.m.struct_defs.contains_key(identifier) {
             return Err(self.err(CompileErrorType::AlreadyDefined(identifier.to_string())));
@@ -2161,26 +2161,31 @@ impl<'a> CompileState<'a> {
             self.compile_global_let(global_let)?;
         }
 
-        for effect in &self.policy.effects {
-            // TODO(apetkov) implement field insertion for effects
-            let fields: Vec<StructItem> = effect
-                .inner
-                .fields
-                .iter()
-                .map(|f| StructItem::Field(f.into()))
-                .collect();
-            self.define_struct(&effect.inner.identifier, &fields)?;
-        }
-
         for struct_def in &self.policy.structs {
             self.define_struct(&struct_def.inner.identifier, &struct_def.inner.items)?;
+        }
+
+        for effect in &self.policy.effects {
+            let fields: Vec<StructItem<FieldDefinition>> = effect
+                .inner
+                .items
+                .iter()
+                .map(|i| match i {
+                    StructItem::Field(f) => StructItem::Field(FieldDefinition {
+                        identifier: f.identifier.clone(),
+                        field_type: f.field_type.clone(),
+                    }),
+                    StructItem::StructRef(s) => StructItem::StructRef(s.clone()),
+                })
+                .collect();
+            self.define_struct(&effect.inner.identifier, &fields)?;
         }
 
         // define the structs provided by FFI schema
         for ffi_mod in self.ffi_modules {
             for s in ffi_mod.structs {
                 // TODO(apetkov) implement field insertion for FFI structs
-                let fields: Vec<StructItem> = s
+                let fields: Vec<StructItem<FieldDefinition>> = s
                     .fields
                     .iter()
                     .map(|a| {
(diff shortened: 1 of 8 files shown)


crates/aranya-policy-lang/src/lang/parse/policy.pest
```diff
@@ -42,9 +42,11 @@ vtype = _{ string_t | bytes_t | int_t | bool_t | id_t | struct_t | enum_t | opti
 // A field definition is an identifier followed by a type. Used in
 // struct definitions, function arguments, etc.
 field_definition = { identifier ~ vtype }
+// struct field insertion. Copy fields from another struct
+struct_field_insertion = { "+" ~ identifier }
 // A field definition list is a list of field definitions. Trailing
 // commas are allowed in any list.
-field_definition_list = _{ field_definition ~ ("," ~ field_definition)* ~ ","? }
+field_definition_list = _{ (field_definition | struct_field_insertion) ~ ("," ~ (field_definition | struct_field_insertion))* ~ ","? }
```
apetkov-so, line 49: Just realized this is not the right place to add field insertion, because this rule is used in a number for places that should not allow that, e.g. fact signatures, and function args.

apetkov-so: Pushed commit to resolve this.

crates/aranya-policy-ifgen-build/src/imp.rs
```diff
@@ -15,8 +15,14 @@ pub fn generate_code(policy: &Policy) -> String {
...
             let name = mk_ident(&s.identifier);
-            let names = s.fields.iter().map(|f| mk_ident(&f.identifier));
-            let types = s.fields.iter().map(|f| vtype_to_rtype(&f.field_type));
+            let names = s.items.iter().map(|i| match i {
+                StructItem::Field(f) => mk_ident(&f.identifier),
+                StructItem::StructRef(_) => todo!(),
+            });
+            let types = s.items.iter().map(|i| match i {
+                StructItem::Field(f) => vtype_to_rtype(&f.field_type),
+                StructItem::StructRef(_) => todo!(),
```
apetkov-so, line 24: @jdygert-spok I'm not sure what to do here... This runs before the compiler has had a chance to resolve the struct references to fields, and we don't have access to the Module, so we can't look up the struct ourselves.

apetkov-so: Actually, we can look up the struct in the policy (`policy.structs`). So, would we have to duplicate the compiler's field insertion logic? Then remove the struct reference from the policy struct so the compiler doesn't try to insert the fields a second time?

ssweetney-spok (commented):
Looks good so far. Will do another pass later

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2066,15 +2110,31 @@ impl<'a> CompileState<'a> {
...
         for command in &self.policy.commands {
-            self.define_struct(&command.identifier, &command.fields)?;
+            self.define_struct(
+                &command.identifier,
+                &command
+                    .fields
+                    .iter()
+                    .map(|si| match si {
+                        StructItem::Field(fd) => StructItem::Field(fd.clone()),
+                        StructItem::StructRef(s) => StructItem::StructRef(s.clone()),
```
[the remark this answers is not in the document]
apetkov-so: Yeah, the whole `iter`, `map` thing is unnecessary... Removed.

crates/aranya-policy-derive/src/ffi.rs
```diff
@@ -72,17 +77,26 @@ pub(crate) fn parse(attr: TokenStream, item: TokenStream) -> syn::Result<TokenSt
...
+                StructItem::Field(d) => format_ident!("__field_{}", d.identifier),
+                StructItem::StructRef(_) => todo!(),
+            })
             .collect::<Vec<_>>();
-        let types = d.fields.iter().map(|d| {
-            let vtype = TypeTokens::new(&d.field_type, &alloc, &crypto, &vm);
+        let types = d.items.iter().map(|d| {
+            let vtype = match d {
+                StructItem::Field(f) => TypeTokens::new(&f.field_type, &alloc, &crypto, &vm),
+                StructItem::StructRef(_) => todo!(),
```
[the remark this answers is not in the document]
jdygert-spok: Ideally these would be syn errors instead.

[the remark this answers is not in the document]
apetkov-so: I added message to the `todo!`... Can't get the syntax right when returning an error.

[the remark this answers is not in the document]
jdygert-spok: The first is todo since you can trigger it with `+Foo` in the ffi derive. The others are technically unreachable since they check the same thing but I think todo is more accurate.

crates/aranya-policy-ifgen-build/src/imp.rs
```diff
@@ -15,8 +15,20 @@ pub fn generate_code(policy: &Policy) -> String {
...
+                        .identifier,
+                )
+            });
+            let types = s.items.iter().map(|i| {
+                vtype_to_rtype(
+                    &i.field()
+                        .expect("should not be a struct ref here")
+                        .field_type,
+                )
+            });
```
jdygert-spok, line 31: This means ifgen will fail on any policy containing field insertion :(

This is more important than the ffi derive since we generally expect users to use ifgen on their full policy.

Not sure if this should be done here or a follow-up PR.

We don't really want to have to re-implement logic on stuff like this here. I guess we could compile and take defs from the machine, but that does a lot of extra work :thinking:

ssweetney-spok: We could add a pre-processing step that handles the field insertions and transforms the policy AST. The compiler loses would lose some information about where the field came from though 🤔

ssweetney-spok:
> IMO it is generally worse than having a struct as a field, except perhaps considering optimizations.

How so? It seems like having flatter structs is better for all of the [struct conversion operations](https://github.com/aranya-project/aranya-docs/blob/main/docs/policy-v2.md#struct-conversion) we're adding in v2

> TBH I'm not convinced this feature pulls its weight in complexity

I kind of agree. I mean users can always copy and paste 🤷

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1891,21 +1916,30 @@ impl<'a> CompileState<'a> {
...
+                    map.insert(f.identifier.clone(), f.field_type.clone());
+                }
+                StructItem::StructRef(ref_name) => {
+                    let struct_def =
+                        self.m.struct_defs.get(ref_name).ok_or_else(|| {
+                            self.err(CompileErrorType::NotDefined(ref_name.clone()))
+                        })?;
+                    for fd in struct_def {
+                        map.insert(fd.identifier.clone(), fd.field_type.clone());
+                    }
```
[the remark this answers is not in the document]
apetkov-so: It's technically not necessary, because commands are first [defined as structs](https://github.com/aranya-project/aranya-core/pull/192/files#diff-e2b429adfc01f0f40d066e326bd3bb29a8f43a41819c8ceae5b44de1d60e6c31R2199
), before being [compiled](https://github.com/aranya-project/aranya-core/pull/192/files#diff-e2b429adfc01f0f40d066e326bd3bb29a8f43a41819c8ceae5b44de1d60e6c31R2224). The first pass catches duplicates -  we have a test for this.

crates/aranya-policy-ifgen-build/src/imp.rs
apetkov-so: Compiling (and passing a `Module` to `generate_code` instead of a `Policy`) seems like the most correct way to implement this, but I ran into some issues. The module doesn't have effects - they get compiled into structs.

apetkov-so: Here's my attempt at compiling the policy: https://github.com/aranya-project/aranya-core/pull/233

apetkov-so:
@jdygert-spok Do you feel this PR is useful as-is? Should we merge it and continue working on the "-compile" branch, or wait until that work is merged into this one?

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -188,29 +188,50 @@ impl<'a> CompileState<'a> {
...
-                        ));
+        // Add explicitly-defined fields and those from struct insertions
+
+        let mut field_definitions = Vec::new();
+        for item in items {
+            match item {
+                StructItem::Field(field) => {
+                    if field_definitions
+                        .iter()
+                        .any(|f: &FieldDefinition| f.identifier == field.identifier)
```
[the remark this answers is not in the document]
apetkov-so: True, but there are no expensive operations in the loop, just some hashtable lookups and insertions. Plus, it's at compile time.

jdygert-spok (approved):

PR #195, opened 2025-04-01
(capi-codegen) Enum return types

### LINKED ISSUE #160: capi-codegen: support enums as return types
They need to impl `ByMutPtr` and/or some glue.

ssweetney-spok:
Closes #160

## Changes
- Convert to a "FFI enum" using it's `From` impl when a trampoline function returns the underlying enum type
- When there's an output param, try to convert to a "FFI enum" from it's underlying type by calling `Into::into`
- Add `ByMutPtr` impl for "FFI enums"

### DIFF
--- a/crates/aranya-capi-codegen-test/src/defs.rs
+++ b/crates/aranya-capi-codegen-test/src/defs.rs
@@ -99,11 +99,14 @@ pub fn test_enum_unit(_a: Enum) {}
 pub fn test_unit_enum() -> Enum {
     Enum::A
 }
-// pub fn test_enum_enum(a: Enum) -> Enum {
-//     a
-// }
-// pub fn test_enum_result_enum_error(a: Enum) -> Result<Enum, crate::Error> {
-//     Ok(a)
+pub fn test_enum_enum(a: Enum) -> Enum {
+    a
+}
+pub fn test_enum_result_enum_error(a: Enum) -> Result<Enum, crate::Error> {
+    Ok(a)
+}
+// pub fn test_mut_ref_enum_enum(a: &mut Enum) -> Enum {
+//     a.clone()
 // }
 
 #[derive(Copy, Clone, Debug, Default)]
--- a/crates/aranya-capi-codegen-test/src/defs.rs
+++ b/crates/aranya-capi-codegen-test/src/defs.rs
@@ -96,9 +96,9 @@ pub enum Enum {
 }
 
 pub fn test_enum_unit(_a: Enum) {}
-// pub fn test_unit_enum() -> Enum {
-//     Enum::A
-// }
+pub fn test_unit_enum() -> Enum {
+    Enum::A
+}
 // pub fn test_enum_enum(a: Enum) -> Enum {
 //     a
 // }
--- a/crates/aranya-capi-codegen/src/ast/expand.rs
+++ b/crates/aranya-capi-codegen/src/ast/expand.rs
@@ -240,6 +240,14 @@ impl Ast {
                 unsafe impl #capi::types::ByValue for #name {}
             });
 
+            items.push(parse_quote! {
+                /// SAFETY: The type is a unit-only enumeration
+                /// with a `#[repr(...)]`, and we check for
+                /// invalid representations, so it is FFI safe.
+                #[automatically_derived]
+                unsafe impl #capi::types::ByMutPtr for #name {}
+            });
+
             // Forward `From` impls needed to implement
             // `ErrorCode`.
             items.push(parse_quote! {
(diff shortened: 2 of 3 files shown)


gknopf-aranya:
Looks like there's a clippy check to resolve. Try running this locally:
`cargo make clippy`

ssweetney-spok:
> Looks like there's a clippy check to resolve. Try running this locally: `cargo make clippy`

The lint warnings are in auto generated file. It'd be better to disable the specific lint it's complaining about. We already do that in a few places

crates/aranya-capi-codegen-test/src/defs.rs
```diff
@@ -96,14 +96,17 @@ pub enum Enum {
...
+}
+pub fn test_enum_enum(a: Enum) -> Enum {
+    a
+}
+pub fn test_enum_result_enum_error(a: Enum) -> Result<Enum, crate::Error> {
+    Ok(a)
+}
+// pub fn test_mut_ref_enum_enum(a: &mut Enum) -> Enum {
+//     a.clone()
 // }
```
ssweetney-spok, line 110: @gknopf-aranya This doesn't work currently but do you need a function like this? To be more specific, what doesn't work is taking a mutable reference to an enum as a function argument

Or are the examples in `test_unit_enum` and `test_enum_result_enum_error` enough to unblock you for implementing the C API for Fact DB queries?

gknopf-aranya: This is the function signature we're implementing from the MVP spec:
`QueryDeviceRole(team_id, device_id) -> Role`
It's one of the unchecked boxes here:
https://github.com/aranya-project/aranya/pull/136
And mentioned in this MVP spec updates PR:
https://github.com/aranya-project/aranya-docs/pull/81

I don't think we need mutable reference enum arg support to unblock the API I'm wanting to implement for MVP.
Just to be able to return an enum.
That's assuming the C API doesn't translate an enum return into a mutable arg reference. E.g. like how returning strings works in the C API.

ssweetney-spok: > That's assuming the C API doesn't translate an enum return into a mutable arg reference. E.g. like how returning strings works in the C API.

It could depending on the function args. That's actually what happens for [test_enum_enum](https://github.com/aranya-project/aranya-core/pull/195/files#diff-6748683d7c424770c197ccd2573b118feaa6e00918a8be1c14249e55e7cc4983R1029-R1032) but that's OK since we only care about writing valid data back using that pointer.

So I think we're good to go then

gknopf-aranya: @ssweetney-spok could you create an issue for the last test case that hasn't been implemented yet for enum support?

gknopf-aranya (approved):
lgtm! Would like @jdygert-spok and/or @elagergren-spideroak to review this since they are our capi generator experts

crates/aranya-capi-codegen-test/src/defs.rs
ssweetney-spok: Created one [here](https://github.com/aranya-project/aranya-core/issues/198)

crates/aranya-capi-codegen/src/ast/expand.rs
```diff
@@ -1490,9 +1500,25 @@ fn unpack_newtype_glue(ctx: &Ctx, arg: &ExpandedArg) -> Option<Expr> {
...
+            let new_name: &Ident = named.path.ty_name();
+            let old_name = idents.get_old(new_name).expect("unknown type");
+            if let Some(Node::Enum(_)) = types.get(old_name) {
+                return Some(quote! {
+                    #new_name::from(#ident)
+                });
+            } else {
+                (quote!(from_inner), named)
+            }
+        }
```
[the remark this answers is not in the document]
gknopf-aranya: If you're okay with it to unblock the enum return API method in the MVP API, I'm good with it.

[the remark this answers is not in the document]
gknopf-aranya: Should we add a todo here or open an issue to clean this up later?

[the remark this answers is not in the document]
jdygert-spok: Nah it's fine IMO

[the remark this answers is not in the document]
ssweetney-spok: > I don't love this special casing and lookups required but it works 🤷
> 
> We can revisit if we ever get around to a general cleanup.

I'm not a fan either which is why it took me a while to try coming up with a general solution before settling with this...

I did consider adding a `NewType` impl for the "FFIEnum" and "RustEnum" but then I realized that could break type safety (passing in an integer that doesn't match any discrimant values in the Rust enum). It seems like a similar trait for fallible conversions would be useful. Something like `TryNewType`?

PR #203, opened 2025-04-07
Expose Graph IDs via StorageProvider

nikki-aranya:
Adds a new method to StorageProvider and IoManager to allow listing existing Graph IDs. This includes implementing `fdopendir`, `readdir`, and `closedir` methods in `aranya-libc`.


nikki-aranya:
I'm open to bikeshedding wrt `list` and `list_graph_ids`, those were just the most logical. list on IoManager I think is fine since it matches "open" and "create", but eh.

nikki-aranya:
I also need to implement a test to make sure this works and doesn't break in the future, but should be good for an initial review.

gknopf-aranya:
Seems like we need a test with at least 2-3 graphs to demonstrate listing the graph IDs works for more than 1 graph.

crates/aranya-libc/src/sys/unix.rs
```diff
@@ -83,3 +89,41 @@ pub fn fsync(fd: BorrowedFd<'_>) -> Result<(), Errno> {
...
+        crate::errno::clear_errno();
+    }
+    // SAFETY: FFI call, no invariants.
+    let ret = unsafe { libc::readdir(dir) };
+    if ret.is_null() {
+        Err(errno())
+    } else {
+        Ok(ret)
+    }
+}
```
jdygert-spok, line 118: This should check for `errno == 0`

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -15,6 +15,36 @@ use crate::{
...
+
+impl Iterator for GraphIdIterator {
+    type Item = Result<GraphId, StorageError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // Loop until we find an entry that isn't "." or ".."
+        loop {
+            let Ok(entry) = libc::readdir(&mut self.inner) else {
+                return None;
+            };
```
jdygert-spok, line 38: This should propagate errors.

crates/aranya-runtime/src/storage/linear/libc/tests.rs
```diff
@@ -25,6 +27,59 @@ impl StorageBackend for LinearBackend {
...
+
+    let init_cmd2 = Init([2u8; 64].into());
+    let mut fp2 = provider.new_perspective(PolicyId::new(1));
+    let _ = fp2.add_command(&init_cmd2);
+    let _ = provider.new_storage(fp2).unwrap();
+
+    let mut graph_ids = provider
+        .list_graph_ids()
+        .unwrap()
+        .filter_map(Result::ok)
```
jdygert-spok, line 74:
```suggestion
        .map(Result::unwrap)
```

We shouldn't see any Errs in this test

gknopf-aranya:
@jdygert-spok, @nikki-aranya do we want to test this feature branch against this PR before merging?
https://github.com/aranya-project/aranya/pull/171

I'd rather make sure it works for the intended use-case before releasing new aranya-core crates.
Unless we're really confident this will work on the first try based on testing added to this PR.

gknopf-aranya (dismissed):
lgtm. Just waiting to hear decision regarding whether to test this feature branch against the net identifier queries PR or if unit tests are enough.

crates/aranya-runtime/src/storage/linear/libc/tests.rs
```diff
@@ -25,6 +27,59 @@ impl StorageBackend for LinearBackend {
...
+        .list_graph_ids()
+        .unwrap()
+        .map(Result::unwrap)
+        .collect::<Vec<_>>();
+    graph_ids.sort();
+    assert_eq!(
+        graph_ids,
+        vec![GraphId::from([1u8; 64]), GraphId::from([2u8; 64])]
+    );
+}
```
jdygert-spok, line 81: We should move this to the test suite so it is tested for all impls.

jdygert-spok: Ideally we should write actual dsl tests that create multiple graphs but that can be later. This can be added manually into the test_suite for now.

nikki-aranya: I don't think there's a way to do that? This requires actual code, and the .test stuff is just json which doesn't let me create storage and stuff

crates/aranya-runtime/src/storage/linear/testing.rs
```diff
@@ -22,6 +22,10 @@ impl io::IoManager for Manager {
     fn open(&mut self, _id: GraphId) -> Result<Option<Self::Writer>, StorageError> {
         Ok(None)
     }
+
+    fn list(&self) -> Result<impl Iterator<Item = Result<GraphId, StorageError>>, StorageError> {
+        Ok(core::iter::empty::<_>())
+    }
```
jdygert-spok, line 28: This will fail the test. Guess this manager isn't quite implemented right, especially if we decide to make the linear storage provider not hold onto all writers.

nikki-aranya: I'm not sure why it exists, I was only testing the actual LinearStorage since that requires accessing a filesystem, this can stay stubbed for whatever it needs to be used for

gknopf-aranya:
There are few CI lint checks to resolve

crates/aranya-runtime/src/testing/testdata/generate_graph.test
jdygert-spok, line 1: @benz-spideroak looks like you previously removed the generate_graph test in dsl.rs but not the generate_graph test data file. It currently does not pass and is the only use of `GenerateGraph`. Do we want to fully remove it or get it working?

https://github.com/aranya-project/aranya-core/commit/4aff85f2289877ce313d5c38c267809cd5d8dd21#diff-4eba8f1355e9072078e4374dcf8892c5c4401f323ce7641adbdfb824ba266311L839

crates/aranya-runtime/src/testing/dsl.rs
```diff
@@ -524,6 +561,28 @@ where
...
+            } => {
+                let state = clients
+                    .get_mut(&client)
+                    .ok_or(TestError::MissingClient)?
+                    .get_mut();
+
+                let policy_data = policy.to_be_bytes();
+
+                let cmd_id = CommandId::from([fill; 64]);
+                let init_action = TestActions::InitWithId(cmd_id);
```
jdygert-spok, line 578: We shouldn't need this `NewGraphWithFill` or to hard code IDs, we can just use the "short IDs" instead in `VerifyGraphIds` (looked up in the `graphs` map).

We just need to pass in the policy as the nonce above in `TestAction::Init` to be able to create multiple graphs via `NewGraph`.

crates/aranya-runtime/src/testing/dsl.rs
```diff
@@ -307,6 +317,33 @@ impl Display for TestRule {
...
+                    f,
+                    r#"{{"VerifyGraphIds": {{ "client": {}, "expected_fills": ["#,
+                    client
+                )?;
+                for (i, fill) in expected_fills.iter().enumerate() {
+                    if i > 0 {
+                        write!(f, ", ")?;
+                    }
+                    write!(f, "{fill}")?;
+                }
```
jdygert-spok, line 344: Nit: can just use the debug repr for the vec

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -2,19 +2,56 @@ use alloc::sync::Arc;
...
+                Err(errno) => return Some(Err(errno.into())),
+            };
+
+            let name = entry.name().to_bytes();
+            if name != b"." && name != b".." {
+                match GraphId::decode(name) {
+                    Ok(graph_id) => return Some(Ok(graph_id)),
+                    Err(_) => {
+                        warn!("Unable to decode GraphId: {:?}", entry.name())
+                    }
```
jdygert-spok, line 48: nit: 
```suggestion
                    Err(err) => {
                        warn!("Filename {:?} is not a valid GraphId: {}", entry.name(), err);
                    }
```

crates/aranya-runtime/src/testing/dsl.rs
```diff
@@ -679,6 +738,31 @@ where
...
+                    .ok_or(TestError::MissingClient)?
+                    .borrow_mut();
+
+                let mut actual_ids = state
+                    .provider()
+                    .list_graph_ids()
+                    .unwrap()
+                    .map(Result::unwrap)
+                    .collect::<Vec<_>>();
+                actual_ids.sort();
```
jdygert-spok, line 756: Could use btreeset instead of sorting :shrug:

crates/aranya-runtime/src/testing/testdata/generate_graph.test
jdygert-spok: #212

crates/aranya-runtime/src/testing/dsl.rs
```diff
@@ -679,6 +691,26 @@ where
...
+                let actual_ids: BTreeSet<GraphId> = state
+                    .provider()
+                    .list_graph_ids()
+                    .unwrap()
+                    .map(Result::unwrap)
+                    .collect();
+
+                let expected_ids: BTreeSet<GraphId> = ids
+                    .iter()
+                    .map(|id| graphs.get(id).unwrap().clone())
```
jdygert-spok, line 709: Clippy lint for clone and might as well use `[]` since unwrapping.
```suggestion
                    .map(|id| graphs[id])
```

jdygert-spok (dismissed):

crates/aranya-runtime/src/testing/dsl.rs
nikki-aranya: Huh, I tried that before and it complained but now it's fine

jdygert-spok (dismissed):

jdygert-spok (approved):

jdygert-spok (changes requested):
Nikki discovered the dir fd gets closed early/twice. Requesting change to prevent accidental merge.

crates/aranya-libc/src/sys/unix.rs
```diff
@@ -102,9 +102,12 @@ pub fn dup(old_fd: BorrowedFd<'_>) -> Result<RawFd, Errno> {
 }
 
 /// See `fdopendir(2)`.
-pub fn fdopendir(fd: BorrowedFd<'_>) -> Result<RawDir, Errno> {
+pub fn fdopendir(fd: OwnedFd) -> Result<RawDir, Errno> {
     // SAFETY: FFI call, no invariants.
     let dir = unsafe { libc::fdopendir(fd.fd) };
+    // SAFETY: This is now tied up in the DIR and will be freed once the
+    // OwnedDir gets dropped and we call closedir.
+    core::mem::forget(fd);
```
jdygert-spok, line 110: This should be only in the Ok case

jdygert-spok (approved):

gknopf-aranya (approved):

gknopf-aranya:
lgtm! Thanks for testing the crates out here https://github.com/aranya-project/aranya/pull/171 so we know this will work after the crates are released.

PR #205, opened 2025-04-10
(capi-codegen) Implement ByMutPtr trait for Arrays

ssweetney-spok:
As well as ByInput and ByConstPtr.

To support https://github.com/aranya-project/aranya/pull/183/files#r2036026419

### DIFF
--- a/crates/aranya-capi-codegen-test/src/defs.rs
+++ b/crates/aranya-capi-codegen-test/src/defs.rs
@@ -194,6 +194,12 @@ pub fn test_ptr_ptr_ptr_ptr_u32_ptr_ptr_ptr_ptr_u32(
 
 pub fn test_slice_u8_unit(_a: &[u8]) {}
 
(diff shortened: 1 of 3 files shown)


elagergren-spideroak (approved):

jdygert-spok:
Testing this out since arrays are funky in C. I noticed `[T; N]` doesn't implement `ByValue` or `Output` yet it can be returned, which produces an invalid header since C functions cannot directly return fixed size arrays. :thinking:

ssweetney-spok:
> Testing this out since arrays are funky in C. I noticed `[T; N]` doesn't implement `ByValue` or `Output` yet it can be returned, which produces an invalid header since C functions cannot directly return fixed size arrays. 🤔
> 

Ah, I see that the return value is only written to an output parameter under certain conditions.

> E: Seems like `Output` isn't used anywhere yet. I guess there isn't any restriction on return types?

Good catch. Seems like we could fix that

jdygert-spok:
#209

PR #211, opened 2025-04-10
chore: fix vet config and remove unused path dep versions

jdygert-spok:
Fixes `cargo vet` check: https://github.com/aranya-project/aranya-core/actions/runs/14389133920/job/40351787630

Removes `version` from `path` dependencies in crates that have not been published and from `dev-dependencies`.

### DIFF
--- a/crates/aranya-aqc-util/Cargo.toml
+++ b/crates/aranya-aqc-util/Cargo.toml
@@ -49,7 +49,7 @@ tracing = { workspace = true }
 [dev-dependencies]
 aranya-aqc-util = { path = ".", features = ["testing"] }
 
-aranya-crypto = { version = "0.5.0", path = "../aranya-crypto", default-features = false, features = ["getrandom"] }
+aranya-crypto = { path = "../aranya-crypto", default-features = false, features = ["getrandom"] }
 
 [package.metadata.docs.rs]
 all-features = true
--- a/crates/aranya-capi-codegen-test/Cargo.toml
+++ b/crates/aranya-capi-codegen-test/Cargo.toml
@@ -15,14 +15,14 @@ default = []
 test_cfg = []
 
 [dependencies]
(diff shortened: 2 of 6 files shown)


gknopf-aranya (approved):

jdygert-spok:
Oops, meant to do those other ones on another branch. Think they are fine here anyway?

gknopf-aranya:
> Oops, meant to do those other ones on another branch. Think they are fine here anyway?

I thought we are going to wait to pin the crate versions until right before the MVP release.
Otherwise we end up going back and forth between `main` and crate releases a lot as new crates are changed and published during development.

jdygert-spok:
> > Oops, meant to do those other ones on another branch. Think they are fine here anyway?
> 
> I thought we are going to wait to pin the crate versions until right before the MVP release. Otherwise we end up going back and forth between `main` and crate releases a lot as new crates are changed and published during development.

I'm not following. The second commit just removes unused version fields from crates that we are not publishing. e.g. aranya-model is not published so it always uses the local path dep of aranya-runtime, so the version field is pretty much meaningless.

gknopf-aranya (approved):

nikki-aranya:
I wanted to ask since I was thinking about this earlier, is there anything preventing us from moving these to workspace dependencies? It would be nice to have one source of truth so stuff doesn't get majorly out of step. The only thing I could see is release-plz or cargo vet complaining

djohnson-spideroak (approved):

jdygert-spok:
> I wanted to ask since I was thinking about this earlier, is there anything preventing us from moving these to workspace dependencies? It would be nice to have one source of truth so stuff doesn't get majorly out of step. The only thing I could see is release-plz or cargo vet complaining

There's a few reasons I don't like it that I've brought up before. A big one is that git no longer sees any change to a crate when you bump it's dependencies.

PR #213, opened 2025-04-11
Fix #168: Make sure function return types are actually defined.

### LINKED ISSUE #168: policy-compiler: check return type exists
```
struct Foo {}
function get_foo() struct Nonexistent {
    return Foo {}
}
```

This gives the error ``Return value of `get_foo()` must be struct Nonexistent``. We should first check that the type exists and return an "unknown type" error. This might apply somewhere else too.

apetkov-so:
Functions now ensure that all types (including argument types and return type) in the signature are actually defined. So, the following now produces an `NotDefined("struct Nonexistent")` error.
```
struct Foo {}
function get_foo() struct Nonexistent {
    return Foo {}
}

function get_bar(e enum Undefined) int {}
```


crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1220,6 +1221,27 @@ impl<'a> CompileState<'a> {
                     self.append_instruction(Instruction::Publish);
                 }
                 (ast::Statement::Return(s), StatementContext::PureFunction(fd)) => {
+                    // make sure return type actually exists
+                    match &fd.return_type {
+                        VType::Struct(name) => {
+                            if !["Envelope"].contains(&name.as_str())
```
jdygert-spok, line 1227: Hmm, on the one hand, not having this makes all the compiler tests much more annoying. On the other hand, you _do_ generally need something to supply a definition for envelope in order to actually use the vm.

I wonder if we want a flag or something, that makes seal/open blocks not required. This would also solve the issue of `seal { return None }` not working once we add a none type for typechecking.

jdygert-spok: Maybe make seal/open required if and only if `struct Envelope` is defined? Is that too magical?

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1220,6 +1221,27 @@ impl<'a> CompileState<'a> {
                     self.append_instruction(Instruction::Publish);
                 }
                 (ast::Statement::Return(s), StatementContext::PureFunction(fd)) => {
+                    // make sure return type actually exists
```
jdygert-spok, line 1224: Consider pulling this out to an `is_defined` method.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1220,6 +1221,27 @@ impl<'a> CompileState<'a> {
...
+                            }
+                        }
+                        VType::Enum(name) => {
+                            if self.m.enum_defs.get(name).is_none() {
+                                return Err(
+                                    self.err(CompileErrorType::NotDefined(format!("enum {name}")))
+                                );
+                            }
+                        }
+                        _ => {}
```
ssweetney-spok, line 1241: Might want to check `VType::Optional` too

crates/aranya-policy-compiler/src/compile.rs
ssweetney-spok, line 1227:
```suggestion
                            if "Envelope" != &name
```

apetkov-so: I expected more special structs, but since it's only this one, yeah, it makes sense.

crates/aranya-policy-compiler/src/compile.rs
apetkov-so: Done

crates/aranya-policy-compiler/src/compile.rs
apetkov-so: Added case for optionals, and a test case

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1485,6 +1489,25 @@ impl<'a> CompileState<'a> {
         self.m.progmem[r].iter().any(pred)
     }
 
+    /// Checks if the given type is defined. E.g. check struct/enum definitions.
+    fn ensure_type_is_defined(&self, vtype: &VType) -> Result<(), CompileError> {
```
ssweetney-spok, line 1493: I thought the doc comment wasn't necessary but it does show up if you use the `--document-private-items` flag with `cargo doc`

ssweetney-spok (approved):
LGTM 👍

crates/aranya-policy-compiler/src/compile.rs
apetkov-so: https://github.com/aranya-project/aranya-core/issues/216

jdygert-spok (approved):

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1220,6 +1221,9 @@ impl<'a> CompileState<'a> {
                     self.append_instruction(Instruction::Publish);
                 }
                 (ast::Statement::Return(s), StatementContext::PureFunction(fd)) => {
+                    self.ensure_type_is_defined(&fd.return_type)?;
```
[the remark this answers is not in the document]
ssweetney-spok: That makes sense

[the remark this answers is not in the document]
apetkov-so: We may as well move the type-matches-signature there too. But I don't see the gain... We'd still have to check every return statement in the function body. And that would be harder, because we don't have access to the return _statements_, only the instructions, which don't tell us about the return value.

jdygert-spok:
I just noticed we have the same issue with nonexistent function argument types.

crates/aranya-policy-compiler/src/compile.rs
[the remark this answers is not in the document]
apetkov-so: Ah yes, sorry, I was thinking of the case `return Nonexistent`, which we already handle. Still, the question of how to get the return type given only the return instruction remains.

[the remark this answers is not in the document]
apetkov-so: Ah yes, sorry, I was thinking of the case `return Nonexistent`, which we already handle.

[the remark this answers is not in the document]
apetkov-so: Hmm, removing the type-is-defined check from the ::Return statement causes it to insist that the "Return value should be Nonexistent" again. I guess we have to move the return-type-matches-signature check to compile_function as well, but that requires figuring out the return expression type, without compiling it - the compilation needs to stay in the ::Return statement.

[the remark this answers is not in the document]
jdygert-spok: We just need to move it up a line before `compile_statements`.

jdygert-spok (approved):

apetkov-so:
> I just noticed we have the same issue with nonexistent function argument types.
> 
> E: And struct fields... I'm guessing it's true everywhere a type is used.

Pushed commit that adds a check for each arg type.

ssweetney-spok (approved):
LGTM. Can you update the PR description to mention that undefined function args were handled too? Or include it in the description for the merge commit.

apetkov-so:
> LGTM. Can you update the PR description to mention that undefined function args were handled too? Or include it in the description for the merge commit.

Done.

jdygert-spok (approved):

PR #214, opened 2025-04-14
(capi-codegen) Restrict return types

### LINKED ISSUE #209: capi-codegen: restrict C return types
Add `deny(improper_ctypes*)` (and ensure it works when `include!`ing), and/or use `Output` to restrict return types.

ssweetney-spok:
Closes #209

- Add the `deny(improper_ctypes*)`attribute  to `extern "C"` functions
- When a value is directly returned from an `extern "C"` function, check that the value implements `aranya-capi-core::types::Output`


crates/aranya-capi-core/src/types/allowed.rs
```diff
@@ -101,6 +101,9 @@ impl_trait! {
 // SAFETY: `T: ByValue`, so `T` is FFI safe.
 unsafe impl<T: ByValue> Output for T {}
 
+// Wrapper type
+unsafe impl<T: Output, E> Output for Result<T, E> {}
+
```
ssweetney-spok, line 106: This shouldn't be necessary

crates/aranya-capi-codegen/src/ast/expand.rs
```diff
@@ -585,6 +587,14 @@ impl Ast {
                 })
                 .unwrap_or_else(|| quote!(#pattern.into()));
 
+            let result = if f_is_infallible {
+                let util = &ctx.util;
+
+                quote! (#util::check_valid_output_ty(#result))
+            } else {
+                result.clone()
+            };
```
ssweetney-spok, line 596: refactor so there's no need to clone

jdygert-spok:
`Output` will need some more thought, there's a difference between out params and actual returning, e.g. you can out-param an array but not return it. Same for `Safe` wrapper since it tracks its address. Maybe we should use `Output` for return values, and `ByMut` or a new trait for out params?

ssweetney-spok:
> `Output` will need some more thought, there's a difference between out params and actual returning, e.g. you can out-param an array but not return it. Same for `Safe` wrapper since it tracks its address. Maybe we should use `Output` for return values, and `ByMut` or a new trait for out params?

Isn't that handled already? `check_valid_output_ty` should only be called when the function returns a value and doesn't use an "out param". The `ByMut` check for the out param is there since it's a normal function argument

ssweetney-spok:
Or are you saying that if a type doesn't implement Output then we should rewrite the function to use an out param instead?

jdygert-spok:
> Isn't that handled already? `check_valid_output_ty` should only be called when the function returns a value and doesn't use an "out param". The `ByMut` check for the out param is there since it's a normal function argument

Oh I see it does, I misread why there was two instances of `check_valid_output_ty` added.

> Or are you saying that if a type doesn't implement Output then we should rewrite the function to use an out param instead?

That'd be nice, although I don't think we can do it generically so we'd have to special case arrays and the safe wrapper. Not a priority IMO.

elagergren-spideroak (approved):

PR #217, opened 2025-04-16
runtime: dedup dsl test list

jdygert-spok:


crates/aranya-runtime/src/testing/dsl.rs
```diff
@@ -880,6 +880,22 @@ macro_rules! test_vectors {
...
+        macro_rules! test_suite {
+            ($backend:expr) => {
+                $(
+                    #[test]
+                    fn $name() -> ::core::result::Result<(), $crate::testing::dsl::TestError> {
+                        $crate::testing::dsl::vectors::$name($backend)
+                    }
+                )*
+            };
+        }
```
jdygert-spok, line 897: TBH I'm surprised you can define a macro within another macro using an expansion like this.

elagergren-spideroak: Wait until you see the nasty stuff in aranya-crypto lol

elagergren-spideroak (approved):

nikki-aranya (approved):
Smart, makes me wonder why we didn't do this before now.

gknopf-aranya (approved):

PR #219, opened 2025-04-17
vm-policy: use `CS::tuple_hash` for merge ID

jdygert-spok:


crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -655,11 +656,18 @@ impl<E: aranya_crypto::Engine> Policy for VmPolicy<E> {
...
         let (left, right) = ids.into();
         let c = VmProtocolData::Merge { left, right };
+        let id = {
+            let digest = tuple_hash::<<E::CS as CipherSuite>::Hash, _>([
+                b"merge-command-id",
+                left.id.as_bytes(),
+                right.id.as_bytes(),
+            ]);
+            aranya_crypto::Id::from(digest.into_array()).into()
+        };
```
elagergren-spideroak, line 666: Command IDs are created aranya-crypto, maybe this should also live there.

elagergren-spideroak (approved):

crates/aranya-runtime/src/vm_policy.rs
elagergren-spideroak: I also think we should have more contextual binding, if possible. Eg, the not-yet-implemented PolicyId (I forget the issue # for it).

jdygert-spok: > Command IDs are created aranya-crypto, maybe this should also live there.

`CommandId` is defined in `crates/aranya-runtime/src/command.rs`, but yeah this could probably live somewhere better than here anyway :shrug:

> I also think we should have more contextual binding, if possible. Eg, the not-yet-implemented PolicyId (I forget the issue # for it).

Is it sufficient to have such info included "transitively", from ancestors' IDs which were generated including that info?

elagergren-spideroak: > > Command IDs are created aranya-crypto, maybe this should also live there.
> 
> `CommandId` is defined in `crates/aranya-runtime/src/command.rs`, but yeah this could probably live somewhere better than here anyway 🤷

See https://docs.rs/aranya-crypto/0.5.0/aranya_crypto/struct.SigningKey.html#method.sign_cmd
 
> > I also think we should have more contextual binding, if possible. Eg, the not-yet-implemented PolicyId (I forget the issue # for it).
> 
> Is it sufficient to have such info included "transitively", from ancestors' IDs which were generated including that info?

Yes and no. Yes in that what you're saying is correct: the right and left command IDs do carry with them their contextual binding, meaning you cannot create the same ID without including that same contextual binding.

But it's still good practice to include contextual binding in each step. For example, the "transitive" contextual binding doesn't help the case where a bug uses the wrong `VmPolicy` to create the merge command ID.

jdygert-spok: Oh I see, seems like it'd fit in aranya-crypto then.

Huh, guess I forgot about `CmdId`. I wonder if runtime should just re-export that :thinking:

Up to you what we should include. I guess even if we don't add the policy ID here we still probably need to finalize that and how policy is stored in init command before mvp, for ID stability...

elagergren-spideroak (approved):

gknopf-aranya (approved):

PR #220, opened 2025-04-18
Switch to 32 byte IDs

jdygert-spok:
Note from @elagergren-spideroak: We chose 512 bit IDs long before NIST’s PQ guidance (flow2, perhaps even flow1). 512 bit IDs don’t provide a meaningful improvement in collision resistance over 256 bit IDs, but they do have a negative performance impact. So, make the switch before MVP.


jdygert-spok:
Need to update a few tightly coupled tests.

jdygert-spok:
I "fixed" the dsl tests by making them match the current behavior :/

elagergren-spideroak (approved):

benz-spideroak (approved):

elagergren-spideroak (approved):

PR #228, opened 2025-04-24
(aranya-runtime) Storage addition and removal

ssweetney-spok:
- [x] Return the "init" command from `aranya_runtime::client::ClientState::new_graph`
- [x] Adds `aranya_runtime::client::ClientState::add_graph`
- [ ] TODO changes needed to implement `removeTeam`


crates/aranya-runtime/src/client.rs
```diff
@@ -88,9 +88,14 @@ where
...
         sink.commit();
 
-        let (graph_id, _) = self.provider.new_storage(perspective)?;
+        let (graph_id, storage) = self.provider.new_storage(perspective)?;
 
-        Ok(graph_id)
+        let segment = storage
+            .get_head()
+            .and_then(|location| storage.get_segment(location))?;
+        let init_command = segment.first().bytes().to_vec();
```
ssweetney-spok, line 96: The command at the head of the graph at this point should be the init command right?

jdygert-spok: No not the head, but the code is right. It's the first command in the segment containing the head of the graph. If the init action produces multiple commands, then the head will be the last of those commands.

Technically the storage impl _could_ split up the commands from the action into multiple segments though...

jdygert-spok: FYI `Command::bytes` isn't generally enough information to reconstruct the command, you also need the other "fields" of `Command`.

ssweetney-spok: > Technically the storage impl _could_ split up the commands from the action into multiple segments though...

To cover that case, in this scenario can I assume the first segment starts at index 0? So the init command always has the location `Location { segment: 0, command: 0}`?

> FYI Command::bytes isn't generally enough information to reconstruct the command, you also need the other "fields" of Command

By other fields, do you mean "id", "parent", "priority", "policy", and "max_cut"?

jdygert-spok: > So the init command always has the location Location { segment: 0, command: 0}?

Unfortunately not, the linear storage uses a file offset for the segment number.

I think the safest way to do this with the current interface is to walk backwards from the head with `Segment::prior`/`Segment::skip_list`. But we could add `Storage::get_init_command` or something.

> By other fields, do you mean "id", "parent", "priority", "policy", and "max_cut"?

Yep. See `SessionCommand` which is used for serializing. It's a good example but assumes a few things that would be different for an init command.

crates/aranya-runtime/src/client.rs
```diff
@@ -8,6 +8,7 @@ use crate::{
     Policy, Prior, Priority, Segment, Sink, Storage, StorageError, StorageProvider,
 };
 
+pub(crate) mod init;
```
ssweetney-spok, line 11: I only wanted this module to be visible for tests but I can't seem to find a way to do that 🤷

ssweetney-spok:
> So I understand the use case, this is to boostrap new devices, right? Transfer the init command and the graph ID and they'll be part of that team?

Yea, that's the idea. Although, how "transfer" is done is still [unclear](https://github.com/aranya-project/aranya/issues/167#issuecomment-2830525772)

crates/aranya-runtime/src/client/init.rs
```diff
@@ -0,0 +1,75 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::{Address, Command, CommandId, GraphId, Prior, Priority};
+
+#[derive(Serialize, Deserialize)]
+/// Used for serializing init commands
+pub(crate) struct InitCommand<'a> {
+    storage_id: GraphId,
+    #[serde(default = "Priority::init", skip)]
+    priority: Priority,
```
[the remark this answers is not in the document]
ssweetney-spok: Yea, you're right. I thought it might be useful to have them but now that I'm looking at it I don't see the value in keeping them

crates/aranya-runtime/src/client.rs
ssweetney-spok: Nice, I completely forgot about `not`. There's some weirdness with unresolved import errors appearing but I think this should work

crates/aranya-runtime/src/client.rs
```diff
@@ -77,7 +78,7 @@ where
         policy_data: &[u8],
         action: <E::Policy as Policy>::Action<'_>,
         sink: &mut impl Sink<E::Effect>,
-    ) -> Result<GraphId, ClientError> {
+    ) -> Result<(GraphId, Vec<u8>), ClientError> {
```
gknopf-aranya, line 81: It would be better to define a type alias for the init command to prevent accidentally passing it as a parameter in the wrong place and to prevent someone from passing a random vector of bytes into `add_team`.
```suggestion
    ) -> Result<(GraphId, InitCmd), ClientError> {
```

gknopf-aranya:
Once the init command is returned, we need a way to use it to initialize a new graph from scratch.
We should verify that if we have the team_id and init_command, a new graph can be created from that.

crates/aranya-runtime/src/client.rs
ssweetney-spok: Do you mean a newtype?

gknopf-aranya:
@ssweetney-spok, @chip-so, @jdygert-spok it looks like we still need unit-test coverage for `add_graph`

crates/aranya-runtime/src/client.rs
gknopf-aranya: Yeah, like this:
`pub type InitCmd = Vec<u8>`

ssweetney-spok: Type aliases don't provide the type safety that you described above. [See](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=8472ff5b2ff8f33ab8622d1fcbf8132a)

I think you are referring to the [newtype](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) pattern

gknopf-aranya: That works

gknopf-aranya (approved):

crates/aranya-runtime/src/client.rs
ssweetney-spok: > should we derive Serialize and Deserialize for the type? The command will likely need to be serialized before sending it to another peer.

Probably

As I'm working through it feels like adding a newtype around `Vec<u8>` just adds extra steps that don't seem necessary. To validate that the bytes are actually an init command we'd have to de-serialize them to an `InitCommand` and then serialize them back to bytes to create the `InitCmd(Vec<u8>)` newtype 🤔 

The first line in `ClientState::add_graph` tries to deserialize the bytes into a `InitCommand` so passing in random bytes will fail there.

gknopf-aranya: That's a runtime check vs. a compile time type check though.

gknopf-aranya: If a type is not defined for the `Vec<u8>` in `aranya-core`, we'll end up defining a type for this in `aranya`.
Seems better to define the type where it originated from.

ssweetney-spok: > That's a runtime check vs. a compile time type check though.

It is but we don't get any useful benefits from compile time type checking in this case. We don't expect the init command from `ClientState::new_graph` to be passed as an arg to `ClientState::add_graph` on the same `ClientState` instance or even in the same program. The init command will need to be serialized to be passed around to different devices. It's only when the bytes are de-serialized that we can know for sure what they are.

We'd be recreating the problem again but with another layer of indirection. Let's say we do add a `InitCmd(Vec<u8>)` newtype, we'll need to serialize it to pass it around. How do we tell that some random bytes actually came from serializing `InitCmd(Vec<u8>)` without executing code at runtime?

gknopf-aranya: You would first deserialize the bytes into a `InitCmd`, if that fails then they are rejected.
One the bytes have been deserialized into the `InitCmd` type, you can verify that the correct type is passed into `add_graph`.

gknopf-aranya: The type may pass through several levels of the architecture before getting to `add_graph`.
E.g. C API, Aranya client, IPC, daemon, graph. Having a way to do some sort of type checking would be useful to ensure whatever bytes were deserialized when they were received from another peer were properly passed through to `add_graph`.

crates/aranya-runtime/src/client/init.rs
```diff
@@ -0,0 +1,64 @@
...
+pub(crate) struct InitCommand<'a> {
+    storage_id: GraphId,
+    id: CommandId,
+    #[serde(borrow)]
+    data: &'a [u8],
+    #[serde(borrow)]
+    policy: &'a [u8],
+}
+
+pub(crate) type InitCmd = Vec<u8>;
```
gknopf-aranya, line 20: If this type is going to be used by other crates in the `aranya` repo, it will need to be `pub`:
```suggestion
pub type InitCmd = Vec<u8>;
```

crates/aranya-runtime/src/client.rs
ssweetney-spok: Added a type alias for `Box<[u8]>` with the name `InitCmd`

gknopf-aranya (dismissed):
lgtm

gknopf-aranya:
@chip-so, @elagergren-spideroak could you please take another look at this PR?

crates/aranya-runtime/src/storage/mod.rs
```diff
@@ -265,6 +265,26 @@ pub trait Storage {
     }
 }
 
+// TODO(Steve): Replace this with a method on the [`Storage`] trait
```
djohnson-spideroak, line 268: lets make sure this is covered by another issue, or create one.

crates/aranya-runtime/src/storage/mod.rs
```diff
@@ -265,6 +265,26 @@ pub trait Storage {
...
+pub(crate) fn get_first_segment<S: Storage>(storage: &S) -> Result<S::Segment, StorageError> {
```
djohnson-spideroak, line 269: I think this would benefit from a doc. From the other code/context, I can see this is intended to be called during new_graph

djohnson-spideroak (commented):
LGTM pending doc

crates/aranya-runtime/src/storage/mod.rs
ssweetney-spok: Gotcha, just created [one](https://github.com/aranya-project/aranya-core/issues/236)

djohnson-spideroak (approved):

djohnson-spideroak (approved):

ssweetney-spok:
So this PR was originally going to be kept small but now I'm updating this to make the all changes I need in "aranya-core" that are necessary to implement both `addTeam` and `removeTeam` APIs

crates/aranya-model/src/tests/mod.rs
```diff
@@ -611,6 +614,9 @@ fn should_sync_basic_clients() {
     // B will receive all the new commands it doesn't yet know about from client
     // A. At this stage of the test, that's the init, create, and
     // increment commands.
+    test_model
```
gknopf-aranya, line 617: Comment above is out-of-date. The graph needs to be added first before syncing.

crates/aranya-model/src/tests/mod.rs
gknopf-aranya, line 1: If we don't have one already, we should add a test that shows sync no longer works if `add_graph` wasn't called first.

ssweetney-spok:
It seems like we want to keep the existing behavior where if a sync response contains an init command and there isn't an existing storage, then one will be created

PR #235, opened 2025-05-01
policy-ifgen: use compile target to handle struct composition

apetkov-so:


apetkov-so:
The `tictactoe.md' policy now fails when it tries to `publish Start`, because that command is not defined. How hard would it be to add a test to use the generated code and run the policy to make sure what we output actually runs?

apetkov-so:
Pushed some fixes, but still facing some issues:
1. Fixing the policy codes causes diff errors against the golden versions. Can't figure out how to rebuild them, so I had to hand-edit the tictactoe.rs files to get the test to pass.
2. Tried adding a "use_tictactoe" test to run the generated code, but that always passes, even when the golden test fails due to policy compilation errors.
3. Fixed some tts policy issues (`fact Device` and `struct Device` duplicate name, and `has_role` being referenced before declaration). But there is a remaining problem: we call an undefined `crypto_derive_key_id` function - can't find that defined anywhere.

jdygert-spok:
> 1. Fixing the policy codes causes diff errors against the golden versions. Can't figure out how to rebuild them, so I had to hand-edit the tictactoe.rs files to get the test to pass.

`UPDATE_GOLDENFILES=1 cargo test -p aranya-policy-ifgen --tests`, we should note that somewhere...
 
> 2. Tried adding a "use_tictactoe" test to run the generated code, but that always passes, even when the golden test fails due to policy compilation errors.

Yeah it's very separated, and different from how you'd usually call the ifgen in a build script. It instead is only called via the golden.rs test. And `cargo test` will only make sure that the `use_tictactoe.rs` example compiles, which will just use the existing `tictactoe.rs`. If you want to actually run the examples you need `cargo run --example use_tictactoe`. But note that won't really "use" the policy, just the generated interface.

> 3. Fixed some tts policy issues (`fact Device` and `struct Device` duplicate name, and `has_role` being referenced before declaration). But there is a remaining problem: we call an undefined `crypto_derive_key_id` function - can't find that defined anywhere.

This was lazily grabbed from some WIP policy a while ago, it could probably be replaced by the policy from `aranya`.

It'd be nice to have a good single source of policy to integration test each of our policy-related crates against.

jdygert-spok:
I think the ideal long term solution would be to have the compiler produce an intermediate output which is just the interface. This would also open up some things like allowing use before definition, instead of requiring bottom-up ordering of definitions.

apetkov-so:
> This appears to have commits from PRs that are already merged. Should the target branch be changed?

Huh, I must have messed up a rebase somewhere. The target (#186) is still open, so the target branch is correct.

apetkov-so:
I'm tempted to just delete the ttc.rs test. I have to remove so much from it to get it to compile, it doesn't seem worth keeping it. We still have the tictactoe test. Thoughts? @jdygert-spok, @chip-so

crates/aranya-policy-ifgen/examples/use_tictactoe.rs
```diff
@@ -16,6 +16,9 @@ impl Actor for PrintClient {
 fn main() {
     let mut client = PrintClient;
     client
-        .assign_afc_label(Id::default(), 42, String::from("foo"))
+        .StartGame(Players {
```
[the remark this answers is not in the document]
jdygert-spok: It's now an example for tictactoe instead of ttc. Github is showing it as a file rename.

crates/aranya-policy-ifgen/tests/data/tictactoe.rs
```diff
@@ -19,8 +19,8 @@ pub struct Players {
 /// Player policy enum.
 #[value]
 pub enum Player {
-    X,
     O,
+    X,
 }
```
jdygert-spok, line 24: Uh oh, this reordered the enum variants which will break some other things.

jdygert-spok: I think I'd prefer to keep the original order, although we could also make it specify the value like `X = 1, O = 0`.

crates/aranya-policy-module/src/module.rs
```diff
@@ -76,6 +76,8 @@ pub struct ModuleV0 {
     pub action_defs: BTreeMap<String, Vec<ast::FieldDefinition>>,
     /// Command definitions
     pub command_defs: BTreeMap<String, BTreeMap<String, ast::VType>>,
+    /// Effect identifiers. The effect definitions can be found in `struct_defs`.
+    pub effects: Vec<String>,
```
jdygert-spok, line 80: Might be more proper to split effects off from regular structs, although more annoying.

jdygert-spok: Maybe it shouldn't be in the module, and compiling should return the compile target? :thinking:

crates/aranya-policy-ifgen/tests/data/tictactoe.rs
apetkov-so: It seems `UPDATE_GOLDEN_FILES...` changes the order... Not sure why, the source (`tictactoe.md`) defines them in X,O order.

apetkov-so: You mean add another compile method that returns the target instead of the module?
And `CompileTarget::into_module` would have to convert `enum_defs` from an `IndexMap` into a `BTreeMap`?

crates/aranya-policy-compiler/src/compile/target.rs
```diff
@@ -51,6 +55,16 @@ impl CompileTarget {
...
     pub fn into_module(self) -> Module {
+        // Convert enum defs IndexMap into BTreeMap.
+        let mut enum_defs = BTreeMap::new();
+        for (name, values) in self.enum_defs {
+            let mut ev = BTreeMap::new();
+            for (k, v) in values {
+                ev.insert(k, v);
+            }
+            enum_defs.insert(name, ev);
+        }
```
jdygert-spok, line 66:
```suggestion
        let enum_defs = self
            .enum_defs
            .into_iter()
            .map(|(k, v)| (k, v.into_iter().collect()))
            .collect();
```

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2357,6 +2359,27 @@ impl<'a> Compiler<'a> {
 
         Ok(cs.into_module())
     }
+
+    pub fn compile_target(self) -> Result<CompileTarget, CompileError> {
```
jdygert-spok, line 2363: `compile` can use `compile_target` now and be very short.

crates/aranya-policy-compiler/src/compile/target.rs
```diff
@@ -18,12 +19,14 @@ pub struct CompileTarget {
...
     pub command_defs: BTreeMap<Identifier, BTreeMap<Identifier, ast::VType>>,
+    /// Effect identifiers. The effect definitions can be found in `struct_defs`.
+    pub effects: Vec<Identifier>,
     /// Fact schemas
     pub fact_defs: BTreeMap<Identifier, FactDefinition>,
     /// Struct schemas
     pub struct_defs: BTreeMap<Identifier, Vec<ast::FieldDefinition>>,
     /// Enum definitions
-    pub enum_defs: BTreeMap<Identifier, BTreeMap<Identifier, i64>>,
+    pub enum_defs: IndexMap<Identifier, IndexMap<Identifier, i64>>,
```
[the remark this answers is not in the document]
apetkov-so: It's less hassle to stay with `BTreeMap`... Otherwise we'd have to decorate all `IndexMap` fields with ` #[serde(with = "serde_seq")]` to satisfy `Module`'s `Serialize/Deserialize` traits. And deal with custom hashers for no-std crates.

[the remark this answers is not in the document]
jdygert-spok: I don't see why either of those are needed :thinking:

crates/aranya-policy-compiler/Cargo.toml
```diff
@@ -14,6 +14,8 @@ workspace = true
 [features]
 default = []
 
+std = []
```
jdygert-spok, line 17: revert, compiler always requires std currently

jdygert-spok:
LGTM after last two things

crates/aranya-policy-compiler/src/compile/target.rs
```diff
@@ -18,12 +19,14 @@ pub struct CompileTarget {
     pub action_defs: BTreeMap<Identifier, Vec<ast::FieldDefinition>>,
     /// Command definitions (`fields`)
     pub command_defs: BTreeMap<Identifier, BTreeMap<Identifier, ast::VType>>,
+    /// Effect identifiers. The effect definitions can be found in `struct_defs`.
+    pub effects: Vec<Identifier>,
```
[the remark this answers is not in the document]
apetkov-so: That changes the order of `Effects` in tictactoe.rs. I don't like seeing GameOver appear before GameStart, and GameUpdate, but it doesn't break anything.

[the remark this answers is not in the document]
jdygert-spok: Since everything else is changed to be in sorted order, I'd like to do the same for effects. I agree it is a bit awkward to see stuff out of the defined order, but I like that it keeps everything consistent.

crates/aranya-policy-compiler/src/compile/target.rs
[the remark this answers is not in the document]
apetkov-so: Maybe I configured the `indexmap` dependency wrong, but I couldn't compile without the `serde` tag on the field, even when enabling the `serde` feature.
And IndexMap's default hasher comes from std, so we'd need an alternative.

https://docs.rs/indexmap/latest/indexmap/#feature-flags
https://docs.rs/indexmap/latest/indexmap/#no-standard-library-targets
https://docs.rs/indexmap/latest/indexmap/#alternate-hashers

crates/aranya-policy-compiler/src/compile/target.rs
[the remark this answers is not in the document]
apetkov-so: Ok

crates/aranya-policy-compiler/src/compile/target.rs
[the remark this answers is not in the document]
apetkov-so: Just to clarify, the serde/std issues only arise in `policy-module`, not `policy-compiler`. `CompileTarget` doesn't use `Serialize/Deserialize`, and the compiler crate is always std, so it can switch to `IndexMap` without issue. But the policy-module is more complicated, though perhaps due to my cluelessness.

jdygert-spok (approved):

PR #238, opened 2025-05-12
crypto: pair PSKs with cipher suites

### LINKED ISSUE #237: aqc: map each PSK to a hash algorithm
Spec: https://github.com/aranya-project/aranya-docs/issues/97

elagergren-spideroak:
Fixes #237

### DIFF
--- a/crates/aranya-crypto/src/aqc/bidi.rs
+++ b/crates/aranya-crypto/src/aqc/bidi.rs
@@ -1,21 +1,26 @@
 use core::fmt;
 
 use serde::{Deserialize, Serialize};
+use zerocopy::{ByteEq, Immutable, IntoBytes, KnownLayout, Unaligned};
 
 use crate::{
-    aqc::shared::{RawPsk, RootChannelKey},
+    aqc::{
+        shared::{RawPsk, RootChannelKey},
+        suite::CipherSuiteId,
+    },
     aranya::{DeviceId, Encap, EncryptionKey, EncryptionPublicKey},
     ciphersuite::SuiteIds,
     csprng::Random,
     engine::unwrapped,
     error::Error,
     hash::{tuple_hash, Digest, Hash},
-    hpke::{Hpke, Mode},
+    hpke::{Hpke, Mode, RecvCtx, SendCtx},
     id::{custom_id, Id},
     import::ImportError,
     kem::Kem,
     misc::sk_misc,
     subtle::{Choice, ConstantTimeEq},
+    util::Either,
     CipherSuite, Engine,
 };
 
@@ -34,6 +39,8 @@ use crate::{
 ///         BidiPeerEncap,
 ///         BidiPsk,
 ///         BidiSecrets,
+///         BidiSecret,
+///         CipherSuiteId,
 ///     },
 ///     CipherSuite,
 ///     Csprng,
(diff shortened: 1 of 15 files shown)


crates/aranya-crypto/src/aqc/suite.rs
```diff
@@ -0,0 +1,117 @@
...
+
+/// A TLS 1.3 cipher suite.
+#[repr(u16)]
+#[derive(
+    Copy, Clone, Debug, Eq, PartialEq, Immutable, IntoBytes, KnownLayout, Serialize, Deserialize,
+)]
+#[non_exhaustive]
+pub enum CipherSuiteId {
+    /// TLS_AES_128_GCM_SHA256
+    TlsAes128GcmSha256 = 0x1301u16.to_be(),
```
jdygert-spok, line 14: nit: use separated `0x1301_u16` for clarity, or `u16::to_be(0x1301)`

jdygert-spok (approved):

crates/aranya-crypto/src/aqc/suite.rs
elagergren-spideroak: They're BE so that `zerocopy` when transmutes the enum to `[u8; 2]` the bytes will be in the right order.

crates/aranya-crypto/src/aqc/bidi.rs
```diff
@@ -269,15 +281,17 @@ impl<CS: CipherSuite> BidiSecrets<CS> {
...
 
-/// A PSK for a bidirectional channel.
-pub struct BidiPsk<CS> {
+/// The shared bidirectional channel secret used by both the
+/// channel author and channel peer to derive individual PSKs.
+pub struct BidiSecret<CS: CipherSuite> {
     id: BidiChannelId,
-    psk: RawPsk<CS>,
+    #[allow(clippy::type_complexity)]
+    ctx: Either<SendCtx<CS::Kem, CS::Kdf, CS::Aead>, RecvCtx<CS::Kem, CS::Kdf, CS::Aead>>,
```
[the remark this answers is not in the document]
elagergren-spideroak: It usually is better to use a specific enum. I don't usually mind `Either` so long as it isn't part of the public API.

crates/aranya-crypto/src/aqc/bidi.rs
```diff
@@ -303,16 +317,15 @@ impl<CS: CipherSuite> BidiPsk<CS> {
...
 
-        let id = BidiPeerEncap::<CS>(Encap(enc)).id();
-
-        // See section 9.8 of RFC 9180.
-        let psk = ctx.export(b"aqc bidi psk")?;
-
-        Ok(Self { id, psk })
+        Ok(Self {
+            id: BidiPeerEncap::<CS>(Encap(enc)).id(),
+            ctx: Either::Left(ctx),
```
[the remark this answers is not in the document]
elagergren-spideroak: The compiler remembers, though :)

crates/aranya-crypto/src/aqc/suite.rs
elagergren-spideroak: It's not an optimization, I'm just using zerocopy so that I don't have to write the encoding by hand.

jdygert-spok (approved):

mooreoak (approved):
Looks solid

crates/aranya-crypto/src/aqc/suite.rs
elagergren-spideroak: It would be. But CipherSuiteId is a field of PskCtx, and PskCtx uses zerocopy to serialize itself. That means CipherSuiteId must be in the correct byte order in memory so that zerocopy serializes it correctly.

PR #239, opened 2025-05-13
capi: make opaque types an exact size and alignment

### LINKED ISSUE #227: capi: opaque size mismatch causes slice issues
We assumed a larger size for an opaque type would be fine since the real type would fit inside. This doesn't work when using the opaque type in a slice, since the stride does not match between C and Rust.

If we make the `opaque` macro require exact size/alignment, then we will need many `cfg_attr`s for different platforms/configurations.

Suggested fix: pad the Rust type to fit the declared size/align.

jdygert-spok:
The opaque macro now serves two functions. It first wraps the original type in the `Opaque` wrapper. Then it applies the `[u8; SIZE]` representation to the generated renamed type that cbindgen sees.

Closes #227.

### DIFF
--- a/crates/aranya-capi-core/src/opaque.rs
+++ b/crates/aranya-capi-core/src/opaque.rs
@@ -1,0 +1,70 @@
+use std::{
+    mem::ManuallyDrop,
+    ops::{Deref, DerefMut},
+};
+
+#[repr(C)]
+pub struct Opaque<const SIZE: usize, const ALIGN: usize, T>(OpaqueRepr<SIZE, ALIGN, T>)
+where
+    elain::Align<ALIGN>: elain::Alignment;
+
+impl<const SIZE: usize, const ALIGN: usize, T> Deref for Opaque<SIZE, ALIGN, T>
+where
+    elain::Align<ALIGN>: elain::Alignment,
+{
+    type Target = T;
+    fn deref(&self) -> &Self::Target {
+        const { assert!(SIZE >= size_of::<T>()) };
+        const { assert!(ALIGN >= align_of::<T>()) };
+
+        // SAFETY: `OpaqueRepr` is always the `inner` variant.
+        unsafe { &self.0.inner }
+    }
+}
+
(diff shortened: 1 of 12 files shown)


crates/aranya-capi-macro/src/opaque.rs
```diff
@@ -1,70 +1,86 @@
...
-        Item::Struct(s) => (&s.vis, &s.ident, &s.attrs),
-        Item::Type(t) => (&t.vis, &t.ident, &t.attrs),
-        item => {
-            return Err(Error::new_spanned(
-                item,
-                "`#[capi::opaque]` can only be applied to `struct`s",
-            ))
-        }
+    let Item::Type(mut t) = item else {
+        // TODO(jdygert): Allow structs again?
```
ssweetney-spok, line 23: Should the doc comment in "crates/aranya-capi-codegen/src/syntax/opaque.rs" be updated?

crates/aranya-capi-core/src/opaque.rs
```diff
@@ -0,0 +1,157 @@
...
+
+    #[repr(C)]
+    pub(super) union OpaqueRepr<const SIZE: usize, const ALIGN: usize, T>
+    where
+        elain::Align<ALIGN>: elain::Alignment,
+    {
+        // Invariant: always the `inner` variant.
+        // Invariant: only dropped in the drop impl.
+        inner: ManuallyDrop<T>,
+        __backing_for_size_align_only: Backing<SIZE, ALIGN>,
```
ssweetney-spok, line 22: Can you explain how this works? Shouldn't the size of `T` be used to determine how much padding is needed to get the specified size and alignment?

ssweetney-spok: Just now realized this is a union not a struct 🤦. It makes sense now

ssweetney-spok (approved):
Looks great! I found this approach really interesting

Very minor nit: The usage of `core::convert::Infallible` confused me at first. It might make things easier to understand by defining an enum with no variants and adding some doc comments that explain that it only exists to prevent a `Backing` from being constructed

elagergren-spideroak (approved):
Much better. Thanks.

PR #247, opened 2025-05-21
afc: use workspace lints

jdygert-spok:


crates/aranya-fast-channels/src/lib.rs
```diff
@@ -248,13 +244,10 @@
...
     clippy::unimplemented,
-    clippy::unwrap_used,
-    missing_docs,
-    rust_2018_idioms,
-    unused_lifetimes,
-    unused_qualifications
+    missing_docs
 )]
 #![cfg_attr(not(any(feature = "std", test)), deny(clippy::std_instead_of_core))]
+#![allow(clippy::arithmetic_side_effects, reason = "TODO")]
```
elagergren-spideroak, line 250: TODO?

elagergren-spideroak (approved):
lgtm other than my one comment

djohnson-spideroak (commented):
LGTM pending CI and Erics comment

crates/aranya-fast-channels/src/lib.rs
jdygert-spok: Yeah there's quite a lot of cases to fix :/

I just fixed a few but the rest could use a closer look and/or benefit from switching to zerocopy... I would love to get sucked into that but it's low priority right now.

crates/aranya-fast-channels/src/shm/le.rs
```diff
@@ -38,13 +38,15 @@ macro_rules! little_endian {
 
             impl ::core::ops::AddAssign<$type> for $name {
                 fn add_assign(&mut self, rhs: $type) {
-                    self.0 = <$type>::from_le(self.0) + rhs
+                    #![allow(clippy::arithmetic_side_effects, reason = "keeping behavior")]
+                    *self = Self::new(Self::into(*self) + rhs);
```
jdygert-spok, line 42: These were actually wrong AFAICT. I think they would do some weird flip flopping on BE.

crates/aranya-fast-channels/src/lib.rs
jdygert-spok: #253

PR #257, opened 2025-05-28
crypto: impl `Hash` for `CipherSuiteId` and rename serde repr

jdygert-spok:


elagergren-spideroak (approved):

elagergren-spideroak:
is this for sending them via IPC?

jdygert-spok:
Yes-ish, the string repr is only for human readable formats, which  I believe the IPC we use isn't.

PR #263, opened 2025-06-05
crypto: update to spideroak-crypto v0.5.x

elagergren-spideroak:


crates/aranya-crypto/src/lib.rs
```diff
@@ -40,7 +40,7 @@
 #![allow(unstable_name_collisions)]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![cfg_attr(not(any(test, doctest, feature = "std")), no_std)]
-#![cfg_attr(not(all(test, feature = "trng")), forbid(unsafe_code))]
+#![cfg_attr(not(all(test, feature = "trng")), warn(unsafe_code))]
```
benz-spideroak, line 43: Why the change to allow unsafe_code?

crates/aranya-crypto/src/ciphersuite/ext.rs
```diff
@@ -0,0 +1,699 @@
...
+        T: Expand;
+}
+
+impl<CS: CipherSuite> CipherSuiteExt for CS {
+    fn tuple_hash<const N: usize>(tag: &[u8], context: [&[u8]; N]) -> Digest<Self> {
+        hash::tuple_hash::<Self::Hash, _>(TupleHashItems::<Self, N> {
+            tag: iter::once(tag),
+            oids: CS::OIDS.into_iter(),
+            items: context.iter(),
+        })
```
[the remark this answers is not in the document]
elagergren-spideroak: I looked at a few different options (manual impl, chain, copying into an array), and the manual impl generated better assembly and used less stack space. I did not benchmark the different options, though.

Normally it wouldn't matter and `chain` would be the right approach, but I'm a little worried about the performance impact of adding a bunch of new contextual binding. E.g.,  we use `tuple_hash(A, B, C)` in a number of places where we would ordinarily use `op(A, B, C)`, but can't because we don't have a way to do no-alloc `op(A, B, C)`. :/

crates/aranya-crypto/src/ciphersuite/mod.rs
```diff
@@ -50,66 +80,24 @@ use crate::{
 /// Additionally, please test your implementation using the
 /// `test_util` module.
 pub trait CipherSuite {
+    /// OIDS contains the OIDs from the algorithms in the cipher
+    /// suite.
+    const OIDS: Oids<Self> = Oids::new();
```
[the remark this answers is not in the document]
jdygert-spok: Oh, you can't override since `Oids::new` is not pub, and it is a ZST anyway. Doesn't matter too much then.

crates/aranya-crypto/src/oid.rs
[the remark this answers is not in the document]
elagergren-spideroak: Yeah. We require all our primitives to implement `oid::Identified`, but not every cryptographic construction has been assigned an OID (e.g., [this commit](https://github.com/aranya-project/crypto/commit/8c92941d0b196653bb451f5a4f7e926fc059f8ca)). I don't want users to have to manually implement newtype wrappers just to add an OID to a type.

```rust
struct Foo;
impl CipherSuite for Foo {
    type Kem = DhKemP256HkdfSha256;
    [...]
}
kem_with_oid! {
    struct DhKemP256HkdfSha256(rust::DhKemP256HkdfSha256) => DHKEM_P256_HKDF_SHA256
}
```

crates/aranya-crypto/src/ciphersuite/ext.rs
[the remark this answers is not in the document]
elagergren-spideroak: Oh, yeah, I actually know for a fact that approach will work better. Let's merge that, cut v0.5.3, then update this to use chain.

elagergren-spideroak:
> * Can remove byteorder from aranya-crypto very easily

low on my todo list, but it's on there!

Cargo.toml
```diff
@@ -76,6 +76,7 @@ panic = "unwind"
 [profile.release]
 lto = "thin"
 panic = "abort"
+strip = "debuginfo" # TODO
```
jdygert-spok, line 79: ?

crates/aranya-crypto/Cargo.toml
```diff
@@ -126,16 +143,20 @@ trng = ["spideroak-crypto/trng"]
 buggy = { version = "0.1.0", default-features = false }
 
 spideroak-base58 = { version = "0.1.0", default-features = false }
-spideroak-crypto = { version = "0.2.0", default-features = false }
+spideroak-crypto = { version = "0.5", default-features = false, features = ["serde"] }
```
jdygert-spok, line 146: Should specify 0.5.3 since we do need it (or at least need 0.5.2)

crates/aranya-crypto/Cargo.toml
```diff
@@ -126,16 +143,17 @@ trng = ["spideroak-crypto/trng"]
 buggy = { version = "0.1.0", default-features = false }
 
 spideroak-base58 = { version = "0.1.0", default-features = false }
-spideroak-crypto = { version = "0.2.0", default-features = false }
+spideroak-crypto = { version = ">=0.5.2", default-features = false, features = ["serde"] }
```
jdygert-spok, line 146:
```suggestion
spideroak-crypto = { version = "0.5.2", default-features = false, features = ["serde"] }
```

elagergren-spideroak: ?

elagergren-spideroak: oh, nvm

jdygert-spok: `>=` allows e.g. 0.6

elagergren-spideroak: yeah i always forget `0.5.2` is `0.5.2` >= v < `0.6.0` and not _only_ `v == 0.5.2`.

PR #265, opened 2025-06-09
shrink Repr size 

elagergren-spideroak:


jdygert-spok:
The tagging breaks on BE, try `cargo +nightly miri test --target powerpc64-unknown-linux-gnu`.

elagergren-spideroak:
> The tagging breaks on BE, try `cargo +nightly miri test --target powerpc64-unknown-linux-gnu`.

I was thinking that'd be the case.

elagergren-spideroak:
> > The tagging breaks on BE, try `cargo +nightly miri test --target powerpc64-unknown-linux-gnu`.
> 
> I was thinking that'd be the case.

We could just... not do inlining on BE systems?

jdygert-spok:
Close in favor of #272?

PR #268, opened 2025-06-10
storage: remove_graph

gknopf-aranya:
Remove graph from storage based on the graph ID.
For mem storage, remove the graph from the map to remove it from memory.
For linear storage, remove the graph file from the filesystem.

Test scenario:
- A creates a graph
- B syncs with A (to get the graph from A)
- Graphs A & B match
- B locally adds a command to graph
- Graphs A & B no longer match
- B deletes graph
- B syncs with A (to get graph again from A)
- Graphs A & B match again


crates/aranya-runtime/src/storage/linear/mod.rs
```diff
@@ -258,6 +258,13 @@ impl<FM: IoManager> StorageProvider for LinearStorageProvider<FM> {
         Ok(entry.insert(LinearStorage::open(file)?))
     }
 
+    fn remove_storage(&mut self, graph: GraphId) -> Result<(), StorageError> {
```
gknopf-aranya, line 261: Add better error handling to this method

crates/aranya-model/src/model.rs
```diff
@@ -366,6 +373,30 @@ where
         Ok(sink.effects)
     }
 
+    /// Remove a graph from a client
+    fn remove_graph(
+        &mut self,
+        proxy_id: Self::GraphId,
+        client_proxy_id: Self::ClientId,
+    ) -> Result<(), ModelError> {
+        let Some(storage_id) = self.storage_ids.remove(&proxy_id.into()) else {
```
jdygert-spok, line 382: If we're only removing it from one client then we should not remove it from the storage_id map

crates/aranya-model/src/model.rs
```diff
@@ -366,6 +373,30 @@ where
...
+            return Err(ModelError::GraphNotFound);
+        };
+
+        let mut state = self
+            .clients
+            .get_mut(&client_proxy_id.into())
+            .ok_or(ModelError::ClientNotFound)?
+            .state
+            .borrow_mut();
```
jdygert-spok, line 391: Nit: use infallible `get_mut()` when you have a `&mut RefCell`. (I see existing code was the same)

```suggestion
            .get_mut();
```

crates/aranya-runtime/src/storage/linear/mod.rs
```diff
@@ -258,6 +258,17 @@ impl<FM: IoManager> StorageProvider for LinearStorageProvider<FM> {
...
+        self.manager.remove(graph)?;
+
+        let _ = self
+            .storage
+            .remove(&graph)
+            .ok_or(StorageError::NoSuchStorage);
```
jdygert-spok, line 267: Why ignoring error?

crates/aranya-runtime/src/storage/memory.rs
```diff
@@ -110,6 +110,15 @@ impl StorageProvider for MemStorageProvider {
             .ok_or(StorageError::NoSuchStorage)
     }
 
+    fn remove_storage(&mut self, graph: GraphId) -> Result<(), StorageError> {
+        let _ = self
+            .storage
+            .remove(&graph)
+            .ok_or(StorageError::NoSuchStorage);
```
jdygert-spok, line 117: Why ignoring error?

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -127,6 +127,13 @@ impl IoManager for FileManager {
         Ok(Some(Writer::open(fd)?))
     }
 
+    fn remove(&mut self, id: GraphId) -> Result<(), StorageError> {
+        let path = id.to_path()?;
+        libc::remove(path)?;
+
+        Ok(())
+    }
+
```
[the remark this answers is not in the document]
jdygert-spok: We should have a test that checks the graph is empty when recreated after deletion. Also FYI the model does not use the file based linear storage.

crates/aranya-model/src/model.rs
gknopf-aranya: Not planning to implement a counter to track this on this PR.
Aiming to get the minimum functionality required for MVP on this PR.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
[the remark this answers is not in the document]
gknopf-aranya: Is there a way to test that without adding file-based linear storage to the model?
How is the rest of the file-based linear storage functionality currently tested?

crates/aranya-runtime/src/testing/dsl.rs
```diff
@@ -536,6 +545,18 @@ where
...
             }
+            TestRule::RemoveGraph { client, id } => {
+                let state = clients
+                    .get_mut(&client)
+                    .ok_or(TestError::MissingClient)?
+                    .get_mut();
+                let storage_id = graphs.get(&id).ok_or(TestError::MissingGraph(id))?;
+                state.remove_graph(*storage_id)?;
+                // TODO: is this needed?
+                //graphs.remove(&id);
```
jdygert-spok, line 556: No, at least for now. Same situation as the model's proxy id map.

jdygert-spok (dismissed):

crates/aranya-runtime/src/testing/testdata/remove_graph.test
ssweetney-spok, line 1: Is it worth adding a command to client0/clientA just before the final sync to verify that new commands can be added after a graph is removed? It looks like client0 has an empty graph during that last sync

gknopf-aranya: Yeah, I think it's worth it. Easy to add this additional testing and verifies a common scenario that could happen.

crates/aranya-runtime/src/testing/dsl.rs
gknopf-aranya: Will remove

ssweetney-spok (approved):

PR #278, opened 2025-06-12
Remove custom IDs' dependencies on `aranya-crypto`

jdygert-spok:


jdygert-spok:
```rust
pub trait Identified {
    type Id: ... + Into<Id>; // :(
}
```

jdygert-spok:
To remove `aranya_policy_module::Value` related dependency on aranya-crypto, I think the only options are:

- a separate crate for the ID type (Either separate crypto ID crate or vm-id crate which crypto depends on, which seems like a bad inversion)
- Only allowing bare `VmId`s in ffi-derive and policy-ifgen (or maybe we could work some autoderef specialization magic? IDK)

PR #296, opened 2025-06-17
crypto: encrypt PSK seeds for other devices

### LINKED ISSUE #295: crypto: allow encrypting PSK for device

elagergren-spideroak:
Fixes #295

### DIFF
--- a/crates/aranya-crypto/src/policy.rs
+++ b/crates/aranya-crypto/src/policy.rs
@@ -10,6 +10,12 @@ use crate::{
 };
 
 custom_id! {
+    /// Uniquely identifies a group.
+    #[derive(Immutable, IntoBytes, KnownLayout, Unaligned)]
+    pub struct GroupId;
+}
+
+custom_id! {
     /// Uniquely identifies a policy.
     #[derive(Immutable, IntoBytes, KnownLayout, Unaligned)]
     pub struct PolicyId;
--- a/crates/aranya-crypto/src/tls/psk.rs
+++ b/crates/aranya-crypto/src/tls/psk.rs
@@ -271,7 +271,7 @@ impl<CS: CipherSuite> ConstantTimeEq for PskSeed<CS> {
 }
 
 /// An encrypted [`PskSeed`].
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Serialize, Deserialize)]
 #[serde(bound = "CS: CipherSuite")]
 pub struct EncryptedPskSeed<CS: CipherSuite> {
     // NB: These are only `pub(crate)` for testing purposes.
@@ -286,6 +286,15 @@ impl<CS: CipherSuite> Clone for EncryptedPskSeed<CS> {
             ciphertext: self.ciphertext.clone(),
             tag: self.tag.clone(),
         }
+    }
+}
+
+impl<CS: CipherSuite> fmt::Debug for EncryptedPskSeed<CS> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("EncryptedPskSeed")
+            .field("ciphertext", &self.ciphertext)
+            .field("tag", &self.tag)
+            .finish()
     }
 }
 
--- a/crates/aranya-crypto/src/tls/psk.rs
+++ b/crates/aranya-crypto/src/tls/psk.rs
@@ -2,18 +2,25 @@ use core::{cell::OnceCell, fmt, marker::PhantomData};
 
 use buggy::{Bug, BugExt};
 use serde::{Deserialize, Serialize};
+use spideroak_crypto::{
+    aead::Tag,
+    hpke::Mode,
+    kdf::{self, Kdf},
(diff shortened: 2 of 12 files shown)


crates/aranya-crypto/src/policy.rs
```diff
@@ -9,6 +9,12 @@ use crate::{
     id::{custom_id, Id},
 };
 
+custom_id! {
+    /// Uniquely identifies a group.
```
gknopf-aranya, line 13: What is a group in this context?

crates/aranya-crypto/src/tls/psk.rs
```diff
@@ -173,6 +270,25 @@ impl<CS: CipherSuite> ConstantTimeEq for PskSeed<CS> {
     }
 }
 
+/// An encrypted [`PskSeed`].
+#[derive(Debug, Serialize, Deserialize)]
```
gknopf-aranya, line 274: Confirming this `#[derive(Serialize)]` will allow us to serialize the encrypted PSK before sending it to a peer?

crates/aranya-crypto/src/tls/psk.rs
```diff
@@ -120,6 +159,64 @@ impl<CS: CipherSuite> PskSeed<CS> {
...
+        let EncryptedPskSeed {
+            mut ciphertext,
+            tag,
+        } = ciphertext;
+
+        // info = H(
+        //     "PskSeed-v1",
+        //     suite_id,
+        //     group,
+        // )
```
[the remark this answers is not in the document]
gknopf-aranya: I thought this may be part of the documentation

crates/aranya-crypto/src/tls/psk.rs
```diff
@@ -120,6 +159,64 @@ impl<CS: CipherSuite> PskSeed<CS> {
...
+        Ok((Encap(enc), EncryptedPskSeed { ciphertext, tag }))
+    }
+
+    /// Uses `self` to decrypt and authenticate a [`PskSeed`]
+    /// that was encrypted by `peer_pk`.
+    pub fn open_psk_seed(
+        &self,
+        encap: &Encap<CS>,
+        ciphertext: EncryptedPskSeed<CS>,
+        peer_pk: &EncryptionPublicKey<CS>,
```
ssweetney-spok, line 198: So devices that want to join a team by calling `add_team` also need the public key of the owner?

crates/aranya-crypto/src/policy.rs
elagergren-spideroak: Currently a team, but it's deliberately underspecified right now since we have vague plans to have "groups" in the future, and it's TBD if a team is also a group or not.

crates/aranya-crypto/src/tls/psk.rs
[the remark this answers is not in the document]
elagergren-spideroak: It's not accidental, it's docs.

gknopf-aranya:
Does this PR introduce any breaking changes to aranya-crypto?

crates/aranya-crypto/src/tls/psk.rs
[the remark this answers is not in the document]
ssweetney-spok: Only to have less fields in the config.

```Rust
pub struct QuicSyncConfig {
    sealed_seed: Box<[u8]>, // encap and encrypted seed are contained here
}
```

vs
```
pub struct QuicSyncConfig {
    encrypted_seed: Box<[u8]>,
    encap_key: Box<[u8>],
}
```

crates/aranya-crypto/src/tls/psk.rs
elagergren-spideroak: We could actually probably have a flow like this. Note that "admin" doesn't literally mean the admin role.

- Device sends device keys to admin
- Admin publishes the AddDevice command
- Admin sends the team ID to the device
- Admin sends an ephemeral graph command to the device with other info (like the PSK)

ssweetney-spok: Can devices receive commands before they've called `add_team`?

gknopf-aranya: Devices wouldn't be able to sync before adding the PSK to the QUIC syncer.
But devices can receive ephemeral commands via the Rust & C API (we do this for AQC).
The problem is, we'd at least need the Init Command and the AddDevice commands already on the graph before encrypting an ephemeral PSK command.

elagergren-spideroak: > The problem is, we'd at least need the Init Command and the AddDevice commands already on the graph before encrypting an ephemeral PSK command.

I don't think that's a problem. My previous comment explicitly mentions this.

elagergren-spideroak:
> I patched the aranya-crypto crate with the changes on this branch. I'll post here if I run into any issues while integrating

once this is approved i'll send a PR to bump aranya-crypto's version

gknopf-aranya (approved):

PR #297, opened 2025-06-17
all: use derive-where for better derive bounds

jdygert-spok:
Removes some unneeded bounds like `where CS: Clone` which the regular derive introduces, and converts some manual impls to derives.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ byteorder = { version = "1", default-features = false }
 bytes = { version = "1.9.0" }
 cfg-if = { version = "1", default-features = false }
 const_format = { version = "0.2", default-features = false }
-derive-where = { version = "1" }
+derive-where = { version = "1", default-features = false }
 errno = { version = "0.3", default-features = false }
 heapless = { version = "0.8", default-features = false }
 libc = { version = "0.2", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ byteorder = { version = "1", default-features = false }
 bytes = { version = "1.9.0" }
 cfg-if = { version = "1", default-features = false }
 const_format = { version = "0.2", default-features = false }
+derive-where = { version = "1" }
 errno = { version = "0.3", default-features = false }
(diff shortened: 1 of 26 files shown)


jdygert-spok:
I noticed some usage of `OnceCell` for IDs gets cloned but some create a new one on clone. Similarly some get evaluated during debug and some don't. I didn't change any of those (I think) but we probably want to be more consistent.

Cargo.toml
```diff
@@ -40,6 +40,7 @@ byteorder = { version = "1", default-features = false }
 bytes = { version = "1.9.0" }
 cfg-if = { version = "1", default-features = false }
 const_format = { version = "0.2", default-features = false }
+derive-where = { version = "1" }
```
elagergren-spideroak, line 43: `default-features = false`, just in case

crates/aranya-crypto/src/aqc/bidi.rs
```diff
@@ -413,8 +408,10 @@ struct PskCtx {
 }
 
 /// A PSK for a bidirectional channel.
+#[derive_where(Debug)]
 pub struct BidiPsk<CS> {
     id: BidiPskId,
+    #[derive_where(skip(Debug))]
```
elagergren-spideroak, line 414: We could also just make `RawPSk` implement debug with `RawPsk { ... }`.

elagergren-spideroak: but this is also fine

elagergren-spideroak:
> I noticed some usage of `OnceCell` for IDs gets cloned but some create a new one on clone. Similarly some get evaluated during debug and some don't. I didn't change any of those (I think) but we probably want to be more consistent.

we can fix that here or on another PR, doesn't matter

jdygert-spok:
> > I noticed some usage of `OnceCell` for IDs gets cloned but some create a new one on clone. Similarly some get evaluated during debug and some don't. I didn't change any of those (I think) but we probably want to be more consistent.
> 
> we can fix that here or on another PR, doesn't matter

#299

PR #300, opened 2025-06-17
crypto: update to `spideroak-crypto` v0.6

elagergren-spideroak:
Change some usages of `tuple_hash` to `zerocopy` structs. Whether this is better depends on the KDF. For a KDF like HKDF, this is generally better if the output key size is less than or equal to the digest size. For a KDF like [KMAC], this is almost always better. (NB: It also assumes that all struct fields are fixed size and unambiguous.)

This PR only changes calls to `tuple_hash` that were passed to HPKE. Other usages will be changed in later PRs.

[KMAC]: https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-108r1-upd1.pdf

For #298


crates/aranya-crypto/src/afc/bidi.rs
```diff
@@ -168,47 +167,49 @@ pub struct BidiChannel<'a, CS: CipherSuite> {
...
-                self.their_id.as_bytes(),
-                self.our_id.as_bytes(),
-                &self.label.to_be_bytes(),
-            ],
-        )
+        Info {
+            domain: *b"AfcBidiKeys-v1",
+            parent_cmd_id: self.parent_cmd_id,
+            their_id: self.our_id,
+            our_id: self.their_id,
```
gknopf-aranya, line 197: Are `their_id` and `our_id` swapped here on purpose?

gknopf-aranya (dismissed):

Cargo.lock
```diff
@@ -525,7 +526,7 @@ dependencies = [
  "aranya-runtime",
  "buggy",
  "clap",
- "criterion",
+ "criterion 0.5.1",
```
jdygert-spok, line 529: quic syncer is using a separate version still

gknopf-aranya (dismissed):

jdygert-spok (approved):
nit: many `[] -> iter::once`

jdygert-spok (dismissed):
nit: many `[] -> iter::once`

PR #307, opened 2025-06-18
crypto: do not use `stringify!(...)` as context for generating IDs

### LINKED ISSUE #292: crypto: do not use `stringify!(...)` as context for generating IDs
We currently do [this](https://github.com/aranya-project/aranya-core/blob/f9f7ad6c6bc2d3115defdb9c02738d0a884397a6/crates/aranya-crypto/src/misc.rs#L272).

But that effectively prevents us from renaming the type. E.g., we'd like to have a KEM key specifically for AQC called `Device AQC Key` or whatever.

elagergren-spideroak:
Fixes #292


crates/aranya-crypto/src/afc/bidi.rs
```diff
@@ -218,6 +218,10 @@ pub struct BidiAuthorSecret<CS: CipherSuite> {
 
 sk_misc!(BidiAuthorSecret, BidiAuthorSecretId);
 
+impl<CS: CipherSuite> BidiAuthorSecret<CS> {
+    pub(crate) const CONTEXT: &'static str = "AFC Bidi Author Secret";
+}
```
elagergren-spideroak, line 223: can you make CONTEXT part of `keypair!` (and therefore part of `secret_key!` and `kem_key!`)?

jdygert-spok (approved):

PR #312, opened 2025-06-18
text: create text from cstr

jdygert-spok:


crates/aranya-policy-text/src/text.rs
```diff
@@ -120,6 +121,15 @@ impl TryFrom<String> for Text {
     }
 }
 
+impl TryFrom<&CStr> for Text {
+    type Error = core::str::Utf8Error;
+    fn try_from(value: &CStr) -> Result<Self, Self::Error> {
+        let s: &str = value.to_str()?;
+        // NB: CStr cannot contain nul.
```
gknopf-aranya, line 128: nit: is `nul` supposed to be `null`?

jdygert-spok: Rust uses `nul` when refering to `b'\0'`, e.g. [`from_bytes_with_nul`](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_bytes_with_nul)

Issue #299, opened 2025-06-17
crypto: make usage of oncecell consistent in debug and clone


jdygert-spok:
We are using `OnceCell` to cache computing IDs. When we clone we sometimes clone it and sometimes create a new one. in `Debug` we sometimes force evaluation and sometimes don't.

PR #313, opened 2025-06-18
crypto: "simplify" key pair creation

elagergren-spideroak:
Updates #299

Signed-off-by: Eric Lagergren <elagergren@spideroak.com>

### DIFF
--- a/crates/aranya-crypto/src/afc/bidi.rs
+++ b/crates/aranya-crypto/src/afc/bidi.rs
@@ -295,8 +295,8 @@ impl<CS: CipherSuite> BidiSecrets<CS> {
         let root_sk = RootChannelKey::random(eng);
         let peer = {
             let (enc, _) = hpke::setup_send_deterministically::<CS>(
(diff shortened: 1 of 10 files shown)


jdygert-spok:
There's still some other oncecells for #299 but we can merge this in

elagergren-spideroak:
> There's still some other oncecells for #299 but we can merge this in

yeah, GroupKey and stuff. I've marked it as `Updates` instead of `Fixes`.

jdygert-spok (approved):

elagergren-spideroak:
> > > There's still some other oncecells for #299 but we can merge this in
> > 
> > 
> > yeah, GroupKey and stuff. I've marked it as `Updates` instead of `Fixes`.
> 
> The symmetric stuff like GroupKey and TopicKey will need to happen on another PR.

...because we don't actually use them in Aranya right now

djohnson-spideroak (approved):

PR #323, opened 2025-06-30
policy-compiler: determine type for command serialization and deserialization

### LINKED ISSUE #105: Policy Lang: Enforce that serialize/deserialize can only be used in seal/open
Because the compiler treats seal/open like pure functions, we can erroneously use serialize/deserialize in pure functions. We should track seal/open as their own kind of `StatementContext` so we can enforce this.

jdygert-spok:
Closes #105.

### DIFF
--- a/crates/aranya-policy-vm/src/lib.rs
+++ b/crates/aranya-policy-vm/src/lib.rs
@@ -6,6 +6,7 @@
 #![warn(missing_docs)]
 
 #[cfg(feature = "bench")]
+#[cfg_attr(docsrs, doc(cfg(feature = "bench")))]
 mod bench;
 mod data;
 mod derive;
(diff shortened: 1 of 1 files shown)


jdygert-spok:
Along with #321, the only remaining indeterminate is from stubbing ffi.

apetkov-so (dismissed):

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -677,32 +677,49 @@ impl<'a> CompileState<'a> {
...
+                        _ => {
+                            return Err(
+                                self.err(CompileErrorType::InvalidExpression((**e).clone()))
+                            );
+                        }
                     }
+                    let struct_type = self
+                        .identifier_types
+                        .get(&ident!("this"))
+                        .assume("seal must has `this`")?;
```
apetkov-so, line 694: "must has" => "must have"

apetkov-so (approved):

PR #324, opened 2025-07-02
id: add `trait Identity`

jdygert-spok:


elagergren-spideroak:
I like this.

jdygert-spok:
I agree, just need to figure out what exactly we want for the supertrait bounds.

elagergren-spideroak:
> I agree, just need to figure out what exactly we want for the supertrait bounds.

Well, (unfortunately?) probably most or all. The general idea is that it functions as if it were an `Id`. 

```
    Copy
    + Clone
    + Display
    + Debug
    + Default
    + Hash
    + Eq
    + PartialEq
    + Ord
    + PartialOrd
    + Serialize
    + DeserializeOwned
    + ToBase58<Output = String32>
    + ConstantTimeEq
    + AsRef<[u8]>
    + AsRef<[u8; 32]>
    + FromStr
    + From<Id>
    + Into<Id>
```

I suppose you could drop `AsRef` in favor of `Deref<Target = Id>`. But then you'd probably also want `Borrow<Id>`? :/

PR #327, opened 2025-07-02
id: add `TaggedId<Tag>`

jdygert-spok:


crates/aranya-id/src/id.rs
```diff
@@ -20,117 +21,149 @@ use zerocopy::FromBytes;
...
 )]
-pub struct Id([u8; 32]);
+pub struct TaggedId<Tag> {
+    bytes: [u8; 32],
+    tag: PhantomData<Tag>,
+}
 
-impl Id {
+/// Tag for [`Id`]
+pub struct BaseIdTag;
```
elagergren-spideroak, line 38: 'hide' this in `mod private`?

crates/aranya-id/src/id.rs
```diff
@@ -20,117 +21,151 @@ use zerocopy::FromBytes;
...
-    Ord,
-    PartialOrd,
     zerocopy_derive::Immutable,
     zerocopy_derive::IntoBytes,
     zerocopy_derive::KnownLayout,
     zerocopy_derive::Unaligned,
     zerocopy_derive::FromBytes,
 )]
-pub struct Id([u8; 32]);
+pub struct TaggedId<Tag> {
```
elagergren-spideroak, line 32: maybe this isn't a huge concern, but do we want people to write `type MyId = TaggedId<u64>`? That _sorta_ defeats the purpose IMO. Perhaps we have `impl<Tag: Tag> TaggedId<Tag>` where `pub trait Tag {}` is in `private`? Then `custom_id` could implement `Tag` for the generated tag type.

jdygert-spok: Added `trait IdTag`

crates/aranya-runtime/src/command.rs
```diff
@@ -13,29 +8,6 @@ aranya_crypto::custom_id! {
     pub struct CommandId;
 }
 
-impl CommandId {
```
jdygert-spok, line 16: The main downside to this approach is that you can't `impl` anything for the custom IDs due to orphan rules.

jdygert-spok: The other downside being worse error messages since they expand out the type alias.

elagergren-spideroak: i don't think we have many `impl`s for custom ID types, though

gknopf-aranya:
@jdygert-spok is there an associated issue or description for this PR? I don't have enough context to review these changes.

gknopf-aranya (approved):

chip-so (dismissed):

crates/aranya-id/src/id.rs
```diff
@@ -14,118 +15,157 @@ use spideroak_base58::ToBase58;
...
     pub fn decode<T: AsRef<[u8]>>(s: T) -> Result<Self, ParseIdError> {
         let s32 = spideroak_base58::String32::decode(s).map_err(ParseIdError)?;
         Ok(Self::from_bytes(s32))
     }
+
+    /// Cast between two tagged IDs.
+    #[inline]
+    const fn cast<Other: IdTag>(self) -> Id<Other> {
+        Id::from_bytes(self.bytes)
+    }
```
ssweetney-spok, line 82: Could you add something similar like a `ref_cast`? I noticed in the aranya repo in "crates/aranya-daemon/src/api.rs", there's a lot of lines like `&team.into_id().into()`

crates/aranya-aqc-util/src/testing.rs
```diff
@@ -680,18 +683,18 @@ pub fn test_create_multi_bidi_channels_same_label<T: TestImpl>() {
                 .expect("author should be able to create a bidi channel");
 
             let created = BidiChannelCreated {
-                channel_id: channel_id.into(),
+                channel_id: BidiChannelId::from_base(channel_id),
```
gknopf-aranya: May not be worth updating the AQC code here if we are about to remove it

crates/aranya-aqc-util/src/testing.rs
```diff
@@ -419,14 +419,14 @@ pub fn test_create_bidi_channel<T: TestImpl>() {
         .bidi_channel_created(
             &mut author.eng,
             &BidiChannelCreated {
-                channel_id: channel_id.into(),
+                channel_id: channel_id.from_id(),
```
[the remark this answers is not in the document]
jdygert-spok: This is now `BidiChannelId::from_base(channel_id)`

crates/aranya-id/src/id.rs
jdygert-spok: I am changing all uses to pass IDs by value like we did in core a while ago. So those will become e.g. `TeamId::transmute(team)`. (Although I have a rough plan to change what those IDs look like since they aren't truly public...)

PR #337, opened 2025-07-15
policy-compiler: permit function use before definition

### LINKED ISSUE #336: policy-lang: functions should be allowed to be used before they're defined
For example, `get_assigned_role_id` is defined on line 1747, but is first used on line 902 (separate policy code blocks). This is much too restrictive and means I have to rewrite the policy code in a somewhat non-literate programming manner.

```
Error: unable to init daemon

Caused by:
    0: should be able to compile policy
    1: not defined: get_assigned_role_id at line 902 col 5:
       	let device_role_id = check_unwrap get_assigned_role_id(device_id)

           // At this point we believe the following to be true:
           //
           // - `device_role_id` refers to a role that exists
           // - `device_role_id` refers to the role revoked to
           //   `device_id`
           //
           // We do NOT know whether `device_id` refers to a device
           // that exists.
           //
           // We do NOT know whether `target_role_id` refers to a role
           // that exists.



failures:
    test_add_devices
```

/cc: @chip-so @jdygert-spok @apetkov-so

elagergren-spideroak:
Fixes #336


jdygert-spok (changes requested):
This lets us do unbounded recursion.

elagergren-spideroak:
> This lets us do unbounded recursion.

We can already do it, so it doesn't make things worse. And a real fix is a lot bit more complex :/

PR #338, opened 2025-07-15
Add `ephemeral` keyword to actions and commands and enforce ephemeral/persistent rules at compile time.

### LINKED ISSUE #106: policy-lang: add ephemeral marker in command metadata
From the [MVP](https://github.com/aranya-project/aranya-docs/blob/main/docs/aranya-mvp.md) spec.

cc @chip-so @apetkov-so @ycarmy

apetkov-so:
Closes #106: Commands (and actions) now have a `persistence` field which specifies whether the command will be stored on-graph. This is specified via an optional `ephemeral` keyword, and the compiler asserts that persistent actions can only publish persistent commands, and ephemeral actions only ephemeral commands. E.g.

```
ephemeral command Foo { ... }
command Bar { ... }
ephemeral action foo() {
    publish Foo { ... }
    publish Bar { ... } // error: ephemeral action `foo` cannot publish persistent command `Bar`
}
```

### DIFF
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -1299,22 +1299,23 @@ impl<'a> CompileState<'a> {
                                     )));
                                 }
 
-                                // Check persistence constraints:
-                                // - Persistent actions can publish both ephemeral and persistent commands
-                                // - Ephemeral actions can only publish ephemeral commands
-                                if action.ephemeral {
-                                    let command = self
-                                        .policy
-                                        .commands
-                                        .iter()
-                                        .find(|c| c.identifier == *ident)
-                                        .assume("command must be defined")?;
-                                    if !command.ephemeral {
-                                        return Err(CompileErrorType::InvalidType(format!(
-                                            "Ephemeral action `{}` cannot publish persistent command `{}`",
-                                            action.identifier, ident
-                                        )));
-                                    }
+                                //  Persistent actions can publish only persistent commands, and vice versa.
+                                let command = self
+                                    .policy
+                                    .commands
+                                    .iter()
+                                    .find(|c| c.identifier == *ident)
+                                    .assume("command must be defined")?;
+                                if action.ephemeral != command.ephemeral {
+                                    let (action_type, command_type) = if action.ephemeral {
+                                        ("ephemeral", "persistent")
+                                    } else {
+                                        ("persistent", "ephemeral")
+                                    };
+                                    return Err(CompileErrorType::InvalidType(format!(
+                                        "{} action `{}` cannot publish {} command `{}`",
+                                        action_type, action.identifier, command_type, ident
+                                    )));
                                 }
                                 Ok(nty)
                             }
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -1306,11 +1306,10 @@ impl<'a> CompileState<'a> {
                                     .iter()
                                     .find(|c| c.identifier == *ident)
                                     .assume("command must be defined")?;
-                                if action.ephemeral != command.ephemeral {
-                                    let (action_type, command_type) = if action.ephemeral {
-                                        ("ephemeral", "persistent")
-                                    } else {
-                                        ("persistent", "ephemeral")
+                                if action.persistence != command.persistence {
+                                    let (action_type, command_type) = match action.persistence {
+                                        ast::Persistence::Ephemeral => ("ephemeral", "persistent"),
+                                        ast::Persistence::Persistent => ("persistent", "ephemeral"),
                                     };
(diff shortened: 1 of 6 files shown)


apetkov-so:
Request for comments... Sessions now use Ephemeral mode, but that breaks a bunch of tests. I guess we have to mark the test commands as ephemeral.

crates/aranya-runtime/src/testing/protocol.rs
```diff
@@ -301,6 +302,7 @@ impl Policy for TestPolicy {
         action: Self::Action<'_>,
         facts: &mut impl Perspective,
         sink: &mut impl Sink<Self::Effect>,
+        persistence: PersistenceMode, // TODO what should we do with this here?
```
ssweetney-spok, line 305: Can you use this value to determine whether the `facts.add_command` calls should be invoked?

apetkov-so:
> Trying to gain some context. What's causing the tests to break?
> 
> Is it that there's no entry in `aranya_policy_vm::machine::Machine::command_attributes` for those commands?

The lack of `ephemeral: true` in test policies is definitely a cause. I just have to read through the policies more carefully to understand which commands need it.
Also the fact the `TestPolicy` doesn't use the persistence mode. :)

Mostly I want to confirm this is the right direction, though.

apetkov-so:
Ephemeral actions can now publish only ephemeral commands. Persistent actions can publish either.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1250,24 +1250,50 @@ impl<'a> CompileState<'a> {
...
+
+                            // Check persistence constraints:
+                            // - Persistent actions can publish both ephemeral and persistent commands
+                            // - Ephemeral actions can only publish ephemeral commands
+                            if action.ephemeral {
+                                let command = self
+                                    .policy
+                                    .commands
+                                    .iter()
+                                    .find(|c| c.identifier == ident)
```
jdygert-spok, line 1272: Might want to do `command_defs: BTreeMap<Identifier, CommandDef>` with `struct CommandDef { ephemeral: bool, fields: ... }` and look it up there instead.

crates/aranya-policy-lang/src/lang/parse/policy.pest
```diff
@@ -322,20 +322,23 @@ open_block = { "open" ~ statement_block }
...
 // effect Foo { bar id }
 effect_definition = { "effect" ~ identifier ~ effect_struct_def }
 // struct Foo { bar id }
 struct_definition = { "struct" ~ identifier ~ struct_def }
 // enum Result { OK, Err }
 enum_definition = { "enum" ~ identifier ~ "{" ~ identifier ~ ("," ~ identifier)* ~ ","? ~ "}" }
 // command Foo { ... }
-command_definition = { "command" ~ identifier ~ "{" ~ command_blocks+ ~ "}" }
+command_def = { identifier ~ "{" ~ command_blocks+ ~ "}" }
+command_definition = { ephemeral_modifier? ~ "command" ~ command_def }
```
jdygert-spok, line 341: Is it necessary to split these up?

crates/aranya-policy-compiler/src/compile.rs
apetkov-so: We'd have to change `CommandDefinition` to derive `Eq` (to satisfy `CompileTarget`'s attributes). That requires `Eq` on a whole bunch of ast structs, e.g.`Expression`, `FieldDefinition`, `Statement`, `NamedStruct`, etc.

crates/aranya-policy-lang/src/lang/parse.rs
```diff
@@ -1278,7 +1288,17 @@ impl ChunkParser<'_> {
         assert_eq!(item.as_rule(), Rule::command_definition);
 
         let locator = self.add_range(&item)?;
+
         let pc = descend(item);
+        let token = pc.consume()?;
+        let ephemeral = token.as_rule() == Rule::ephemeral_modifier;
```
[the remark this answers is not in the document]
jdygert-spok:
```rust
impl PairContext<'_> {
    fn consume_optional(&self, rule: Rule) -> Option<Pair<'_, Rule>> {
        self.peek()
            .filter(|p| p.as_rule() == rule)
            .inspect(|_| _ = self.next())
    }
}
```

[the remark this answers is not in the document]
apetkov-so: Ah, I thought `self` referred to `ChunkParser`. Updated `parse_command_definition` and `parse_action_definition`, but can't seem to get it working with `parse_fact_definition`.

crates/aranya-policy-compiler/src/compile.rs
apetkov-so: Changed the `ephemeral: bool` field to `persistence: Persistence`.

But for CommandDef, I'm not sure it's worth adding another type just to speed up command lookups.

PR #340, opened 2025-07-18
Fix #332: Show corect line number for errors in policy chunks.

### LINKED ISSUE #332: policy-compiler: "reserved identifier" error throws away line number
Somewhere in my policy file I used `id` as an identifier. The compiler emits this error

```
Reserved identifier: line 6 column 39: id: Reserved identifier: id
```

but the error is not actually at line 6 of the markdown document, it was at line 6 of the specific policy code block. My policy file is ~4k lines long, finding this is quite difficult (even with regex).

/cc: @jdygert-spok @apetkov-so @chip-so

apetkov-so:
Error line numbers now account for starting lines of policy chunks. E.g.

```
---
policy-version: 2
---

'''policy
    let int = 0
'''

Reserved identifier: line 6 column 9: int
```

### DIFF
--- a/crates/aranya-policy-lang/src/bin/parser-explorer/main.rs
+++ b/crates/aranya-policy-lang/src/bin/parser-explorer/main.rs
@@ -71,8 +71,7 @@ fn parse_thing(s: &str, args: &Args) -> anyhow::Result<String> {
             let token = pairs.next().context("No tokens")?;
 
             let pratt = get_pratt_parser();
-            let start = Point::new(0, 0, 0);
-            let mut p = ChunkParser::new(&start, &pratt);
+            let mut p = ChunkParser::new(0, &pratt);
             let ast = p.parse_expression(token)?;
 
             Ok(format!("{:#?}", ast))
--- a/crates/aranya-policy-lang/src/bin/parser-explorer/main.rs
+++ b/crates/aranya-policy-lang/src/bin/parser-explorer/main.rs
@@ -9,7 +9,6 @@ use aranya_policy_lang::lang::{
     ChunkParser, PolicyParser, Rule, Version, extract_policy, get_pratt_parser, parse_policy_str,
 };
 use clap::{Parser, ValueEnum};
-use markdown::unist::Point;
 use pest::Parser as PestParser;
 
 #[derive(Parser, Debug)]
--- a/crates/aranya-policy-lang/src/bin/parser-explorer/main.rs
+++ b/crates/aranya-policy-lang/src/bin/parser-explorer/main.rs
@@ -9,6 +9,7 @@ use aranya_policy_lang::lang::{
     ChunkParser, PolicyParser, Rule, Version, extract_policy, get_pratt_parser, parse_policy_str,
 };
 use clap::{Parser, ValueEnum};
+use markdown::unist::Point;
 use pest::Parser as PestParser;
 
 #[derive(Parser, Debug)]
@@ -70,7 +71,8 @@ fn parse_thing(s: &str, args: &Args) -> anyhow::Result<String> {
             let token = pairs.next().context("No tokens")?;
 
             let pratt = get_pratt_parser();
-            let mut p = ChunkParser::new(0, &pratt);
+            let start = Point::new(0, 0, 0);
+            let mut p = ChunkParser::new(&start, &pratt);
             let ast = p.parse_expression(token)?;
 
             Ok(format!("{:#?}", ast))
--- a/crates/aranya-policy-lang/src/lang/parse.rs
+++ b/crates/aranya-policy-lang/src/lang/parse.rs
@@ -149,15 +149,15 @@ fn remain(p: Pair<'_, Rule>) -> PairContext<'_> {
 
 /// Context information for partial parsing of a chunk of source
 pub struct ChunkParser<'a> {
-    start: &'a Point,
+    offset: usize,
     text_ranges: ast::TextRanges,
     pratt: &'a PrattParser<Rule>,
 }
 
-impl<'a> ChunkParser<'a> {
-    pub fn new(start: &'a Point, pratt: &'a PrattParser<Rule>) -> ChunkParser<'a> {
+impl ChunkParser<'_> {
+    pub fn new(offset: usize, pratt: &PrattParser<Rule>) -> ChunkParser<'_> {
         ChunkParser {
-            start,
+            offset,
             text_ranges: vec![],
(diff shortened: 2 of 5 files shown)


crates/aranya-policy-lang/src/lang/parse/error.rs
```diff
@@ -50,28 +50,39 @@ pub enum ParseErrorKind {
...
+    }
+
+    /// Return a new error with a location starting from the given line.
+    pub fn adjust_line_number(&self, start_line: usize) -> ParseError {
+        ParseError {
+            kind: self.kind.clone(),
+            message: self.message.clone(),
+            location: self
+                .location
+                .map(|(line, col)| (line.saturating_add(start_line), col)),
```
apetkov-so, line 75: Using `saturating_add` feels a bit sketchy, but the effort to handle the `Option<Result<>>` doesn't seem worth it here. How likely are we to ever overflow a `usize` number of lines?

crates/aranya-policy-lang/src/lang/parse.rs
```diff
@@ -1538,10 +1538,17 @@ pub fn parse_policy_chunk(
         ));
     }
     let chunk = PolicyParser::parse(Rule::file, data)
-        .map_err(|e| mangle_pest_error(offset, &policy.text, e))?;
+        .map_err(|e| mangle_pest_error(start_line, &policy.text, e))?;
```
jdygert-spok, line 1541: I think mangle_pest_error is now wrong since it wasn't changed to expect a line number.

crates/aranya-policy-lang/src/lang/parse/error.rs
jdygert-spok, line 1: `impl From<PestError<Rule>> for ParseError` should be updated to set the `location`.

crates/aranya-policy-lang/src/lang/parse.rs
```diff
@@ -167,11 +167,11 @@ impl ChunkParser<'_> {
...
             .start()
-            .checked_add(self.chunk_offset)
+            .checked_add(self.start_line)
             .assume("start + offset must not wrap")?;
         let end = span
             .end()
-            .checked_add(self.chunk_offset)
+            .checked_add(self.start_line)
             .assume("end + offset must not wrap")?;
         self.text_ranges.push((start, end));
```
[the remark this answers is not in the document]
apetkov-so: If we tracked chunk starts using `Position`s instead of character/line offsets, we should be able to safely compute line/col positions, I think.

crates/aranya-policy-lang/src/bin/parser-explorer/main.rs
```diff
@@ -70,7 +71,8 @@ fn parse_thing(s: &str, args: &Args) -> anyhow::Result<String> {
             let token = pairs.next().context("No tokens")?;
 
             let pratt = get_pratt_parser();
-            let mut p = ChunkParser::new(0, &pratt);
+            let start = Point::new(0, 0, 0);
```
apetkov-so, line 74: Ugh, this syntax is pretty ugly.

crates/aranya-policy-lang/src/lang/parse/error.rs
apetkov-so: I'm not sure how to construct the span... We don't have access to the policy string, and without it, the span range would be invalid.

jdygert-spok: Oh right, I remember mentioning in one of those old PRs it should be changed to an explicit conversion so you can supply the source text.

crates/aranya-policy-lang/src/lang/parse.rs
apetkov-so: Reverted to using offset, and passing `start.offset`, where start is a `Point`.

crates/aranya-policy-lang/src/lang/parse/error.rs
jdygert-spok: Actually, that was for another reason. This should work just fine:

```rust
impl From<PestError<Rule>> for ParseError {
    fn from(e: PestError<Rule>) -> Self {
        let p = match e.line_col {
            LineColLocation::Pos(p) => p,
            LineColLocation::Span(p, _) => p,
        };
        ParseError {
            kind: ParseErrorKind::Syntax,
            message: e.to_string(),
            location: Some(p),
        }
    }
}
```

jdygert-spok:
LGTM after last 2 comments.

crates/aranya-policy-lang/src/lang/parse/markdown.rs
```diff
@@ -64,18 +63,10 @@ fn extract_policy_from_markdown(node: &Node) -> Result<(Vec<PolicyChunk>, Versio
...
-                        // The starting position of the code block is
-                        // the triple-backtick, so add three for the
-                        // backticks, six for the language tag, and
-                        // one newline.
-                        let offset = position
-                            .start
-                            .offset
-                            .checked_add(10)
-                            .assume("start.offset + 10 must not wrap")?;
+                        let start_line = position.start.line;
```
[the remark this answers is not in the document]
apetkov-so: I don't think we need this comment anymore... we're just passing the `position.start` without any arithmetic.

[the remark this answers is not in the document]
jdygert-spok: Actually I think this is mangling the pest errors wrong now since we aren't fixing the offset.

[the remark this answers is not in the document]
jdygert-spok: I added a test to cover that and fixed it.

crates/aranya-policy-lang/src/lang/parse/markdown.rs
```diff
@@ -37,7 +37,15 @@ fn parse_front_matter(yaml: &Yaml) -> Result<Version, ParseError> {
...
 pub struct PolicyChunk {
     pub text: String,
-    pub offset: usize,
+    pub start: ChunkOffset,
+}
+
+#[derive(Default, Debug)]
+pub struct ChunkOffset {
+    /// 0-based line offset of policy code within document.
+    pub line: usize,
```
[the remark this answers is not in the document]
jdygert-spok: Yeah, this is used to adjust the line number stored in the error, which is just used for the display impl.

Now that you mention it, this wouldn't be necessary with `miette` since it computes the line numbers from the spans when it grabs the snippets to render the error report.

PR #345, opened 2025-07-23
syncer: fix syncing with multiple messages

jdygert-spok:
Fix syncing with multiple messages and clean up the syncer, including removing the lifetime from the requester.


jdygert-spok:
These are the syncer changes from embedded-wip. (Not the small constants).

benz-spideroak (approved):

PR #347, opened 2025-07-31
policy-compiler: remove `Typeish::Probably` and use `Never` type

jdygert-spok:
The only remaining uses of indeterminate were both "panics" so it should really be a "never" type, which can unify with anything but will never be actually instantiated. I also added a panic instruction when ffi is stubbed instead of just doing nothing which would cause weird things to happen.


ssweetney-spok (approved):
👍

jdygert-spok:
Based on "the VM should still check invariants in case of corruption or something even more bizarre", no.

ssweetney-spok (dismissed):

crates/aranya-policy-compiler/src/compile/types.rs
```diff
@@ -157,42 +157,33 @@ impl IdentifierTypeStack {
...
 impl Typeish {
-    /// Is this an instance of this type or an Indeterminate value? Indeterminate types
-    /// always match.
+    /// Is this an instance of this type or a `Never` value?
     pub fn fits_type(&self, ot: &VType) -> bool {
         match self {
-            Self::Definitely(t) | Self::Probably(t) => t.fits_type(ot),
-            Self::Indeterminate => true,
+            Self::Known(t) => t.fits_type(ot),
+            Self::Never => true,
```
apetkov-so, line 180: Why don't we return `false` here? A `Never` cannot fit a non-`Never`.

jdygert-spok: `Never` can fit any type. This is also how the never type `!` works in rust. This lets e.g `my_function(todo())` typecheck.

apetkov-so (dismissed):

ssweetney-spok (approved):

PR #348, opened 2025-08-01
policy-compiler: Check for duplicate identifiers in current scope

### LINKED ISSUE #174: Detect identifier reuse at compile time
The following should not compile:
```
action foo(n int) {
    let n = n
}
```
It fails at runtime with `name 'n' already defined`, but it should be caught at compile time.

ssweetney-spok:
Resolves #174


crates/aranya-policy-compiler/src/compile/types.rs
```diff
@@ -95,15 +95,15 @@ impl IdentifierTypeStack {
...
             if prev.contains_key(&ident) {
                 return Err(CompileErrorType::AlreadyDefined(ident.to_string()));
             }
         }
         let block = locals.last_mut().expect("no block scope");
         match block.entry(ident) {
-            hash_map::Entry::Occupied(o) => {
-                let _type: Typeish = o.get().clone().unify(value)?; // can we disallow this yet?
+            hash_map::Entry::Occupied(_) => {
+                unreachable!();
```
ssweetney-spok, line 106: Following up on this [thread](https://github.com/aranya-project/aranya-core/pull/321#discussion_r2178411398). I think this is OK? If/match statements use block scoping so things should work as expected.

jdygert-spok (approved):

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -1909,49 +1909,65 @@ fn test_duplicate_definitions() {
...
+            "#,
+            e: None,
+        },
         Case {
             t: r#"
                 function f() bool {
-                    // this will fail at runtime but is allowed by the
-                    // compiler because they are the same type
                     let x = 3
                     let x = 4
```
gknopf-aranya, line 1938: Does the policy support block scopes? If so, this is a case that could be tested:
```
let x = 3
{
    let x = 4
}
```

gknopf-aranya (approved):

crates/aranya-policy-compiler/src/tests.rs
jdygert-spok:
```
let x = {
    let x = 1
    : x
}

File: crates/aranya-policy-compiler/src/tests.rs [hunk 3]
```diff
@@ -1904,6 +1904,19 @@ fn test_duplicate_definitions() {
...
+                    let x = {
+                        let x = 1
+                        : x
+                    }
+
+                    return x
+                }
+            "#,
+            e: None,
+        },
```
jdygert-spok, line 1919: Hmm, I had expected this to fail, but I guess there's technically no shadowing. It's similar to this:

```
let y = {
    let x = 1
    : x
}
let x = y
```

jdygert-spok (approved):

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -1904,54 +1904,114 @@ fn test_duplicate_definitions() {
...
+            t: r#"
+                action foo(n int) {
+                    let n = n
+                }
+            "#,
+            e: Some(CompileErrorType::AlreadyDefined(String::from('n'))),
+        },
+        Case {
+            t: r#"
+                function f(b bool) int {
```
gknopf-aranya, line 1987: This case may not be needed since the other test already fails. I wasn't sure how the compiler was going to handle scope inside the if condition and which cases might fail.
```
let x = 4
if (b) {
    let x = 4
} else {
    let x = 4
}
```

jdygert-spok: Yeah this just fails in the same way as the case below.

ssweetney-spok: > Is the compiler scanning code from top to bottom so it won't process the let x = 4 after the if/else if compilation already failed inside the if/else?

The compiler currently just stops at the first error so it won't pick up on any errors that occur later in the program. So yea, the last `let x=4` statement _should_ be an error but the compiler stops before it gets there

gknopf-aranya: I see. We can probably take that test case out then since the compiler won't even process that line.

apetkov-so (approved):

PR #351, opened 2025-08-04
chore: bump policy lang/compiler/vm/etc versions

elagergren-spideroak:


crates/aranya-aqc-util/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "aranya-aqc-util"
 description = "Utilities for Aranya QUIC Channels"
-version = "0.6.0"
+version = "0.12.0"
```
gknopf-aranya, line 4: Why is the version jumping from `0.6.0` to `0.12.0` instead of to `0.7.0`?

elagergren-spideroak: I goofed the release-plz invocation, am going to fix the PR in a bit

jdygert-spok:
Closing in favor of #378.

PR #352, opened 2025-08-04
Add ActionAnalyzer

### LINKED ISSUE #178: policy-compiler: add analyzer to ensure action publishes a command
Nice-to-have to detect errors.

cc: @chip-so

ssweetney-spok:
Resolves #178

### DIFF
--- a/crates/aranya-policy-compiler/src/tests.rs
+++ b/crates/aranya-policy-compiler/src/tests.rs
@@ -2125,6 +2125,116 @@ fn test_validate_return() {
 }
 
 #[test]
+fn test_validate_publish() {
+    let concat = |text| {
+        let base = r#"
+            command Foo {
+                fields {
+                    a int
+                }
+                seal { return todo() }
+                open { return todo() }
+                policy {
+                    finish {}
+                }
+                recall {
+                    finish {}
+                }
+            }
+        "#;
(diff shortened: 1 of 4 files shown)


crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -2124,6 +2130,116 @@ fn test_validate_return() {
...
+        ),
+        concat(
+            r#"
+            action e() {
+                let n = 0
+                match n {
+                    0 => { publish Foo { a: 0 } }
+                    _ => { publish Foo { a: 1 } }
+                }
+            }
```
ssweetney-spok, line 2202: Rewriting this as:

```suggestion
            action e(n int) {
                match n {
                    0 => { publish Foo { a: 0 } }
                    _ => { publish Foo { a: 1 } }
                }
            }
```

causes the ValueAnalyzer to fail 😕

jdygert-spok: Looks like we shouldn't add the action args to `predefined`, like we don't for functions.

ssweetney-spok: Ah, that makes sense. Removing that causes the 2 invalid cases I wrote to now become valid. Looking into that now...

PR #171, opened 2025-03-20
vm-policy: implement finalization priority

jdygert-spok:
Closes #215 and closes #169.

PR #354, opened 2025-08-05
runtime: require command priority

### LINKED ISSUE #343: policy-compiler: priority attribute should be mandatory
Currently, commands have a default priority of 0. Command priorities should be required and the policy should fail to compile if it is missing.

jdygert-spok:
Persistent commands now require exactly one of `init: true`, `finalize: true`, or `priority: 42`. Ephemeral commands must have none of those.

Closes #343.

### DIFF
--- a/crates/aranya-runtime/src/vm_policy.rs
+++ b/crates/aranya-runtime/src/vm_policy.rs
@@ -129,7 +129,8 @@ use spin::Mutex;
 use tracing::{error, info, instrument};
 
 use crate::{
-    ActionPlacement, CommandPlacement, FactPerspective, MergeIds, Perspective, Prior, Priority,
+    ActionPlacement, Address, CommandPlacement, FactPerspective, MergeIds, Perspective, Prior,
+    Priority,
     command::{CmdId, Command},
     engine::{EngineError, NullSink, Policy, Sink},
 };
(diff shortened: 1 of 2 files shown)


gknopf-aranya:
@jdygert-spok should we add a test that confirms a policy without defined priority attributes for each command fails to compile?

jdygert-spok:
> @jdygert-spok should we add a test that confirms a policy without defined priority attributes for each command fails to compile?

Yes, that goes in the test I need to edit that is ignored.

> Should the priority be moved from the attributes section and directly into the command definition since it's required?

In this implementation it is only required by `aranya_runtime::VmPolicy`, not the compiler.

gknopf-aranya:
> Yes, that goes in the test I need to edit that is ignored.

I think we should test with 2 nearly identical policies. One policy has command priorities defined, while the other does not.
The test would confirm that the first policy compiles, while the second policy does not.

crates/aranya-model/src/tests/basic-policy.md
```diff
@@ -55,6 +55,10 @@ action init(nonce int) {
 
 // `Init` is a command that initializes a graph.
 command Init {
+    attributes {
+        init: true,
```
apetkov-so, line 59: If this is a persistent command, shouldn't it be marked as `persistent`?

As a policy writer, I think I would find this whole block confusing. Which key:value paris are required? What they do? How do they interact with the `persistent` keyword`? Why not just add them implicitly, if they're always required?

crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -229,14 +229,53 @@ fn get_command_priorities(
...
+            Persistence::Persistent => {
+                priority_map.insert(name.clone(), get_command_priority(name, &attrs)?);
+            }
+            Persistence::Ephemeral { .. } => {
+                if attrs != PriorityAttrs::default() {
+                    return Err(AttributeError(
+                        "ephemeral command must not have priority".into(),
+                    ));
+                }
+            }
```
ssweetney-spok: They[Ephemeral commands] don't need priorities but is it really an error if they do have a priority? I'd guess that indicate the policy writer misunderstanding the purpose

jdygert-spok: Yeah, it doesn't _need_ to be an error, but since they wouldn't be used it seems quite misleading to have them.

crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -229,14 +229,53 @@ fn get_command_priorities(
...
+        }
+    }
+    Ok(priority_map)
+}
+
+#[derive(Default, PartialEq)]
```
gknopf-aranya: +Debug?

crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -819,9 +872,51 @@ mod test {
...
+                attributes {}
+                fields {}
+                seal { return todo() }
+                open { return todo() }
+                policy {}
+            }"#,
+            r#"command Test {
+                attributes {
+                    init: false,
+                    finalize: false,
```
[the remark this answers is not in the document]
jdygert-spok: If you're just referring to checking to checking that `finalize: true` sets the priority correctly, we have that below in `test_get_command_priorities`.

If you mean there was no full integration test added back in #171, that's true. Opened #454.

crates/aranya-runtime/src/vm_policy.rs
jdygert-spok: It's private, we can add as needed.

crates/aranya-model/src/tests/basic-policy.md
jdygert-spok: This complexity is already here based on the priorities in the runtime. Just seems like we need to document it https://github.com/aranya-project/aranya-docs/issues/141

PR #370, opened 2025-08-15
fix update statement

### LINKED ISSUE #369: policy-compiler: fix update statements that supply "from" values
Update statements that supply "from" values should validate that the existing fact matches those values.

```
// Can replace 100 with any int
update FooCounter[device: author]=>{count: 100} to {count: new_count}
```

[Update spec](https://github.com/aranya-project/aranya-docs/blob/main/docs/policy-v1.md#update)

ssweetney-spok:
closes #369

### DIFF
--- a/crates/aranya-policy-vm/src/machine.rs
+++ b/crates/aranya-policy-vm/src/machine.rs
@@ -754,7 +754,7 @@ where
             }
             Instruction::Update => {
                 let fact_to: Fact = self.ipop()?;
-                let fact_from: Fact = self.ipop()?;
+                let mut fact_from: Fact = self.ipop()?;
                 let replaced_fact = {
                     let mut iter = self
                         .io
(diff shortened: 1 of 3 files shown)


crates/aranya-policy-vm/src/machine.rs
```diff
@@ -765,6 +765,21 @@ where
...
+                if !fact_from.values.is_empty() {
+                    let replaced_fact_values = &mut replaced_fact.1;
+
+                    replaced_fact_values
+                        .sort_unstable_by(|v1, v2| v1.identifier.cmp(&v2.identifier));
+                    fact_from
+                        .values
+                        .sort_unstable_by(|v1, v2| v1.identifier.cmp(&v2.identifier));
+
+                    if replaced_fact_values.as_slice() != fact_from.values.as_slice() {
```
apetkov-so, line 778: If we use unstable sorts, doesn't that make the comparison unreliable?

jdygert-spok: It means that two of the same identifier won't necessarily stay in their original order, but we shouldn't have two of the same identifier anyway.

PR #375, opened 2025-08-18
policy-compiler: limit key wildcards

jdygert-spok:


ssweetney-spok (approved):

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1426,6 +1426,15 @@ impl<'a> CompileState<'a> {
                         );
                     }
 
+                    if s.fact.key_fields.iter().any(|f| f.1 == FactField::Bind) {
+                        return Err(self.err_loc(
+                            CompileErrorType::BadArgument(String::from(
```
gknopf-aranya, line 1431: Should we define a const error string for this `BadArgument` error variant?

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1469,6 +1478,15 @@ impl<'a> CompileState<'a> {
                     self.append_instruction(Instruction::Update);
                 }
                 (ast::Statement::Delete(s), StatementContext::Finish) => {
+                    if s.fact.key_fields.iter().any(|f| f.1 == FactField::Bind) {
```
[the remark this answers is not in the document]
jdygert-spok: Nah, it's simple and only twice.

crates/aranya-policy-compiler/src/compile.rs
jdygert-spok: No, this is how the compiler errors are generally done. If anything it could be a more specific enum variant but this is fine for now.

gknopf-aranya:
> It doesn't, the policy needs to be updated.

Then the default policy would also need to be updated today if we wan't to merge this PR in today before the release.

PR #379, opened 2025-08-19
Repo Cleanup

nikki-aranya:


jdygert-spok:
This is a lot of changes for one PR, and I don't think I agree with all of them, like no pub(crate) in private. It's hard to tell what lints are being used and to pick out the more questionable ones.

Let's start with the nightly fixes and some additional lints that are hopefully uncontroversial. IMO these are all fine:
- use `Self`
- semicolons
- inline format args
- trait as _
- partialeq with eq

Ideally each lint would have a commit that enables it in the workspace config and fixes the instances of it.

nikki-aranya:
> This is a lot of changes for one PR, and I don't think I agree with all of them, like no pub(crate) in private. It's hard to tell what lints are being used and to pick out the more questionable ones.
> 
> Let's start with the nightly fixes and some additional lints that are hopefully uncontroversial. IMO these are all fine:
> 
>     * use `Self`
> 
>     * semicolons
> 
>     * inline format args
> 
>     * trait as _
> 
>     * partialeq with eq
> 
> 
> Ideally each lint would have a commit that enables it in the workspace config and fixes the instances of it.

Well, the goal for this was to get the discussion started on all this, there's a handful of lints I wrote down that we might want to use but I didn't include here: `missing_inline_in_public_items, std_instead_of_core, std_instead_of_alloc, exhaustive_structs, exhaustive_enums, undocumented_unsafe_blocks, unused_trait_names`

PR #380, opened 2025-08-20
runtime: check persistence when calling actions

jdygert-spok:
Ensures that persistent actions are used on-graph and ephemeral actions are used off-graph.

I changed the API to use `ActionPlacement` and `CommandPlacement` instead of `Persistence` and `OnRecall`. I think it's more clear this way to describe how the methods are being called and have the implementations determine the correct behavior from that.

### DIFF
--- a/crates/aranya-model/src/tests/basic-policy.md
+++ b/crates/aranya-model/src/tests/basic-policy.md
@@ -105,6 +105,30 @@ command Create {
     }
 }
 
+ephemeral action create_action_ephemeral(v int) {
+    publish CreateEphemeral {
+        key_a: 1,
+        value: v,
+    }
+}
+
+ephemeral command CreateEphemeral {
+    fields {
+        key_a int,
+        value int,
+    }
+
+    seal { return envelope::do_seal(serialize(this)) }
+    open { return deserialize(envelope::do_open(envelope)) }
+
+    policy {
+        finish {
+            create Stuff[a: this.key_a]=>{x: this.value}
+            emit StuffHappened{a: this.key_a, x: this.value}
+        }
+    }
+}
+
 // The `increment` action takes a value and passes it to the `Increment` command.
 // For simplicity sake, the fact key is hard coded in all our examples.
 action increment(v int) {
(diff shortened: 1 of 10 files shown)


crates/aranya-model/src/tests/basic-policy.md
```diff
@@ -169,7 +221,7 @@ command Decrement {
 }
 
 // `get_stuff` calls the `GetStuff` command with the hardcoded test key.
-action get_stuff() {
+ephemeral action get_stuff() {
```
apetkov-so, line 224: I didn't see any tests calling this action.

jdygert-spok:
> The target branch is not `main`?

No, this needs the persistence added in #387.

crates/aranya-model/src/tests/basic-policy.md
```diff
@@ -261,6 +335,32 @@ command VerifyGreeting {
...
+    fields {
+        key string,
+        value string,
+    }
+
+    seal { return envelope::do_seal(serialize(this)) }
+    open { return deserialize(envelope::do_open(envelope)) }
+
+    policy {
+        let greeting = unwrap query Message[msg: this.key]=>{value: ?}
```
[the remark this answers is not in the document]
jdygert-spok: Not sure, just copied the existing code.

apetkov-so (approved):

crates/aranya-model/src/tests/basic-policy.md
```diff
@@ -179,7 +231,29 @@ action get_stuff() {
 // returns it in a `StuffHappened` effect. As pointed out elsewhere, there is
 // absolutely nothing stopping us from using this command in an on-graph or
 // ephemeral context. We chose to use it strictly in the ephemeral context.
```
gknopf-aranya, line 233: Should these docs be updated? Does the `ephemeral` keyword mean that the command can no longer be published to the graph?

crates/aranya-runtime/src/testing/vm.rs
```diff
@@ -113,15 +113,51 @@ action increment() {
     }
 }
 
-action incrementFour(n int) {
+ephemeral command IncrementEphemeral {
```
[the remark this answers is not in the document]
jdygert-spok: If done on a separate session, yes.

crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -308,6 +309,17 @@ impl<E: aranya_crypto::Engine> VmPolicy<E> {
...
                 ExitReason::Check => {
                     info!("Check {}", self.source_location(&rs));
+
+                    match placement {
+                        CommandPlacement::OnGraphAtOrigin | CommandPlacement::OffGraph => {
+                            // Immediate check failure.
+                            return Err(EngineError::Check);
+                        }
+                        CommandPlacement::OnGraphInBraid => {
+                            // Perform recall.
```
djohnson-spideroak, line 319: intentionally blank?

jdygert-spok: Yes, it falls through to do it below.

elagergren-spideroak (approved):

PR #385, opened 2025-08-26
Enforce More Clippy Lints

nikki-aranya:
This adds a number of lints that are allowed by default to our warnings and fixes all the instances. This should allow our codebase to be a lot more readable at a glance.


Cargo.toml
```diff
@@ -32,6 +32,7 @@ cast_precision_loss = "warn"
 cast_sign_loss = "warn"
 panic = "warn"
 semicolon_if_nothing_returned = "warn"
+uninlined_format_args = "warn"
```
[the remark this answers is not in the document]
nikki-aranya: I could see why it would be annoying if you need to mix and match inlined variables and function calls and such, I'm happy to remove it being enforced.

apetkov-so (approved):

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1407,7 +1402,7 @@ impl<'a> CompileState<'a> {
                 e.identifier.name, e.value.name
             )))
         })?;
-        Ok(Value::Enum(e.identifier.name.to_owned(), *value))
+        Ok(Value::Enum(e.identifier.name.clone(), *value))
```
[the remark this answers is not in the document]
jdygert-spok: I get what you're saying, but it's backwards in a way. `clone` is specifically `&T -> T` while `to_owned` is `&T -> T::Owned`, so `to_owned` gives you less information about what is happening at a glance.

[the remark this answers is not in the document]
ssweetney-spok: From reading the [`ToOwned` docs](https://doc.rust-lang.org/stable/std/borrow/trait.ToOwned.html), `Clone` seems to be the right tool to go from a reference to an owned value (like Jonathan mentioned above).

`ToOwned` is what you'd use when you have a non-owning type like `str`.

I think the blanket impl of:
```rust
impl<T> ToOwned for T
where
    T: Clone
```


makes it slightly confusing since you can use `to_owned` on any type that implements Clone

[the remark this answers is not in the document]
nikki-aranya: Yeah, it falls under `implicit_clone`: https://rust-lang.github.io/rust-clippy/master/index.html#implicit_clone

> These methods do the same thing as _.clone() but may be confusing as to why we are calling to_vec on something that is already a Vec or calling to_owned on something that is already owned.

These objects are already owned, so cloning makes a lot more sense here.

nikki-aranya:
Alright, I rebased this whole mess now that AFC landed. I forced this back to main, copied the lints added to Cargo.toml, ran --fix to autofix what I could, manually fixed the rest, and did a full pass over the changes. I've looked through the entire changelog, just needs final approval.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -342,11 +342,11 @@ impl File {
                 Err(e) => return Err(e.into()),
             }
         }
-        if !buf.is_empty() {
+        if buf.is_empty() {
+            Ok(())
+        } else {
```
jdygert-spok, line 347: There's a bunch of this type of change that I find kinda awkward. I get the point of negating the if, although I think all of these that I've seen would be better suited as `if !x { return Err(...); } Ok(())`. I don't really care enough to make you change them though...

nikki-aranya: if we want that then we should disable the lint which I'm fine doing

crates/aranya-policy-text/src/repr.rs
```diff
@@ -243,7 +243,7 @@ mod arc {
             // See `std::sync::Arc` for details.
             assert!(old <= MAX_REFCOUNT);
 
-            ArcStr { ptr: self.ptr }
+            Self { ptr: self.ptr }
```
djohnson-spideroak, line 246: is Self here referencing ArcStrInner or ArcStr? It looks like the initial code was ArcStr.

apetkov-so (dismissed):

crates/aranya-capi-macro/src/error.rs
```diff
@@ -23,13 +23,7 @@ pub(super) fn derive(item: TokenStream) -> Result<TokenStream> {
...
             .attrs
             .iter()
-            .filter_map(|a| {
-                if a.path().is_ident("repr") {
-                    Some(a.parse_args::<Repr>())
-                } else {
-                    None
-                }
-            })
+            .filter_map(|a| a.path().is_ident("repr").then_some(a.parse_args::<Repr>()))
```
jdygert-spok, line 26:
```suggestion
            .filter_map(|a| a.path().is_ident("repr").then(|| a.parse_args::<Repr>()))
```

jdygert-spok (dismissed):

jdygert-spok (dismissed):

jdygert-spok (approved):

crates/aranya-runtime/src/sync/responder.rs
```diff
@@ -219,7 +214,7 @@ impl<A: Serialize + Clone> SyncResponder<A> {
                         response_cache.add_command(storage, *command, cmd_loc)?;
                     }
                 }
-                self.to_send = SyncResponder::<A>::find_needed_segments(&self.has, storage)?;
+                self.to_send = Self::find_needed_segments(&self.has, storage)?;
```
gknopf-aranya, line 217: What isn't `Self::<A>` needed still?

crates/aranya-quic-syncer/src/lib.rs
```diff
@@ -176,12 +176,12 @@ where
         sink: Arc<TMutex<S>>,
         sender: mpsc::UnboundedSender<GraphId>,
         server_addr: SocketAddr,
-    ) -> Result<Syncer<EN, SP, S>, QuicSyncError> {
+    ) -> Result<Self, QuicSyncError> {
```
gknopf-aranya, line 179: What aren't generics needed here anymore?

gknopf-aranya (approved):

crates/aranya-quic-syncer/src/lib.rs
jdygert-spok: `Self` already has the type parameters instantiated, it is exactly the type specified in the impl block.

crates/aranya-runtime/src/sync/responder.rs
jdygert-spok: `Self` already has the type parameters instantiated, it is exactly the type specified in the impl block.

Cargo.toml
```diff
@@ -25,13 +25,69 @@ unused_lifetimes = "warn"
 unused_qualifications = "warn"
 
 [workspace.lints.clippy]
+# Warns on any kind of bare arithmetic (2 + 2).
```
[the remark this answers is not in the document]
nikki-aranya: I'm planning to dig through the Rust lints in a separate PR but yes, noted.

PR #387, opened 2025-08-28
policy-module: add `NamedMap` and def types for action and command

jdygert-spok:
Alternate approach to #360.

Since all of the maps in the module use identifier keys, I added a `NamedMap` type which is roughly `IndexSet<HashByName<T>>`.

I only added action and command defs for now since those are needed for #354 and #380.

### DIFF
--- a/crates/aranya-policy-ifgen/tests/data/tictactoe.rs
+++ b/crates/aranya-policy-ifgen/tests/data/tictactoe.rs
@@ -54,6 +54,6 @@ pub struct GameUpdate {
 /// Implements all supported policy actions.
 #[actions]
 pub trait ActorExt {
-    fn MakeMove(&mut self, gameID: Id, x: i64, y: i64) -> Result<(), ClientError>;
     fn StartGame(&mut self, players: Players) -> Result<(), ClientError>;
+    fn MakeMove(&mut self, gameID: Id, x: i64, y: i64) -> Result<(), ClientError>;
 }
--- a/crates/aranya-policy-module/src/named.rs
+++ b/crates/aranya-policy-module/src/named.rs
@@ -31,32 +31,32 @@ pub trait Named {
     deserialize = "V: serde::de::DeserializeOwned + Named"
 ))]
 pub struct NamedMap<V> {
-    map: indexmap::IndexSet<ByName<V>, fnv::FnvBuildHasher>,
+    set: indexmap::IndexSet<ByName<V>, fnv::FnvBuildHasher>,
 }
 
 impl<V> NamedMap<V> {
     /// Create an empty map.
     pub const fn new() -> Self {
         Self {
-            map: indexmap::IndexSet::with_hasher(core::hash::BuildHasherDefault::new()),
+            set: indexmap::IndexSet::with_hasher(core::hash::BuildHasherDefault::new()),
         }
     }
 
     /// Return the number of items in the map.
     pub fn len(&self) -> usize {
-        self.map.len()
+        self.set.len()
     }
 
     /// Returns true if the map is empty.
     pub fn is_empty(&self) -> bool {
-        self.map.is_empty()
+        self.set.is_empty()
     }
 
     /// Returns an iterator over the items of the map.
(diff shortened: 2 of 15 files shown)


crates/aranya-policy-module/src/named.rs
```diff
@@ -0,0 +1,129 @@
...
+/// A mapping of named values which preserves insertion order.
+///
+/// `V` must implement [`Named`].
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize)]
+#[serde(transparent)]
+#[serde(bound(
+    serialize = "V: serde::Serialize + Named",
+    deserialize = "V: serde::de::DeserializeOwned + Named"
+))]
+pub struct NamedMap<V> {
```
apetkov-so, line 33: I assume the use of a struct over a type alias is so we can add the serialization attributes to it, instead of having to decorate each use site?

crates/aranya-policy-module/src/named.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+}
+
+impl<V: Named + PartialEq> PartialEq for NamedMap<V> {
+    fn eq(&self, other: &Self) -> bool {
+        self.len() == other.len()
+            && self
+                .map
+                .iter()
+                .all(|x| other.map.get(x.0.name().as_str()).is_some_and(|y| x == y))
```
apetkov-so, line 103: Why don't we need to compare values as well as key names?

crates/aranya-policy-module/src/named.rs
jdygert-spok: It generally lets us control the interface, and swap out the inner details if needed. As an example, this way `ByName` doesn't show up in the interface.

crates/aranya-policy-module/src/named.rs
jdygert-spok: We do compare the values with `x == y`.

This is basically the same as `IndexMap`'s impl. We check that for each entry in `self`, `other` has an entry under that name with an equal value. (and check the length to reject subsets)

apetkov-so: Oh, it's because 'map' is actually a set, whose items contain the name and value.

crates/aranya-policy-vm/src/machine.rs
```diff
@@ -134,17 +134,15 @@ pub struct Machine {
     /// Mapping of Label names to addresses
     pub labels: BTreeMap<Label, usize>,
     /// Action definitions
-    pub action_defs: BTreeMap<Identifier, Vec<ast::FieldDefinition>>,
+    pub action_defs: NamedMap<ActionDef>,
     /// Command definitions
-    pub command_defs: BTreeMap<Identifier, BTreeMap<Identifier, ast::VType>>,
+    pub command_defs: NamedMap<CommandDef>,
     /// Fact schemas
     pub fact_defs: BTreeMap<Identifier, ast::FactDefinition>,
```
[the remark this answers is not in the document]
jdygert-spok: Yes, I think we should. I only did the two that are blocking other PRs since they need the persistence field.

apetkov-so (approved):

crates/aranya-policy-ifgen/tests/data/tictactoe.rs
```diff
@@ -54,6 +54,6 @@ pub struct GameUpdate {
 /// Implements all supported policy actions.
 #[actions]
 pub trait ActorExt {
-    fn MakeMove(&mut self, gameID: Id, x: i64, y: i64) -> Result<(), ClientError>;
     fn StartGame(&mut self, players: Players) -> Result<(), ClientError>;
+    fn MakeMove(&mut self, gameID: Id, x: i64, y: i64) -> Result<(), ClientError>;
```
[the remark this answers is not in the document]
jdygert-spok: It was in sorted order from being in a btreemap, now it is kept in source order.

elagergren-spideroak (approved):

PR #396, opened 2025-09-09
Fix check error returns in open/seal

chip-so:
Mostly modified the `seal` error handling to look like `open`, since that was much better organized. Now a `check` in a `seal` block will return `EngineError::Check` instead of `EngineError::Panic`. `ExitReason::Yield`, which should not happen in `seal` or `open` are bounced as a `bug!()`. And `MachineError`s returned from `call_seal()` are now an `EngineError::InternalError` just like they are in `open`.

### DIFF
--- a/crates/aranya-runtime/src/vm_policy.rs
+++ b/crates/aranya-runtime/src/vm_policy.rs
@@ -404,7 +404,7 @@ impl<E: aranya_crypto::Engine> VmPolicy<E> {
                 }
                 ExitReason::Panic => {
                     info!("Panicked {}", self.source_location(&rs));
-                    Err(EngineError::Check)
+                    Err(EngineError::Panic)
                 }
             },
             Err(e) => {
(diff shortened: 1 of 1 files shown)


crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -637,18 +637,23 @@ impl<E: aranya_crypto::Engine> Policy for VmPolicy<E> {
...
-                            r @ ExitReason::Yield
-                            | r @ ExitReason::Check
-                            | r @ ExitReason::Panic => {
-                                error!("Could not seal command: {}", r);
-                                return Err(EngineError::Panic);
+                        let status = rs_seal.call_seal(command_name.clone(), command_struct);
+                        match status {
+                            Ok(reason) => match reason {
+                                ExitReason::Normal => (),
+                                ExitReason::Yield => bug!("unexpected yield"),
```
gknopf-aranya, line 644: Should this return a more well-defined `EngineError`? Like check and panic cases do?

crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -404,7 +404,7 @@ impl<E: aranya_crypto::Engine> VmPolicy<E> {
                 }
                 ExitReason::Panic => {
                     info!("Panicked {}", self.source_location(&rs));
-                    Err(EngineError::Check)
+                    Err(EngineError::Panic)
```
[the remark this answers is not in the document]
gknopf-aranya: We can pull this branch into a PR on `aranya` to see if it causes anything to break there

gknopf-aranya (approved):
Looks good to me as far as I can tell. A check failure during seal should no longer cause a panic.

crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -637,18 +637,23 @@ impl<E: aranya_crypto::Engine> Policy for VmPolicy<E> {
 
                         let seal_ctx = rs.get_context().seal_from_action(command_name.clone())?;
                         let mut rs_seal = self.machine.create_run_state(&io, seal_ctx);
-                        match rs_seal
-                            .call_seal(command_name.clone(), command_struct)
-                            .map_err(|e| {
-                                error!("Cannot seal command: {}", e);
-                                EngineError::Panic
-                            })? {
```
jdygert-spok, line 645: nit: I'd prefer to keep this part as `.map_err(...)?`, and do the same in `open`.

djohnson-spideroak:
@chip-so LGTM after conflicts resolved

PR #402, opened 2025-09-15
Add get_command_address_at_location

benz-spideroak:


djohnson-spideroak (approved):

crates/aranya-runtime/src/storage/mod.rs
```diff
@@ -221,6 +221,20 @@ pub trait Storage {
     /// Returns the segment at the given location.
     fn get_segment(&self, location: Location) -> Result<Self::Segment, StorageError>;
 
+    /// Returns the address of the command at the given location by first getting the segment,
+    /// then the command, and finally extracting the address.
+    /// This is a convenience method that combines `get_segment`, `Segment::get_command`, and `Command::address`.
+    fn get_command_address_at_location(
+        &self,
+        location: Location,
```
gknopf-aranya, line 229: Is this the location of the segment or of the command?
Assuming someone reading this doesn't know the difference between a `Location` and an `Address`, using the location of a command to lookup the address of a command is confusing because it seems like if you know the `Location` you would already know where the command is located in the graph.

PR #405, opened 2025-09-16
concurrent shm readers

### LINKED ISSUE #399: afc: concurrent shm readers
Add support for concurrent shm readers in the same process.
 
1. Only allow writer to create the shm so there is only one writer while there can be multiple readers
2. Write reader's PID to shm to prevent readers in more than one process. Multiple readers in the same process is okay.
3. Make `AfcState` cloneable. Put a reference count inside the shm to make it cloneable without `Arc`.

cc: @jdygert-spok, @elagergren-spideroak, @djohnson-spideroak, @ssweetney-spok

ssweetney-spok:
Closes #399


crates/aranya-fast-channels/src/shm/write.rs
```diff
@@ -64,7 +73,7 @@ where
         keys: Directed<Self::SealKey, Self::OpenKey>,
         label_id: LabelId,
     ) -> Result<ChannelId, Error> {
-        let mut rng = self.rng.lock().assume("poisoned")?;
+        let mut rng = self.rng.lock().assume("already borrowed")?;
```
ssweetney-spok, line 76: revert

crates/aranya-fast-channels/src/shm/write.rs
```diff
@@ -136,7 +145,7 @@ where
         keys: Directed<Self::SealKey, Self::OpenKey>,
         label_id: LabelId,
     ) -> Result<(), Error> {
-        let mut rng = self.rng.lock().assume("poisoned")?;
+        let mut rng = self.rng.lock().assume("already borrowed")?;
```
ssweetney-spok, line 148: revert

crates/aranya-fast-channels/src/shm/posix.rs
```diff
@@ -137,10 +139,29 @@ impl<T> Mapping<T> {
...
+            l_whence: SEEK_SET
+                .try_into()
+                .assume("`SEEK_SET` is 0 so this should not fail")?,
+            l_start: off_t::try_from(start).assume("`start` should fit inside `off_t`")?,
+            l_len: off_t::try_from(len).assume("`len` should fit inside `off_t`")?,
+            // TODO(Steve): FIXME
+            l_pid: 0,
+        };
+        fcntl_locking(self.fd, F_SETLK, f_lock)
+    }
```
ssweetney-spok, line 164: This is failing with a "bad file descriptor" error

```rust
Errno(Errno { code: 9, description: Some("Bad file descriptor") })
```

Unsure if it's because the file descriptor that's being referenced with `self.fd` is already closed

crates/aranya-fast-channels/src/shm/write.rs
```diff
@@ -41,14 +45,48 @@ where
...
+        let state = State::open(path, flag, mode, max_chans)?;
+        let shm = state.shm();
+        if shm.increment_writer_count() > 0 {
+            #[cfg(feature = "std")]
+            {
+                let pid = shm.get_writer_pid();
+                debug!("Process with ID: {pid}, has an open writer handle");
+            }
+            bug!("More than 1 writer share memory handle open")
+        }
```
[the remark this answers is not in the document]
jdygert-spok: I don't think they are released on process end :/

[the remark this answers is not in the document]
ssweetney-spok: I did a quick search and that seems to be the case, unfortunately. We could have a separate file just to implement locking with `fcntl`?

gknopf-aranya:
@ssweetney-spok it would help me have more context for reviewing the PR if the high level approach was outlined in the PR description

crates/aranya-fast-channels/src/shm/shared.rs
```diff
@@ -697,6 +709,52 @@ impl<CS: CipherSuite> SharedMem<CS> {
...
+    /// Returns the previous count.
+    pub(crate) fn decrement_reader_count(&self) -> usize {
+        self.reader_count.fetch_sub(1, Ordering::SeqCst)
+    }
+}
+
+#[cfg(feature = "std")]
+impl<CS: CipherSuite> SharedMem<CS> {
+    /// Writes the PID of this process to [Self::writer_pid].
+    pub(crate) fn set_writer_pid(&self) {
```
gknopf-aranya, line 734: It may be safer to only allow `set_writer_pid()` and `set_reader_pid()` on corresponding reader/writer types.

ssweetney-spok:
> @ssweetney-spok it would help me have more context for reviewing the PR if the high level approach was outlined in the PR description

Yea, I'll update this once we land on a good approach. I think there's a few ideas floating around at the moment.

This current code on this PR is trying to implement 
1. ref counting in the shared memory
2. advisory file locking on the shared memory object (unable to do this)

PR #410, opened 2025-09-18
afc: store generate `ChannelId`s internally

elagergren-spideroak:


crates/aranya-fast-channels/benches/lib.rs
```diff
@@ -32,7 +32,7 @@ impl Aead for NoopAead {
 
     type KeySize = U16;
     type NonceSize = U16;
-    type Overhead = U0;
+    type Overhead = U16;
```
gknopf-aranya, line 35: Why is the overhead type changing?

crates/aranya-fast-channels/src/memory.rs
```diff
@@ -22,12 +25,18 @@ use crate::{
     state::{AfcState, AranyaState, Directed},
 };
 
+#[derive_where(Debug, Default)]
+struct Inner<CS: CipherSuite> {
+    next_chan_id: u64,
```
gknopf-aranya, line 30: Do we have a plan for what to do with channels in the shm that the client/daemon no longer have a channel ID for?
Are we going to let the number of channels in the shm continue to grow or implement some sort of garbage collection?

crates/aranya-fast-channels/src/shm/shared.rs
```diff
@@ -350,13 +350,16 @@ impl PartialEq<Op> for ChanDirection {
 pub(super) struct ShmChan<CS: CipherSuite> {
     /// Must be [`ShmChan::MAGIC`].
     pub magic: U32,
+    /// Padding.
+    /// TODO(eric): This is only needed for 64-bit systems.
+    pub _pad0: [u8; 4],
```
gknopf-aranya, line 355: I believe this padding would not be needed if we use a `U64` for `magic`.
Are there drawbacks to making the magic number larger?

crates/aranya-fast-channels/src/memory.rs
elagergren-spideroak: That's not something `aranya-fast-channels` can handle itself.

Callers can use `AranyaState::remove_if` to remove unknown channels.

crates/aranya-fast-channels/src/shm/shared.rs
elagergren-spideroak: What is wrong with padding?

crates/aranya-fast-channels/src/shm/error.rs
```diff
@@ -30,6 +30,12 @@ pub enum Error {
     /// The nodes could not be represented in memory.
     #[error(transparent)]
     Layout(#[from] LayoutError),
+    /// The channel already exists.
+    #[error("channel {0} already exists")]
+    AlreadyExists(ChannelId),
```
ssweetney-spok, line 35: Not being used at the moment

crates/aranya-fast-channels/src/testing/util.rs
```diff
@@ -175,23 +193,38 @@ where
...
+
+    /// Returns the [`ChannelId`] for a particular channel.
+    pub fn get_local_channel_id(&self, id: GlobalChannelId) -> Option<ChannelId> {
+        self.try_get_local_channel_id(id)
     }
 
+    /// Returns the [`ChannelId`] for a particular channel.
+    pub fn try_get_local_channel_id(&self, id: GlobalChannelId) -> Option<ChannelId> {
+        self.chans.get(&id).map(|v| v.0)
+    }
```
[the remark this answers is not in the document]
gknopf-aranya: This isn't a suggested change. I was just trying to highlight the code.

gknopf-aranya (approved):

gknopf-aranya (dismissed):

crates/aranya-fast-channels/src/testing/util.rs
```diff
@@ -162,7 +180,7 @@ where
     ident_sk: IdentityKey<CS>,
     enc_sk: EncryptionKey<CS>,
     state: T::Aranya<CS>,
-    chan_labels: HashSet<(ChannelId, LabelId)>,
+    chans: HashMap<GlobalChannelId, (ChannelId, LabelId)>,
```
ssweetney-spok, line 183: This makes more sense. The way I wrote this originally makes it seem like the local channel IDs had to be the same across devices

ssweetney-spok (approved):

crates/aranya-fast-channels/src/state.rs
```diff
@@ -79,42 +87,24 @@ pub trait AranyaState {
...
 
-/// A unique identifier representing a channel.
+/// Uniquely identifies a channel inside the shared state.
+///
+/// This is strictly a local identifier.
+// TODO(eric): Make this 32 bits on 32-bit platforms?
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
 #[repr(transparent)]
-pub struct ChannelId(u32);
+pub struct ChannelId(u64);
```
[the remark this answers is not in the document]
elagergren-spideroak: Not _exactly_. AFC's state is only used on one machine, so it's ~impossible to overflow a 64-bit counter. It's generally trivial to overflow a 32-bit counter, but I'm assuming that we won't overflow if we use 32-bit counters on 32-bit machines since AFC has other overhead (like deriving the seal/open keys) and 32-bit machines typically have slower CPUs.

My comment about making this 32 bits is just to make things nicer for 32-bit machines (single vs double word comparison, etc.). But I don't know if it's worth it. Perhaps there is a situation where (for example) `AfcState` is running on a 32-bit network card and `AranyaState` is running on a 64-bit CPU?

crates/aranya-fast-channels/src/shm/shared.rs
```diff
@@ -350,13 +350,16 @@ impl PartialEq<Op> for ChanDirection {
...
-    pub channel_id: U32,
+    pub channel_id: U64,
     /// The channel's label.
     pub label_id: LabelId,
     /// Describes the direction that data flows in the channel.
     pub direction: U32,
-
+    // TODO(eric): Padding for 64-bit systems?
     /// The current encryption sequence counter.
     pub seq: U64,
```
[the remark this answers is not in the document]
elagergren-spideroak: I think it mattered more when we were using it cross-language. It's not immediately clear why we need it now.

djohnson-spideroak (dismissed):

jdygert-spok (approved):

crates/aranya-fast-channels/src/state.rs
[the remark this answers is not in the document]
jdygert-spok: If we do ever end up doing something weird across arches like that we'll need to revisit the way we do cache alignment.

djohnson-spideroak (approved):

PR #417, opened 2025-09-24
capi: add blanket Typed impl

jdygert-spok:
We can hash the `core::any::TypeId` to get our type ID.

I verified that it optimizes down to a constant. It'd be nice to have some way to validate that remains true.

This ID won't necessarily be consistent across different compiles, but that should be fine in the way we are using it.

We now have a possibility of random collisions. It's not a huge issue since we are just using this for additional safety checks. We could increase the ID size easily, although it'd increase the size of `Safe` unless we shove the flags into either the ID or the address.

This blanket impl only applies for static types. All uses are currently static. I don't think it really makes much sense to use `Safe` with non-static types. Note we can't add manual impls since they would overlap.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -214,6 +214,7 @@ dependencies = [
  "ciborium-io",
  "const_format",
  "elain",
+ "fnv",
  "libc",
  "thiserror 2.0.7",
  "tracing",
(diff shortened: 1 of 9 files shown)


nikki-aranya (approved):

jdygert-spok:
No, the ID is only used in the way I described. It is not used for serialization.

PR #419, opened 2025-10-02
Close #150: Document AST crate

### LINKED ISSUE #150: aranya-policy-ast: improve rustdocs
To support the open source documentation effort, we want to improve the generated rust documentation that gets built and deployed to doc.rs when we publish to crates.io. 

The scope of this story is to update the crate, function, struct, type and enum descriptions, summaries and examples for the aranya-policy-ast crate.

This crate will also needs a README to describe what the crate does.

https://crates.io/crates/aranya-policy-ast

apetkov-so:


djohnson-spideroak (approved):

crates/aranya-policy-ast/README.md
```diff
@@ -0,0 +1,40 @@
...
+let policy = parse_policy_str("policy code", Version::V2)?;
+
+// Compile AST into module
+let module = Compiler::new(&policy).compile()?;
+
+```
+
+## Features
+
+- **`std`** - Enable standard library support (default: disabled)
```
ssweetney-spok, line 28: This line seems a bit vague since it's not clear what users gain from "standard library support". This feature just enables the "std" feature in a few dependencies like `serde` and `thiserror`.

I think it'd be nice to mention that and have a link, https://serde.rs/feature-flags.html#std.

I couldn't find a similar link for `thiserror` but [this PR thread mentions the differences](https://github.com/dtolnay/thiserror/pull/64) between std and no_std

crates/aranya-policy-ast/README.md
```diff
@@ -0,0 +1,40 @@
...
+- **`proptest`** - Enable property testing support for AST nodes
```
jdygert-spok, line 29: These are both not on by default, we should be consistent however we want to notate that.

ssweetney-spok (approved):

PR #421, opened 2025-10-02
Close #418: Add assert statement

### LINKED ISSUE #418: Add assert statement
Add assert statement, to be used outside command policy contexts. Unlike check, assert does not recall commands. Its intent is to enforce preconditions before any commands are published.

Syntax:
```
assert <expression>, "<message>"
```

apetkov-so:
Close #418: Add `assert` statement.

- Add `assert` statement, to be used in actions and functions. This exits with a panic and a reason message to make diagnosing failures easier.
- The check statement is now only valid in command policy blocks. 
- Command recalls should not be fallible anymore since they cannot use checks or asserts.

### DIFF
--- a/crates/aranya-policy-vm/tests/vm.rs
+++ b/crates/aranya-policy-vm/tests/vm.rs
@@ -1704,6 +1704,35 @@ fn test_debug_assert() -> anyhow::Result<()> {
 }
 
 #[test]
+fn test_assert_default_message() -> anyhow::Result<()> {
+    let text = r#"
+    action test() {
+        let x = 1
+        assert x > 10
+    }
+    "#;
+
+    let policy = parse_policy_str(text, Version::V2)?;
+    let io = RefCell::new(TestIO::new());
+    let module = Compiler::new(&policy).compile()?;
+    let machine = Machine::from_module(module)?;
+
+    let action_name = ident!("test");
+    let ctx = dummy_ctx_action(action_name.clone());
+    let mut rs = machine.create_run_state(&io, ctx);
+    let mut published = Vec::new();
+    let result = call_action(&mut rs, &mut published, action_name, iter::empty::<Value>())?;
+    match result {
+        ExitReason::Panic(msg) => {
+            assert_eq!(msg.as_str().trim(), "assertion failed: x > 10");
+        }
+        _ => panic!("Expected Panic exit reason"),
+    }
+
+    Ok(())
+}
+
+#[test]
 fn test_global_let_statements() -> anyhow::Result<()> {
     let text = r#"
         let x = 42
--- a/crates/aranya-policy-compiler/src/tests.rs
+++ b/crates/aranya-policy-compiler/src/tests.rs
@@ -824,10 +824,23 @@ fn test_enum_reference() {
 }
 
 #[test]
+fn test_assert_without_message() {
+    let text = r#"
+        action test() {
+            assert true
+            let x = 1
+            assert x > 0
+        }
+    "#;
+
+    compile_pass(text);
+}
+
+#[test]
 fn test_undefined_fact() {
     let text = r#"
         action test() {
-            assert exists Foo[], "Foo exists"
+            assert exists Foo[], "Foo should exist"
         }
     "#;
 
@@ -840,7 +853,7 @@ fn test_fact_invalid_key_name() {
     let text = r#"
         fact Foo[i int] => {a string}
         action test() {
-            assert exists Foo[k: 1], "Foo exists"
+            assert exists Foo[k: 1], "Foo should exist"
         }
     "#;
 
@@ -856,7 +869,7 @@ fn test_fact_incomplete_key() {
     let text = r#"
         fact Foo[i int] => {a string}
         action test() {
-            assert exists Foo[], "Foo exists"
+            assert exists Foo[], "Foo should exist"
         }
     "#;
 
@@ -872,7 +885,7 @@ fn test_fact_nonexistent_key() {
     let text = r#"
         fact Foo[i int] => {a string}
         action test() {
-            assert exists Foo[i:0, j:1], "Foo exists"
+            assert exists Foo[i:0, j:1], "Foo should exist"
         }
     "#;
 
@@ -888,7 +901,7 @@ fn test_fact_invalid_key_type() {
     let text = r#"
         fact Foo[i int] => {a string}
         action test() {
-            assert exists Foo[i: "1"], "Foo exists"
+            assert exists Foo[i: "1"], "Foo should exist"
         }
     "#;
 
@@ -901,7 +914,7 @@ fn test_fact_duplicate_key() {
     let text = r#"
         fact Foo[i int, j int] => {a string}
         action test() {
-            assert exists Foo[i: 1, i: 2], "Foo exists"
+            assert exists Foo[i: 1, i: 2], "Foo should exist"
         }
     "#;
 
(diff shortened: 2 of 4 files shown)


apetkov-so:
It turns out the reason assert can't parse expressions is that the `cast = { "as" }` rule is greedy-matching the as in assert! (Thanks for figuring that out, Claude!) Renaming "assert" to "azert" fixes the issue.

This is a worrying discovery - who knows how many other weird name collisions there could be! We need a way to specify that string literals are parsed as whole words only. Not sure how to do that yet. @chip-so

apetkov-so:
> So I think your "hack" of adding explicit whitespace to the "as" token is actually the right move. I'd add it to both sides and do the same for "substruct" and other "word" operators.

It doesn't actually work. :/ Adding trailing/leading whitespaces causes syntax errors in legitimate code:
```
Syntax error: line 13 column 48:   --> 13:48
   |
13 |                 return Foo { a: 1, b: "test" } as Bar
   |                                                ^---
```

apetkov-so:
> It doesn't actually work. :/ Adding trailing/leading whitespaces causes syntax errors in legitimate code:

I got it to work by making the as, substruct rules atomic to suppress automatic whitespace insertion.

chip-so:
> > I think that's should be okay, though, because we have the specifically colored functions that can.
> 
> I thought we're adding `command`-colored functions in a future ticket. I mean that's not even spec'ed yet.

Sure. So disallowing `assert` in functions doesn't need to wait on having a result type, it just needs to wait on the action and policy colored functions.

apetkov-so:
> Sure. So disallowing `assert` in functions doesn't need to wait on having a result type, it just needs to wait on the action and policy colored functions.

My notes say we're adding an "action" function color. These can only be called from an action, and are allowed to assert. Is this correct?

I'm not sure what the "policy" colored functions would do, or if we need them at this point.

chip-so:
> My notes say we're adding an "action" function color. These can only be called from an action, and are allowed to assert. Is this correct?

That is correct.

> I'm not sure what the "policy" colored functions would do, or if we need them at this point. Edit: I think policy means "command" (as per #433), but that can be done in the check branch.

Sorry, I mean "command" colored. It's just what we [discussed earlier](https://github.com/aranya-project/aranya-core/discussions/433#discussioncomment-14713139).

jdygert-spok:
> So `seal` and `open` blocks should also allowed to `assert`? @jdygert-spok, @chip-so

I realized we might as well just have them return optional so they are the same as pure functions. (serialize/deserialize can continue to be special cased the way they currently are by looking at signature)

crates/aranya-policy-module/src/instructions.rs
```diff
@@ -33,7 +33,7 @@ pub enum ExitReason {
     /// Execution was aborted gracefully, due an error.
     Check,
     /// Execution was aborted due to an unhandled error.
-    Panic,
+    Panic(Text),
```
jdygert-spok: Should we just expect a text value on the stack, kind of like a yield?

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -55,6 +57,8 @@ struct FunctionSignature {
 pub enum StatementContext {
     /// An action
     Action(ast::ActionDefinition),
+    /// An action function
+    ActionFunction(ast::FunctionDefinition),
```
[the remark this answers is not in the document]
apetkov-so: I would do it in a separate ticket; I'd like to get this merged with the check branch soon.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2043,15 +2108,18 @@ impl<'a> CompileState<'a> {
             }
         }
 
+        // Generate the call instruction
         let label = Label::new(
             fc.identifier.name.clone(),
-            if is_finish {
+            // TODO document why we use Temporary here for finish functions
+            if color == FunctionColor::Finish {
```
ssweetney-spok: The `Function` variant was added in [this commit](https://github.com/aranya-project/aranya-core/commit/e7761b2df2f2680292bf38076aa873ac6a23b615). It looks like `Temporary` is used here so ~~it~~ finish functions get skipped by the function analyzer

ssweetney-spok (dismissed):

crates/aranya-policy-compiler/src/compile.rs
ssweetney-spok: Could also add a `FinishFunction` variant to `LabelType`; that would make it easier to see which analyzers skip those labels in `crates/aranya-policy-compiler/src/validate.rs`.

crates/aranya-policy-vm/tests/vm.rs
```diff
@@ -2383,17 +2388,23 @@ fn test_boolean_short_circuit() {
         exit
     }
 
-    assert_eq!(run("true && todo()"), ExitReason::Panic);
-    assert_eq!(run("false && todo()"), ExitReason::Check);
+    assert_eq!(run("true && todo()"), ExitReason::Panic(text!("todo")));
+    assert_eq!(
+        run("false && todo()"),
+        ExitReason::Panic(text!("false && todo()"))
```
chip-so: Why isn't this just `"todo"`?

crates/aranya-policy-ast/src/ast.rs
```diff
@@ -61,6 +61,37 @@ impl Spanned for Ident {
...
+    Deserialize,
+    rkyv::Archive,
+    rkyv::Deserialize,
+    rkyv::Serialize,
+)]
+pub struct Txt {
+    /// The text value
+    pub text: Text,
+    /// The source location of this text literal
+    pub span: Span,
```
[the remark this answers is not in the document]
apetkov-so: Renamed to SpannedText.

crates/aranya-policy-vm/tests/vm.rs
chip-so: Okay I was confused about how the `run()` function actually works. This is producing the assert message because the expression gets asserted and this one fails. But on the line above, the assert succeeds but it panics on the `todo()` so the message is just "todo".

chip-so (approved):

crates/aranya-policy-ast/src/ast.rs
```diff
@@ -1051,6 +1117,8 @@ pub struct Policy {
     pub functions: Vec<FunctionDefinition>,
     /// The policy's finish function definitions.
     pub finish_functions: Vec<FinishFunctionDefinition>,
+    /// The policy's action function definitions.
+    pub action_functions: Vec<FunctionDefinition>,
```
jdygert-spok: Please mention action functions in the PR title/description

chip-so:
> We might want to add `command function`s before we remove the ability to use `check` in pure functions, so we don't have to migrate code that would prefer to be a command function to using `optional`, and then migrate back.

I think we'll definitely want to do that, yeah.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1462,13 +1468,7 @@ impl<'a> CompileState<'a> {
...
                     self.append_instruction(Instruction::Def(s.identifier.name.clone()));
                 }
-                (
-                    StmtKind::Check(s),
-                    StatementContext::Action(_)
-                    | StatementContext::PureFunction(_)
-                    | StatementContext::CommandPolicy(_)
-                    | StatementContext::CommandRecall(_),
-                ) => {
+                (StmtKind::Check(s), StatementContext::CommandPolicy(_)) => {
```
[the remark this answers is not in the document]
mooreoak: where did we end up on this. It is my recollection that we could still have checks in functions but asserts only in action flavored functions. (or something like that)

PR #423, opened 2025-10-06
AFC client seal methods takes seq number as input

### LINKED ISSUE #422: afc: shm doesn't properly record sequence number
When performing a seal, the sequence number for that channel is incremented. In short, this is only written to the "read" side of the list.

ssweetney-spok:
Resolves #422

**Changes**: AFC client's seal and open methods take "context" objects

### DIFF
--- a/crates/aranya-fast-channels/src/shm/read.rs
+++ b/crates/aranya-fast-channels/src/shm/read.rs
@@ -64,7 +64,7 @@ where
 /// This type is passed to the seal methods on
 /// [AFC Client][crate::client::Client].
 pub struct SealChannelCtx<CS: CipherSuite> {
-    key_generation: Option<(SealKey<CS>, u32)>,
+    key_generation_index: Option<(SealKey<CS>, u32, Index)>,
     label: LabelId,
 }
 
@@ -72,13 +72,13 @@ impl<CS: CipherSuite> SealChannelCtx<CS> {
     /// Creates a new [SealChannelCtx].
     pub fn new(label: LabelId) -> Self {
         Self {
-            key_generation: None,
+            key_generation_index: None,
             label,
         }
     }
 
-    pub(crate) fn key_gen_mut(&mut self) -> &mut Option<(SealKey<CS>, u32)> {
-        &mut self.key_generation
+    fn key_gen_mut(&mut self) -> &mut Option<(SealKey<CS>, u32, Index)> {
+        &mut self.key_generation_index
     }
 
     /// Returns the label ID.
@@ -90,7 +90,7 @@ impl<CS: CipherSuite> SealChannelCtx<CS> {
 impl<CS: CipherSuite> From<LabelId> for SealChannelCtx<CS> {
     fn from(label: LabelId) -> Self {
         Self {
-            key_generation: None,
+            key_generation_index: None,
             label,
         }
     }
@@ -137,7 +137,7 @@ where
 
         let hint = match ctx.key_gen_mut() {
             // There is a cache entry for this channel.
-            Some((key, ctx_generation)) => {
+            Some((key, ctx_generation, ctx_idx)) => {
                 // SAFETY: we only access an atomic field.
                 let generation = unsafe {
                     mutex
@@ -152,11 +152,11 @@ where
 
                     return Ok(f(key, label_id));
                 }
-                // TODO(Steve): Add `Index` to `SealChannelCtx`.
+
                 // The generations are different, so
                 // optimistically use `idx` to try and speed up
                 // the list traversal.
-                None
+                Some(*ctx_idx)
             }
             _ => None,
         };
@@ -169,16 +169,18 @@ where
         // to avoid ownership conflicts with `chan`.
         let generation = list.generation.load(Ordering::Relaxed);
 
-        let (chan, _idx) = match list.find_mut(id, hint, Op::Seal)? {
+        let (chan, idx) = match list.find_mut(id, hint, Op::Seal)? {
             None => return Err(crate::Error::NotFound(id)),
             Some((chan, idx)) => (chan, idx),
         };
 
         let maybe_key_gen = ctx.key_gen_mut();
         let key_gen = match maybe_key_gen {
-            None => {
-                maybe_key_gen.insert((SealKey::from_raw(&chan.seal_key, Seq::ZERO)?, generation))
-            }
+            None => maybe_key_gen.insert((
+                SealKey::from_raw(&chan.seal_key, Seq::ZERO)?,
+                generation,
+                idx,
+            )),
             Some(key) => key,
         };
 


crates/aranya-fast-channels/src/state.rs
```diff
@@ -20,19 +20,94 @@ pub trait AfcState {
...
+        _op: PhantomData<Op>,
+    }
+
+    mod private {
+        pub struct Open;
+        pub struct Seal;
+    }
+
+    pub type SealChannelCtx<CS: CipherSuite> = ChannelCtx<CS, private::Seal>;
+    pub type OpenChannelCtx<CS: CipherSuite> = ChannelCtx<CS, private::Open>;
```
ssweetney-spok: Unsure if we need a "context" struct for the "open" operation but I added it for consistency. I think this should be reverted 🤔

ssweetney-spok: Considering client resumption isn't supported and the client will have all the keys, could AFC re-deisgned so it doesn't have to rely on the daemon as much?

crates/aranya-fast-channels/src/client.rs
```diff
@@ -215,7 +220,7 @@ impl<S: AfcState> Client<S> {
         }
 
         let label_id = self
-            .do_open(local_channel_id, seq, |aead, ad, seq| {
+            .do_open(local_channel_id, ctx, seq, |aead, ad, seq| {
```
gknopf-aranya: If the `ctx` contains the `seq`, why are both `ctx` and `seq` passed in here?

ssweetney-spok: It's the "ctx" used by the "seal" methods that contain the sequence number. The "open" methods still use the sequence number provided in the message header. I'm changing this so only the seal methods have a "ctx" object

ssweetney-spok:
> What is the purpose of `ctx`...

It's suppose to contain the minimal amount of data needed for sealing. This is the type that will be contained by `aranya_client::afc::SendChannel`

> and what information does it contain? Haven't looked through all the code to figure this out yet.

It's a wrapper around `spideroak_crypto::hpke::SealCtx` and the label ID. The channel ID could go in there too but there was some discussion and it seemed simpler to keep it separate.

I should clarify that it's the `spideroak_crypto::hpke::SealCtx` type that actually has the key and sequence number in it

crates/aranya-fast-channels/src/mutex.rs
```diff
@@ -113,21 +113,6 @@ impl<T: ?Sized> Mutex<T> {
...
 
-    /// Returns the data protected by the mutex without any
-    /// synchronization.
-    ///
-    /// # Safety
-    ///
-    /// You must provide your own synchronization. Otherwise,
-    /// doing anything with `T` is UB.
-    #[cfg(any(feature = "posix", feature = "sdlib"))]
-    pub unsafe fn inner_unsynchronized(&self) -> &T {
```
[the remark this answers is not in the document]
ssweetney-spok: Currently unused but it will be brought back once the "OpenChannelCtx" changes are reverted

crates/aranya-fast-channels/src/shm/read.rs
```diff
@@ -138,112 +72,54 @@ where
...
 
-        let (chan, idx) = match list.find_mut(id, hint, Op::Seal)? {
+        // TODO: Don't eagerly search list. Compare to generation
+        let (chan, _idx) = match list.find_mut(id, None, Op::Seal)? {
             None => return Err(crate::Error::NotFound(id)),
             Some((chan, idx)) => (chan, idx),
         };
 
-        let mut key = SealKey::from_raw(&chan.seal_key, chan.seq())?;
+        let label_id = *ctx.label_id();
```
gknopf-aranya: Seems clearer to have `label_id()` return a copy of `LabelID` instead of a reference to it that is dereferenced later. I noticed a push towards using IDs by value in the repos recently so it seems like that would apply here as well. cc: @jdygert-spok 
```suggestion
        let label_id = ctx.label_id();
```
```
        /// Returns the label ID.
        pub fn label_id(&self) -> LabelId {
            self.label
        }
```

crates/aranya-fast-channels/src/shm/read.rs
```diff
@@ -138,112 +72,54 @@ where
...
+        let maybe_key = ctx.seal_mut()?;
+        let mut key = match maybe_key {
+            None => maybe_key.insert(SealKey::from_raw(&chan.seal_key, Seq::ZERO)?),
```
gknopf-aranya: If a channel is deleted from the shared memory, could this cause a channel key to be reinserted with `Seq::ZERO`? We need to make sure a sequence number is never reused with a particular key.

ssweetney-spok: Oh, I see what you mean. Yea, you could misuse this crate if used directly but we'll want to ensure we don't do that in the main Aranya repo

gknopf-aranya: Is there a way to guard this crate API against misuse?

ssweetney-spok: Hmm, I'm not sure how we would do that if we're relying on the sequence number being passed in so it's up to the caller. The primary motivator for having the caller pass in the sequence number in is because the current way of storing/updating the sequence numbers for each channel is bugged.

Edit:

The way to misuse this crate would be to call `aranya_fast_channels::client::Client::seal` with an "empty" channel ctx multiple times for the same channel

gknopf-aranya:
Wonder if we should call `seq` `nonce_seq` to be clear that it cannot be reused and is implemented as a counter rather than a random nonce.

crates/aranya-fast-channels/src/state.rs
```diff
@@ -20,19 +20,102 @@ pub trait AfcState {
     type CipherSuite: CipherSuite;
 
     /// Invokes `f` with the channel's encryption key.
-    fn seal<F, T>(&self, id: LocalChannelId, f: F) -> Result<Result<T, Error>, Error>
+    fn seal<F, T>(
+        &self,
+        id: LocalChannelId,
+        ctx: &mut SealChannelCtx<Self::CipherSuite>,
```
gknopf-aranya: nit: generally `ctx` would be the first parameter passed into a function

crates/aranya-fast-channels/src/testing/mod.rs
```diff
@@ -299,8 +320,24 @@ pub fn test_multi_client<T: TestImpl, A: Aead>() {
...
-                test(&mut clients, &d.devices, *a, *b, label_id, &mut seqs);
-                test(&mut clients, &d.devices, *b, *a, label_id, &mut seqs);
+                test(
+                    &mut clients,
+                    &d.devices,
+                    *a,
+                    *b,
+                    label_id,
+                    &mut seqs,
+                    &mut ctxs,
```
gknopf-aranya: I'm not sure it's a good idea to pass these in separately when there's a 1-1 mapping between `ctx` and `seq`.
Also the `test` uses other parameters passed in to look up the specific `ctx` and `seq` from the maps, so it seems like those elements can be passed in directly. From what I could tell, the test doesn't iterate over the maps or use multiple elements from each, so passing the entry in directly could make more sense.

crates/aranya-fast-channels/src/testing/mod.rs
```diff
@@ -140,7 +142,10 @@ pub fn test_seal_open_basic<T: TestImpl, A: Aead>() {
             let d1_channel_id = d1.get_local_channel_id(global_id).unwrap_or_else(|| {
                 panic!("device {id1} should have channel for global_id {global_id:?}")
             });
-            c1.seal(d1_channel_id, &mut dst[..], GOLDEN.as_bytes())
+
+            let mut ctx = SealChannelCtx::new(label_id);
```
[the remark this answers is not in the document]
ssweetney-spok: > If I call let mut ctx = SealChannelCtx::new(label_id); twice then attempt to seal with each one, would that cause a sequence number to be reused for the same key?

Yep, that's the [misuse](https://github.com/aranya-project/aranya-core/pull/423#discussion_r2457171565) I was referring to.

There are some tests where "seal" is multiple times for a given channel to test sequence number incrementing like "test_monotonic_seq_by_one". Those use map to keep track of the channel contexes

crates/aranya-fast-channels/src/shm/read.rs
gknopf-aranya: What if we group `chan_id` into `ctx` so it's not possible to accidentally initialize a context for the same channel ID twice.
```
let mut ctx = SealChannelCtx::new(chan_id, label_id);
seal(&mut ctx, &mut ciphertext, plaintext);
```

jdygert-spok:
Regarding possible misuse, I don't think there's a way to fully guard against it without either keeping the seq in the shm or writing back some other info from the afc side, which basically gets us back into the same issue. I think we just document it for now and consider #479 once we have more time.

crates/aranya-fast-channels/src/memory.rs
```diff
@@ -56,7 +56,12 @@ where
 {
     type CipherSuite = CS;
 
-    fn seal<F, T>(&self, id: LocalChannelId, f: F) -> Result<Result<T, Error>, Error>
+    fn seal<F, T>(
+        &self,
+        id: LocalChannelId,
+        _: &mut SealChannelCtx<Self::CipherSuite>,
+        f: F,
+    ) -> Result<Result<T, Error>, Error>
```
ssweetney-spok: The "in-memory" implementation of `AfcState` doesn't need to this the context here. I remember we (@jdygert-spok) discussed using an associated type on `AfcState` for the context... I chose not to do that to keep things simple but I wonder if it'd be better in this case since we could use a unit struct here

ssweetney-spok: https://github.com/aranya-project/aranya-core/pull/423/commits/6794be942d54607f3c6cb88d25db6f8357127b24

PR #427, opened 2025-10-08
runtime: move parent ID from envelope to open/policy context

jdygert-spok:


chip-so (approved):

jdygert-spok:
Trying to decide how to expose the `parent_id` to policy code. We could keep `envelope::parent_id()` but it feels funky since it takes from the context instead of the envelope. Not sure there's currently any better ffi. It's _kinda_ associated with the perspective but not really. I don't want to add another ffi...

Issue #425, opened 2025-10-07
afc: remove bidi channels from ffis


ssweetney-spok:
Remove support for creation of:
- AFC bidi channels

PR #428, opened 2025-10-08
Remove AFC Bidi channel related items

ssweetney-spok:
Resoloves #425


crates/aranya-afc-util/src/ffi.rs
```diff
@@ -134,6 +84,16 @@ function create_uni_channel(
         open_id: DeviceId,
         label_id: LabelId,
     ) -> Result<AfcUniChannel, FfiError> {
+        use CommandContext::*;
+        match ctx {
+            Policy(inner_ctx) | Recall(inner_ctx) => {
+                if inner_ctx.author == open_id {
+                    return Err(FfiError::AuthorIsOpener);
+                }
```
jdygert-spok: create_uni_channel is called from the action

jdygert-spok: The author needs to create the encap to put it in the command to send it to the peer.

ssweetney-spok: Hmm, I guess that would make it too restrictive. It should be up to the policy to enforce usage?

ssweetney-spok: > The author needs to create the encap to put it in the command to send it to the peer.

Oh, that makes more sense

crates/aranya-fast-channels/src/testing/util.rs
```diff
@@ -133,43 +106,11 @@ pub enum ChanOp {
     SealOnly,
     /// Can only decrypt.
     OpenOnly,
-    /// Can encrypt and decrypt.
-    Any,
```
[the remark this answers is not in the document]
ssweetney-spok: It seems like `aranya_fast_channels::shm::shared::Op::Any` can be removed though right?

djohnson-spideroak (approved):
Please wait to merge until @jdygert-spok has a chance to re-review based on the feedback he left.

jdygert-spok:
I have some more cleanup in mind, but I don't think its necessary to do in the same PR: #428.

crates/aranya-afc-util/src/handler.rs
```diff
@@ -204,9 +45,11 @@ impl<S: KeyStore> Handler<S> {
         SK: for<'a> Transform<(&'a UniChannel<'a, E::CS>, UniAuthorSecret<E::CS>)>,
         OK: for<'a> Transform<(&'a UniChannel<'a, E::CS>, UniAuthorSecret<E::CS>)>,
     {
-        if (self.device_id != effect.seal_id && self.device_id != effect.open_id)
-            || self.device_id != effect.author_id
-        {
+        if self.device_id == effect.open_id || self.device_id != effect.seal_id {
```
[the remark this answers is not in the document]
ssweetney-spok: Yea, they don't seem necessary anymore. I think you did something similar when you simplified the policy for the bidi channel removal in the main Aranya repo

ssweetney-spok:
> I have some more cleanup in mind, but I don't think its necessary to do in the same PR: #428.

We should probably do something similar to https://github.com/aranya-project/aranya/issues/528, for the crates in this repo

crates/aranya-afc-util/src/handler.rs
```diff
@@ -204,10 +45,8 @@ impl<S: KeyStore> Handler<S> {
         SK: for<'a> Transform<(&'a UniChannel<'a, E::CS>, UniAuthorSecret<E::CS>)>,
         OK: for<'a> Transform<(&'a UniChannel<'a, E::CS>, UniAuthorSecret<E::CS>)>,
     {
-        if (self.device_id != effect.seal_id && self.device_id != effect.open_id)
-            || self.device_id != effect.author_id
-        {
-            return Err(Error::NotAuthor);
+        if self.device_id == effect.open_id {
+            return Err(Error::AuthorMustBeSealer);
```
jdygert-spok: We could remove these too since they end up checked when creating the key, but I suppose it's nice to have the more clear error. Looks like we can remove `NotAuthor` and `NotRecipient` though.

ssweetney-spok: Yea, we'll never actually get this error assuming we're using the default policy

jdygert-spok (approved):

Issue #436, opened 2025-10-14
policy-compiler: check for identifier reuse between different top level items


jdygert-spok:
> All top-level identifiers share a single namespace. So it is not valid,
for example, to give the same name to an action and a struct.

https://github.com/aranya-project/aranya-docs/blob/8d4b07d69421e7843c25376eda061df6b640372c/docs/policy-v1.md?plain=1#L402-L403

Additionally, we probably want to prevent having a variable or function parameter with the same name as a top level item.

This policy currently compiles:

```
command foo {
    seal { return todo() }
    open { return todo() }
    policy { finish {} }
    recall { finish {} }
}
action foo() {}
function foo(foo int) int {
    return foo
}

function bar(x int) int {
    let bar = x
    return bar
}

let baz = 42
action baz() {}
```

PR #439, opened 2025-10-16
policy-compiler: define math as builtin functions

jdygert-spok:


crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2977,6 +2965,22 @@ impl<'a> CompileState<'a> {
...
+            sig![function(x int, y int) -> int],
+        );
+        self.function_signatures
+            .insert(ident!("sub"), sig![function(x int, y int) -> optional int]);
+        self.function_signatures.insert(
+            ident!("saturating_sub"),
+            sig![function(x int, y int) -> int],
+        );
+        Ok(())
+    }
```
ssweetney-spok: This is clever.

https://github.com/aranya-project/aranya-core/issues/436 should probably be completed right after this

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2977,6 +2989,73 @@ impl<'a> CompileState<'a> {
...
+    fn define_builtins(&mut self) -> Result<(), CompileError> {
+        self.define_builtin(
+            sig! {
+                function add(x int, y int) optional int
+            },
+            |this| {
+                this.append_instruction(Instruction::Add);
+                Ok(())
+            },
+        )?;
```
jdygert-spok: defining a builtin ^

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2054,15 +2059,20 @@ impl<'a> CompileState<'a> {
...
+            let label = Label::new(
+                fc.identifier.name.clone(),
+                if is_finish {
+                    LabelType::Temporary
+                } else {
+                    LabelType::Function
+                },
+            );
+            self.append_instruction(Instruction::Call(Target::Unresolved(label)));
+        }
```
jdygert-spok: Using defined builtins or falling back to function call ^

chip-so (commented):
I think this approach is a little more complex than is necessary but I can see how it provides flexibility for future builtins.

ssweetney-spok (approved):

chip-so (approved):

PR #443, opened 2025-10-17
Exhaustive pattern matching

### LINKED ISSUE #481: policy-compiler: error when all cases in a match statement/expression aren't handled

ssweetney-spok:
Closes #481: handle all cases in match

### DIFF
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -90,9 +90,8 @@ struct CompileState<'a> {
     /// A map between function names and signatures, so that they can
     /// be easily looked up for verification when called.
     function_signatures: BTreeMap<&'a Identifier, FunctionSignature>,
-    /// The last locator seen, for imprecise source locating.
-    // TODO(chip): Push more precise source tracking further down into the AST.
-    last_locator: usize,
+    /// The last span seen, for imprecise source locating.
+    last_span: Span,
     /// The current statement context, implemented as a stack so that it can be
     /// hierarchical.
     statement_context: Vec<StatementContext>,
@@ -321,9 +320,9 @@ impl<'a> CompileState<'a> {
                 };
                 Ok(e.insert(signature))
             }
-            Entry::Occupied(_) => Err(CompileError::from_locator(
+            Entry::Occupied(_) => Err(CompileError::from_span(
                 CompileErrorType::AlreadyDefined(def.identifier.to_string()),
-                def.span.start(),
+                def.span,
                 self.m.codemap.as_ref(),
             )),
         }
@@ -348,9 +347,9 @@ impl<'a> CompileState<'a> {
                 };
                 Ok(e.insert(signature))
             }
-            Entry::Occupied(_) => Err(CompileError::from_locator(
+            Entry::Occupied(_) => Err(CompileError::from_span(
                 CompileErrorType::AlreadyDefined(def.identifier.to_string()),
-                def.span.start(),
+                def.span,
                 self.m.codemap.as_ref(),
             )),
         }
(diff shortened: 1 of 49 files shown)


crates/aranya-policy-compiler/src/compile/target.rs
```diff
@@ -79,6 +79,31 @@ impl CompileTarget {
             }),
         }
     }
+
+    pub(in crate::compile) fn cardinality(&self, kind: &TypeKind) -> Option<u64> {
+        match kind {
+            TypeKind::String | TypeKind::Bytes => None,
+            // With 2^(32 * 8) choices. It's unlikely for someone to want to match against IDs exhaustively
+            TypeKind::Id => None,
```
ssweetney-spok: Also realizing that you can't do this anyway because the patterns are limited to literals and AFACIT you can't create literal values with the ID type

crates/aranya-policy-compiler/src/compile/target.rs
```diff
@@ -79,6 +79,35 @@ impl CompileTarget {
             }),
         }
     }
+
+    pub(in crate::compile) fn cardinality(&self, kind: &TypeKind) -> Option<u64> {
+        match kind {
+            TypeKind::String | TypeKind::Bytes => None,
+            // With 2^(32 * 8) choices, it's unlikely for someone to want to match against IDs exhaustively.
+            TypeKind::Id => None,
+            // TODO: This should really be 1 above the max.
```
chip-so: I'd rate this as technically correct but a useless distinction. 😆 Ints are just as intractably enumerable as IDs or strings.

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -1477,6 +1479,61 @@ fn test_match_expression() {
...
+            }
+
+            action f(c struct ColorName) {
+                let x = match c {
+                    ColorName {  color: LightColor::Red, name: "red" } => 0
+                    ColorName {  color: LightColor::Yellow, name: "yellow" } => 1
+                    ColorName {  color: LightColor::Green, name: "green" } => 2
+                }
+            }"#,
+            CompileErrorType::MissingDefaultPattern,
```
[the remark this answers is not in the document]
ssweetney-spok: The [logic here](https://github.com/aranya-project/aranya-core/pull/443/files#diff-fcb8d2dfbbbedddc17f649464784e8c28bea74980202e0304c9c6d94df6fc0c4R95-R104), makes it so that if a struct has any field that can't be matched against exhaustively, then that makes it so the struct can't be matched against exhaustively either

crates/aranya-policy-compiler/src/compile/target.rs
jdygert-spok: This should be `None` then

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2558,6 +2558,19 @@ impl<'a> CompileState<'a> {
...
+                let maybe_cardinality = self.m.cardinality(&known_type.kind);
+
+                maybe_cardinality.is_none_or(|c| c > all_values.len() as u64)
+            } else {
+                true
+            };
+
+        if need_default {
+            return Err(self.err_loc(CompileErrorType::MissingDefaultPattern, span));
+        }
```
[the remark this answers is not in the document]
ssweetney-spok: Oh yea, I think this was also mentioned in the demo call. I think indicating some type of warning rather than an error would be better.

We could use [`Warning` variant](https://docs.rs/codespan-reporting/latest/codespan_reporting/diagnostic/enum.Severity.html) if/when we finish implementing the [compiler diagnostic updates](https://github.com/aranya-project/aranya-core/pull/377)

PR #449, opened 2025-10-20
Fix #447: Require policy block in command.

### LINKED ISSUE #447: Command compiles without policy block
This compiles when it shouldn't:
```
    command Cmd {
        fields {}
        seal { return todo() }
        open { return todo() }
}
```
Compilation should fail with something like "command requires policy block".

apetkov-so:
- Require a command to have a `policy` block
- Enforce order of command blocks

### DIFF
--- a/crates/aranya-policy-lang/tests/data/seal_open.snap
+++ b/crates/aranya-policy-lang/tests/data/seal_open.snap
@@ -1,10 +1,60 @@
 ---
 source: crates/aranya-policy-lang/tests/tests.rs
-expression: err
+expression: ast
 ---
-Syntax error: line 9 column 1:  --> 9:1
-  |
-9 | }
-  | ^---
-  |
-  = expected policy_block
+Policy {
+    version: V2,
+    ffi_imports: [],
+    facts: [],
+    actions: [],
+    effects: [],
+    structs: [],
+    enums: [],
+    commands: [
+        CommandDefinition {
+            persistence: Persistent,
+            attributes: [],
+            identifier: "Foo" @ 8..11,
+            fields: [],
+            seal: [
+                Return(
+                    ReturnStatement {
+                        expression: FunctionCall(
+                            FunctionCall {
+                                identifier: "bar" @ 40..43,
+                                arguments: [
+                                    Identifier(
+                                        "this" @ 44..48,
+                                    ) @ 44..48,
+                                ],
+                            },
+                        ) @ 40..49,
+                    },
+                ) @ 33..54,
+            ],
+            open: [
+                Return(
+                    ReturnStatement {
+                        expression: FunctionCall(
+                            FunctionCall {
+                                identifier: "baz" @ 82..85,
+                                arguments: [
+                                    Identifier(
+                                        "envelope" @ 86..94,
+                                    ) @ 86..94,
+                                ],
+                            },
+                        ) @ 82..95,
+                    },
+                ) @ 75..100,
+            ],
+            policy: [],
+            recall: [],
+            span: 0..131,
+        },
+    ],
+    functions: [],
+    finish_functions: [],
+    global_lets: [],
+    text: "command Foo {\n    seal {\n        return bar(this)\n    }\n    open {\n        return baz(envelope)\n    }\n    policy {\n        \n    }\n}\n",
+}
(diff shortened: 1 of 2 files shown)


apetkov-so:
There's a follow up ticket for duplicate blocks: https://github.com/aranya-project/aranya-core/issues/450

ssweetney-spok (commented):
I didn't realize omitting the policy block was possible... good catch.

How do we want to handle this? As is, you could still write

```
policy {
    let num = 1
}
```

which seems as useful as an empty/missing policy block.

Edit: We do similar checks for the seal/open blocks so this should be OK

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -1902,6 +1959,8 @@ fn test_type_errors() {
         Case {
             t: r#"
                 command Foo {
+                    seal { return todo() }
```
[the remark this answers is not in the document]
jdygert-spok: The policy block is compiled first, and we don't check for the seal block until we try to compile it.

ssweetney-spok (dismissed):

apetkov-so:
> ... which seems as useful as an empty/missing policy block.

We could add a command analyzer to the tracer, that would ensure there's a finish block, and a least one publish.

chip-so:
> We could add a command analyzer to the tracer, that would ensure there's a finish block, and a least one publish.

We already have the [FinishAnalyzer](https://github.com/aranya-project/aranya-core/blob/main/crates/aranya-policy-compiler/src/tracer/analyzers/finish_analyzer.rs).

chip-so (dismissed):

chip-so:
> How do we feel about changing the pest rule to require a certain order?

I guess that's fine but it also seems unimportant.

jdygert-spok:
> > How do we feel about changing the pest rule to require a certain order?
> 
> I guess that's fine but it also seems unimportant.

I ask because it solves this issue and #450 in a simpler way.

crates/aranya-policy-lang/src/lang/parse/tests.rs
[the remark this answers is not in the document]
apetkov-so: Ugh, merge hell. Let me go figure out what got moved and where.

crates/aranya-policy-lang/tests/data/command_attributes.snap
ssweetney-spok: This test and the other ones that became errors should probably be updated to have the missing blocks then a create separate test to show this error (due to missing the blocks).

The attribute parsing this is supposed to test isn't present in "tictactoe.md" so it'd be good to test it

apetkov-so: But the new tests would then be error tests, no?

ssweetney-spok: Yea, I was thinking that there would be just one new test (not multiple) to show the "expected fields_block or seal_block" error.

This file shouldn't be an error though since we still want to have a test to demonstrate that parsing attributes works as expected.

chip-so (commented):
I guess this is fine, but I don't love that this turns a missing block into a syntax error. Pest Syntax errors are the worst kind of errors we produce, and are unfortunately opaque about what you need to do to fix them.

chip-so (dismissed):

crates/aranya-policy-lang/tests/data/policy_test.snap
ssweetney-spok: This is outputting an error again. It looks like there's a missing block in `crates/aranya-policy-lang/tests/data/policy_test.policy` somewhere

ssweetney-spok (commented):
LGTM. Just some small comments left

crates/aranya-policy-lang/tests/data/policy_test.snap
apetkov-so: What do you mean? The tests pass.

crates/aranya-policy-lang/tests/data/seal_open.snap
jdygert-spok: Probably want to add policy block to make this one pass again

jdygert-spok: Might be a redundant test at this point though.

chip-so (approved):

PR #451, opened 2025-10-20
aqc: remove

jdygert-spok:


crates/aranya-crypto/src/test_util/engine.rs
```diff
@@ -115,32 +114,6 @@ macro_rules! for_each_engine_test {
...
-            test_aqc_derive_uni_recv_psk_same_device_id,
-            test_aqc_derive_uni_send_psk_psk_too_short,
-            test_aqc_derive_uni_recv_psk_psk_too_short,
-            test_aqc_derive_uni_send_psk_psk_too_long,
-            test_aqc_derive_uni_recv_psk_psk_too_long,
-            test_aqc_wrap_uni_author_secret,
-
             // TLS
 
             test_tls_psk_different_suites,
```
gknopf-aranya: Do we still need these TLS tests? The only thing using TLS now would be the QUIC syncer.

jdygert-spok: Hm maybe not since they are specifically for PSKs.

gknopf-aranya: Doesn't the QUIC syncer still use PSKs?

gknopf-aranya:
Adding @ssweetney-spok as a reviewer since he's worked with ffis recently

ssweetney-spok (approved):

PR #452, opened 2025-10-20
ci: catch aranya-core changes that break the aranya-embedded repo

gknopf-aranya:
Create a CI workflow that verifies the embedded repo builds with latest aranya-core code.

### DIFF
--- a/.github/workflows/embedded.yml
+++ b/.github/workflows/embedded.yml
@@ -40,6 +40,7 @@ jobs:
             repository: aranya-project/aranya-embedded
             path: aranya-embedded
 
+      # Refer to https://github.com/aranya-project/aranya-embedded/blob/main/crates/chat-app/TUTORIAL.md for installation and build instructions.
       - name: Install esp toolchain
         run: |
             sudo apt-get update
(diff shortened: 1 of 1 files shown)


gknopf-aranya:
Should we have the CI job pass while printing the build errors if `aranya-embedded` breaks?
Or should we have the job fail when `aranya-embedded` breaks with the latest `aranya-core` changes but not make the CI job required in order to merge `aranya-core` PRs?

A failing CI job seems loud enough to get developer's attention. I'm worried that if we have the job silently pass, no one will notice `aranya-embedded` has broken until much later.
The downside of this is that CI on developer PRs and the `main` branch could have a red/failing status for a while until we have bandwidth to fix `aranya-embedded`. Or a developer could be mislead into thinking their PR is not ready to be merged yet.

.github/workflows/embedded.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Install esp toolchain
+        run: |
+            sudo apt-get update
+            sudo apt-get install libudev-dev --fix-missing
+            cargo install espup --locked
+            espup install
+            . /home/runner/export-esp.sh
+            cargo install espflash@^3 --locked
+
+      - name: Build code
```
gknopf-aranya:
```suggestion
      - name: Build chat-app crate in aranya-embedded repo
```

gknopf-aranya:
> Looks good to me but I don't know anything about git workflows

If you navigate to the "Checks" tab, it shows a list of workflows that ran on the PR. "Verify aranya-embedded repo builds" -> "build-embedded" is the workflow that was added which shows all the output from toolchain installation and the passing build:
https://github.com/aranya-project/aranya-core/actions/runs/18786344812/job/53605475930?pr=452

PR #466, opened 2025-10-30
Add channel direction to `RemoveIfParams`

### LINKED ISSUE #464: afc: add channel direction info to remove_if params
We need to know the channel direction when running `remove_if()` so we can lookup which channels are still valid according to the policy.

The policy query is defined like this:
`ephemeral action query_afc_channel_is_valid(sender_id id, receiver_id id, label_id id)`

Adding the channel direction to the `remove_if()` params will solve this.

cc: @djohnson-spideroak, @jdygert-spok, @ssweetney-spok

ssweetney-spok:
Resolves #464

### DIFF
--- a/crates/aranya-fast-channels/src/memory.rs
+++ b/crates/aranya-fast-channels/src/memory.rs
@@ -145,7 +145,7 @@ where
                     id,
                     *label_id,
                     *peer_id,
-                    keys.directed(),
+                    keys.direction(),
                 ))
             },
         );
--- a/crates/aranya-fast-channels/src/state.rs
+++ b/crates/aranya-fast-channels/src/state.rs
@@ -207,7 +207,7 @@ impl<S, O> Directed<S, O> {
     }
 
     /// Returns the corresponding [ChannelDirection].
-    pub fn directed(&self) -> ChannelDirection {
+    pub fn direction(&self) -> ChannelDirection {
         match self {
             Self::SealOnly { .. } => ChannelDirection::Seal,
             Self::OpenOnly { .. } => ChannelDirection::Open,


crates/aranya-fast-channels/src/memory.rs
```diff
@@ -16,7 +16,7 @@ use buggy::BugExt as _;
 use derive_where::derive_where;
 
 use crate::{
-    LocalChannelId, RemoveIfParams,
+    ChanDirection, LocalChannelId, RemoveIfParams,
```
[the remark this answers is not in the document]
chip-so: I dunno about convention but I'd prefer to spell things out.

crates/aranya-fast-channels/src/shm/shared.rs
```diff
@@ -276,18 +276,6 @@ impl From<Offset> for usize {
     }
 }
 
-/// Describes one of [`ShmChan`]'s variants.
-// NB: see `ShmChan::matches` to understand the
-// discriminants.
-#[repr(u32)]
-#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, Ord, PartialOrd)]
-enum ChanDirection {
```
gknopf-aranya: Why is this being removed? I wouldn't expect this to be removed on a PR that is adding this as a param to remove_if. Unless it was moved somewhere else.

ssweetney-spok: It was moved to a different module

crates/aranya-fast-channels/src/state.rs
```diff
@@ -273,6 +281,28 @@ impl<S, O> Debug for Directed<S, O> {
     }
 }
 
+/// Describes the flow of data for an AFC channel.
+// NB: see `crate::shm::shared::ShmChan::matches` to understand how
+// the discriminants relate to `crate::shm::shared::Op`.
+#[repr(u32)]
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, Ord, PartialOrd)]
+pub enum ChanDirection {
```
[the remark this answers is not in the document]
gknopf-aranya: Good point. Then we don't need `pub(crate)` methods on the public type.

gknopf-aranya:
Should we rename `ChanOp` to `ChannelOp`? That could go on a different PR.

gknopf-aranya (approved):

crates/aranya-fast-channels/src/state.rs
```diff
@@ -197,6 +205,14 @@ impl<S, O> Directed<S, O> {
             Self::SealOnly { .. } => None,
         }
     }
+
+    /// Returns the corresponding [ChannelDirection].
+    pub fn directed(&self) -> ChannelDirection {
```
jdygert-spok: `direction`

gknopf-aranya (approved):

gknopf-aranya (approved):

PR #477, opened 2025-11-04
Fix #465: Detect unreachable code

### LINKED ISSUE #465: policy: unreachable code not detected
Unreachable code is not detected when returning early:
```
function afc_uni_channel_is_valid(sender_id id, receiver_id id, label_id id) bool {
    return true

    // The label must exist.
    let label = query Label[label_id: label_id]
    if label is None {
        return false
    }

    // Devices cannot create channels with themselves.
    //
    // This should have been caught by the AFC FFI
    if sender_id == receiver_id {
        return false
    }
    ...
    return true
}
```

I stumbled on this when debugging why a function wasn't returning the expected value.
It was helpful to move the return statement around to see where the incorrect value was being returned.
Not sure if this is something we want to fix, but it seems to be a feature of many other languages.

cc: @chip-so, @jdygert-spok, @apetkov-so

apetkov-so:
Fix #465: Detect unreachable code in functions, e.g. statements after `return`.

### DIFF
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -1028,7 +1028,6 @@ impl<'a> CompileState<'a> {
             Label::new(function_node.identifier.name.clone(), LabelType::Function),
         )?;
         self.exit_statement_context();
-        self.append_instruction(Instruction::Meta(Meta::FunctionEnd));
         Ok(())
     }
 
@@ -1107,8 +1106,8 @@ impl<'a> CompileState<'a> {
         )?;
         // Actions cannot have return statements, so we add a return instruction manually.
         self.append_instruction(Instruction::Return);
-        self.exit_statement_context();
         self.append_instruction(Instruction::Meta(Meta::FunctionEnd));
+        self.exit_statement_context();
         Ok(())
     }
 
@@ -1184,6 +1183,7 @@ impl<'a> CompileState<'a> {
             &command.policy,
             Label::new(command.identifier.name.clone(), LabelType::CommandPolicy),
         )?;
+        // Policy blocks cannot have return statements, so we mark the end.
         self.append_instruction(Instruction::Meta(Meta::FunctionEnd));
         // Policy blocks should exit via a finish block, so panic if it doesn't.
         self.append_instruction(Instruction::Exit(ExitReason::Panic));
@@ -1340,6 +1340,7 @@ impl<'a> CompileState<'a> {
             {
                 return Err(self.err_loc(CompileErrorType::NoReturn, span));
             }
+            self.append_instruction(Instruction::Meta(Meta::FunctionEnd));
             // If execution does not hit a return statement, it will panic here.
             self.append_instruction(Instruction::Exit(ExitReason::Panic));
         }
--- a/crates/aranya-policy-compiler/src/tracer/analyzers/value_analyzer.rs
+++ b/crates/aranya-policy-compiler/src/tracer/analyzers/value_analyzer.rs
@@ -1,7 +1,7 @@
 use std::collections::BTreeSet;
 
 use aranya_policy_ast::Identifier;
-use aranya_policy_module::{Instruction, Meta, ModuleV0};
+use aranya_policy_module::{Instruction, ModuleV0};
 
 use super::{Analyzer, AnalyzerStatus};
 use crate::tracer::TraceError;
@@ -46,14 +46,16 @@ impl Analyzer for ValueAnalyzer {
                 self.value_sets.push(BTreeSet::new());
             }
             Instruction::Return => {
-                self.value_sets.pop();
+                if self.value_sets.len() > 1 {
+                    self.value_sets.pop();
+                }
             }
-            Instruction::Meta(Meta::Let(s)) => {
+            Instruction::Def(s) => {
                 if !self.insert(s.clone()) {
                     return Ok(AnalyzerStatus::Failed(format!("Value `{s}` is set twice")));
                 }
             }
-            Instruction::Meta(Meta::Get(s)) => {
+            Instruction::Get(s) => {
                 if !self.contains(s) {
                     return Ok(AnalyzerStatus::Failed(format!("Value `{s}` is not set")));
                 }


crates/aranya-policy-compiler/src/tracer/analyzers/function_analyzer.rs
```diff
@@ -18,20 +20,42 @@ impl FunctionAnalyzer {
...
+        pc: usize,
         i: &Instruction,
         _m: &ModuleV0,
     ) -> Result<AnalyzerStatus, TraceError> {
         match i {
-            Instruction::Return => self.have_return = true,
+            Instruction::Return => {
+                self.have_return = true;
+            }
             Instruction::Exit(_) => {
```
jdygert-spok: Just noticed we're using the panic we add to the end of functions to detect when the function ends. This interacts poorly with other sources of panics.

crates/aranya-policy-compiler/src/tracer/analyzers/function_analyzer.rs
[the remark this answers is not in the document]
chip-so: Yeah you definitely need to track your entry/exit to have the correct scope. See how the [`ValueAnalyzer` works](https://github.com/aranya-project/aranya-core/blob/main/crates/aranya-policy-compiler/src/tracer/analyzers/value_analyzer.rs).

crates/aranya-policy-compiler/src/tracer/analyzers/action_analyzer.rs
```diff
@@ -1,36 +1,88 @@
...
+                                    responsible_instruction: pc,
+                                    message: "no publish".to_string(),
+                                });
+                            }
+                        }
+                        Instruction::Publish => {
+                            if call_depth == 0 {
+                                have_publish = true;
+                            }
+                        }
```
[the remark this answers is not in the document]
ssweetney-spok: Oh, nevermind. I thought we allowed actions to call other actions

crates/aranya-policy-compiler/src/tracer/analyzers/action_analyzer.rs
```diff
@@ -1,36 +1,88 @@
...
     ) -> Result<AnalyzerStatus, TraceError> {
-        match i {
-            Instruction::Publish => self.have_publish = true,
-            Instruction::Return => {
-                if !self.have_publish {
-                    return Ok(AnalyzerStatus::Failed("no publish".to_string()));
-                }
-            }
-            _ => (),
-        }
```
chip-so: Why has all of the analysis moved to `post_analyze()`? I guess you want to filter out the unsuccessful branches for some reason. I'm wondering if this means the architecture of the tracer isn't fit for purpose.

chip-so (dismissed):

crates/aranya-policy-compiler/src/tracer/analyzers/action_analyzer.rs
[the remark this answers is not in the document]
apetkov-so: Added https://github.com/aranya-project/aranya-core/issues/505 to detect indirect `publish`.

crates/aranya-policy-compiler/src/tracer/analyzers/action_analyzer.rs
apetkov-so: We still break on `Exit`, we don't break on `Return`. Since a function can have multiple returns, we need something else to mark the end - that's the `Meta::FunctionEnd` instruction.

ssweetney-spok (dismissed):

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -2727,14 +2727,262 @@ fn test_validate_return() {
...
+            return baz()
+        }
+        function should_be_valid() int {
+            let y = foobar()
+            return y
+        }
+        "#,
+    ];
+
+    let invalid = [
```
gknopf-aranya: How about testing this case?
```
function unreachable(x int) int {
    match x {
        0 => return 0,
        _ => return 1,
    }
    return 2
}
```

apetkov-so: Those returns aren't supported yet. I'll add that case after #501 lands.
Edit: Actually we can do `=> { return 0 }`.

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -2727,14 +2727,262 @@ fn test_validate_return() {
...
+        r#"function unreachable3(x int) int {
+            if x > 0 {
+                return 1
+            } else {
+                return 0
+            }
+            let y = 2
+        }"#,
+        r#"
+        function unreachable() int {
```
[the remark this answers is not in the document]
apetkov-so: done

ssweetney-spok (dismissed):

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -2727,14 +2727,269 @@ fn test_validate_return() {
...
+        "#,
+        // Conditional paths all reaching finish
+        r#"
+            command Conditional {
+                fields { a int }
+                seal { return todo() }
+                open { return todo() }
+                policy {
+                    if true {
+                        finish {}
```
gknopf-aranya: Is it possible to have a `finish{}` block inside a match? Or is that only in if-else statements?

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -2727,14 +2727,269 @@ fn test_validate_return() {
...
+                }
+                recall {
+                    finish {}
+                }
+            }
+        "#,
+    ];
+
+    // Invalid cases: policy or recall blocks with statements but no finish
+    let invalid = [
```
gknopf-aranya: This would be a case of an unreachable finish:
```
policy {
    check false
    finish{}
}
```

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -2727,14 +2727,269 @@ fn test_validate_return() {
...
+            }
+        "#,
+        // Not all paths in policy reach finish
+        r#"
+            command NoFinishConditional {
+                fields { a int }
+                seal { return todo() }
+                open { return todo() }
+                policy {
+                    if true {
```
gknopf-aranya: How about testing for a missing finish block in the else branch:
```
if false {
    finish {}
}
else {
    let z = 3 // missing finish here
}
```

crates/aranya-policy-compiler/src/tracer/analyzers/action_analyzer.rs
chip-so: It does trace all possible branches. It stops tracing each branch when it would normally stop executing.

crates/aranya-policy-compiler/src/tests.rs
gknopf-aranya: @apetkov-so, @chip-so do we have a way to detect unused functions? claude recently found an unused function in the default policy of the daemon where the policy compiler didn't report any warning/error:
```
// Reports whether a device can remove itself from the team.
// Owners can only remove themselves if there are other owners
// remaining. Other roles can always remove themselves.
function can_remove_self(device_id id) bool {
    let maybe_role = try_get_assigned_role(device_id)
    if maybe_role is None {
        // Device has no role, can be removed
        return true
    }
    let role = unwrap maybe_role
    if is_owner(role) {
        // Owner can only remove self if there are other owners
        return at_least 2 RoleAssignmentIndex[role_id: role.role_id, device_id: ?]=>{}
    }
    // All other roles can remove themselves
    return true
}
```

ssweetney-spok: > claude recently found an unused function in the default policy of the daemon where the policy compiler didn't report any warning/error

Oh, that's neat.

That should probably be a lint/warning that's handled in the compiler rather than a tracer.

apetkov-so: Not yet. #523.

apetkov-so: > > claude recently found an unused function in the default policy of the daemon where the policy compiler didn't report any warning/error

Maybe we should integrate Claude into our policy verification? 😂 It can detect security issues, unreachable and unused code, probably logic errors...
I'm joking, but given how much these tools can do, it's probably worth adding a Claude action to the CI pipeline.

gknopf-aranya (approved):

gknopf-aranya:
Looks good from what I can tell. Probably want another approval from @chip-so or @jdygert-spok.

ssweetney-spok (approved):

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1027,6 +1027,8 @@ impl<'a> CompileState<'a> {
         if !self.instruction_range_contains(from..self.wp, |i| matches!(i, Instruction::Return)) {
             return Err(self.err_loc(CompileErrorType::NoReturn, function_node.span));
         }
+        // Mark end function body; used for additional validation, e.g. unreachable-code detection.
+        self.append_instruction(Instruction::Meta(Meta::FunctionEnd));
```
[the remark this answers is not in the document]
apetkov-so: We now have them before all exit_function calls.

crates/aranya-policy-compiler/src/tracer.rs
```diff
@@ -207,21 +216,32 @@ impl TraceAnalyzer<'_> {
...
                         successful_branch_paths.push(self.branches.clone());
+                        successful_instruction_paths.push(self.instruction_path.clone());
                         return Ok(TraceIntermediate {
                             failures,
                             successful_branch_paths,
+                            successful_instruction_paths,
                         });
                     }
-                    pc = self.call_stack.pop().expect("impossible empty stack");
+                    // Continue to allow unreachable code detection in nested functions
```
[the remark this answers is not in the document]
chip-so: That's correct. This way you shouldn't have to modify the behavior of the existing tracers. I expect this normal termination behavior will continue to be the most common case.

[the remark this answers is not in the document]
chip-so: > But the action and finish analyzers would still need to visit all successful paths (in post_analyze) to ensure they publish/finish.

I don't understand why. The basic operation of these analyzers is quite simple. As the instructions are traced, it sets a flag when the path reaches the thing we're looking for (a finish block or a publish instruction), and when it hits a terminal instruction (exit or return), it checks to see if that flag was set. If it wasn't, that's a failed run.

We don't need to check afterwards. By nature of how the tracer follows every code path, the absence of any failure proves that all code paths met the conditions.

PR #480, opened 2025-11-10
afc: add `SealCtx` to fix `Seq` tracking

### LINKED ISSUE #422: afc: shm doesn't properly record sequence number

jdygert-spok:
Fixes #422 with minimal changes by changing seal to take a mutable context.

### DIFF
--- a/crates/aranya-fast-channels/benches/lib.rs
+++ b/crates/aranya-fast-channels/benches/lib.rs
@@ -174,8 +174,8 @@ macro_rules! bench_impl {
 					})
 				});
 
-				// TODO: seal_miss no longer misses since each channel gets its
-                // own cache. There are still misses when the list generation
+                // TODO(#482): seal_miss no longer misses since each channel gets
+                // its own cache. There are still misses when the list generation
                 // changes, so we should measure those.
 				let mut iter = chans.iter_mut();
 				g.bench_function(BenchmarkId::new("seal_miss", *size), |b| {
--- a/crates/aranya-fast-channels/src/client.rs
+++ b/crates/aranya-fast-channels/src/client.rs
@@ -55,8 +55,10 @@ impl<S: AfcState> Client<S> {
 
     /// Set up the seal context for the given channel.
     ///
-    /// This must only be called once for any given ID. Failure to
-    /// do so will result in repeated sequence numbers.
+    /// This must only be called once for any given channel ID. Failure to
+    /// do so will result in repeated usage of sequence numbers and thus a
+    /// repeated use of a nonce. This will open the channel up to a variety
+    /// of attacks and remove any security guarantees.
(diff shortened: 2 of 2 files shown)


crates/aranya-fast-channels/benches/lib.rs
```diff
@@ -169,12 +176,12 @@ macro_rules! bench_impl {
...
 				// never cached.
-				let mut iter = chans.iter().cycle().copied();
+				let mut iter = chans.iter_mut();
 				g.bench_function(BenchmarkId::new("seal_miss", *size), |b| {
-					let (seal_channel_id, _) = iter.next().expect("should repeat");
+                    let (seal_ctx, _) = iter.next().expect("not enough channels");
 					b.iter(|| {
 						black_box(client.seal(
-							black_box(seal_channel_id),
+							black_box(seal_ctx),
```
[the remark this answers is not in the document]
jdygert-spok: Yeah, we also should distinguish the benchmark cases of whether the channel moves to a different index. I don't think it's really worth fixing right now. Unsure whether to keep the benchmark the same to show that this specific case has improved since there's essentially a cache for every seal now.

ssweetney-spok (dismissed):

crates/aranya-fast-channels/src/client.rs
```diff
@@ -148,10 +156,10 @@ impl<S: AfcState> Client<S> {
             /* ad: */ &AuthData,
         ) -> Result<Seq, Error>,
     {
-        debug!("finding seal info: id={id}");
+        // debug!("finding seal info: id={id}");
 
-        let seq = self.state.seal(id, |aead, label_id| {
-            debug!("encrypting id={id}");
+        let seq = self.state.seal(ctx, |aead, label_id| {
+            // debug!("encrypting id={id}");
```
gknopf-aranya: Remove this?

crates/aranya-fast-channels/src/client.rs
```diff
@@ -53,14 +53,22 @@ impl<S: AfcState> Client<S> {
         v
     }
 
+    /// Set up the seal context for the given channel.
+    ///
+    /// This must only be called once for any given ID. Failure to
+    /// do so will result in repeated sequence numbers.
```
gknopf-aranya:
```suggestion
    /// This must only be called once for any given channel ID. Failure to
    /// do so will result in repeated sequence numbers (a.k.a. a repeated nonce).
```
What is the consequence of a repeated nonce in this case? Does it leak the key or plaintext? Does it make us vulnerable to a chosen plaintext/ciphertext attack? May be worth documenting the risk here.

gknopf-aranya (dismissed):

gknopf-aranya (approved):

PR #485, opened 2025-11-18
afc: Pass `OpenCtx` to client methods

### LINKED ISSUE #399: afc: concurrent shm readers

ssweetney-spok:
Closes #399

[Usage in Aranya](https://github.com/aranya-project/aranya/pull/611)

### DIFF
--- a/crates/aranya-policy-ast/src/ast.rs
+++ b/crates/aranya-policy-ast/src/ast.rs
@@ -7,21 +7,21 @@ use crate::{Identifier, Span, Spanned, Text, span::spanned};
 
 /// An identifier.
 #[derive(
-    Debug,
-    Clone,
-    Eq,
-    PartialEq,
-    Serialize,
-    Deserialize,
-    rkyv::Archive,
-    rkyv::Deserialize,
-    rkyv::Serialize,
+    Clone, Eq, PartialEq, Serialize, Deserialize, rkyv::Archive, rkyv::Deserialize, rkyv::Serialize,
 )]
 pub struct Ident {
     /// The identifier name
     pub name: Identifier,
     /// The source location of this identifier
     pub span: Span,
+}
+
+impl fmt::Debug for Ident {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.name.fmt(f)?;
+        write!(f, " @ {:?}", self.span)?;
+        Ok(())
+    }
 }
 
 impl Ident {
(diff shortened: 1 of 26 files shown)


crates/aranya-fast-channels/src/shm/read.rs
```diff
@@ -44,6 +44,11 @@ struct Cache<K> {
 }
 
 /// The reader's view of the shared memory state.
+///
+/// # Notes
+///
+/// Dropping a `ReadState` instance and its clones will not unmap
```
gknopf-aranya: I still can't run the aranya C example locally due to this without changing shm paths. Because each time the shm layout changes, it is incompatible with the previous one.

ssweetney-spok:
I realized that `ReadState` is still `!Sync` so we'd have to remove `ReadState::_no_sync` to be able to use it in an `Arc`. Coming back to [this thread](https://git.spideroak-inc.com/spideroak-inc/aps/issues/95), do we still need additional synchronization on the `CSPRNG`?

cc: @jdygert-spok

jdygert-spok:
I was confused what rng that thread was referring to, found it in an old version. Looks like it was needed to fill in a random nonce. I think that is no longer done due to use of sequence numbers. So I don't think there is anything stopping us from this change.

https://git.spideroak-inc.com/spideroak-inc/aps/blob/cdc701dace1cc2fddb75f8e4ec295e6937b7d9e3/aps/src/client.rs#L139

crates/aranya-fast-channels/src/shm/read.rs
```diff
@@ -51,14 +50,6 @@ where
...
     // `pub(super)` for testing.
     pub(super) inner: State<CS>,
-
-    // APS is typically used to seal/open many messages with the
-    // same peer, so cache the most recent successful invocations
-    // of seal/open.
-    last_open: StdMutex<Option<Cache<OpenKey<CS>>>>,
-
-    /// Make `State` `!Sync` pending issues/95.
-    _no_sync: PhantomData<Cell<()>>,
```
jdygert-spok: Check if this is `Send + Sync` now. I think we need to add an impl to `Mapping`

PR #486, opened 2025-11-18
policy-compiler: remove `Typeish` and `NullableVType`

jdygert-spok:
Rather than wrapping `VType` in two layers to add `null` and `never`, we can add `TypeKind::Never` and represent `None` as `Optional(Never)`. This simplifies things a lot, especially for the upcoming `Result` constructors.

I also changed where we allow `never` to typecheck. We used to allow it almost everywhere, which is valid but not entirely necessary. It will still unify with other types, so e.g. `if b { 42 } else { todo() }` has type `int`. But cases like `todo().field` will now error out rather than propagate the `never` type. This is a tad simpler to implement and also seems common, looking at Rust and Typescript.

### DIFF
--- a/crates/aranya-policy-compiler/src/tests.rs
+++ b/crates/aranya-policy-compiler/src/tests.rs
@@ -2261,7 +2261,7 @@ fn test_struct_composition() {
     let valid_cases = [Case {
         t: r#"
                 struct Foo { x int, y bool }
-                struct Bar { x int, y bool, z string}
+                struct Bar { x int, y bool, z string }
                 function baz(foo struct Foo) struct Bar {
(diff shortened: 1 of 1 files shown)


chip-so:
> Doesn't seem like that much of an issue.

Agreed.

> I suppose we could consider properly inferring the type of `None` expressions in the future but I don't think we'll be able to do that if we want to keep a single-pass compiler

I'm not sure we have a strictly single-pass compiler now. I'm not opposed to making the compiler more complex to support better ergonomics and safety.

chip-so (approved):

ssweetney-spok (approved):

PR #496, opened 2025-11-21
policy-compiler: add `compile_interface`

jdygert-spok:
`Compiler::compile_interface` lets you compile just the "public interface" of the policy, without actually compiling functions/actions/blocks etc. This is used for the policy interface generator.

### DIFF
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -967,7 +967,11 @@ impl<'a> CompileState<'a> {
                     f.module.name.clone(),
                     f.identifier.name.clone(),
                 )));
+
                 if self.stub_ffi {
+                    for arg_e in &f.arguments {
+                        let _arg_ty = self.compile_expression(arg_e)?;
+                    }
                     self.append_instruction(Instruction::Exit(ExitReason::Panic));
                     VType {
                         kind: TypeKind::Never,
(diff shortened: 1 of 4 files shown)


crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2642,22 +2646,11 @@ impl<'a> CompileState<'a> {
...
     }
 
-    /// Compile a policy into instructions inside the given Machine.
-    pub fn compile(&mut self) -> Result<(), CompileError> {
-        // Panic when running a module without setup.
-        self.append_instruction(Instruction::Exit(ExitReason::Panic));
-
-        self.define_builtins()?;
-
+    fn define_interfaces(&mut self) -> Result<(), CompileError> {
```
jdygert-spok: Facts are currently here for convenience. Could keep the fact struct def here and move the `define_fact`. Or move them both down if we don't want fact structs to be usable in the public interface. I think that was mentioned at some point.

jdygert-spok: The question is whether they _should_ be allowed as part of the public interface. If we want actions only used for code reuse reasons that might be able to use private types, that's a separate issue that would probably be solved with `pub/priv`.

ssweetney-spok: That makes sense. I was under the impression that we'd keep all actions as public but adding visibility modifiers sounds like it could work

ssweetney-spok (approved):

PR #503, opened 2025-12-03
capi-codegen: expand opaque during codegen

jdygert-spok:


jdygert-spok:
> Do we still need the `self.opaque.to_tokens(tokens);` line now that expansion happens during code generation?

Good call, we can remove that and the `generated` arg from opaque macro.

Issue #17, opened 2025-01-07
policy-lang: Properly parse negative integers


chip-so:
Currently we're parsing negative integers as two pieces - a negation token (`Rule::neg`) and the integer (`Rule::int_literal`). Then [when parsing `Rule::neg`](https://github.com/aranya-project/aranya-core/blob/main/crates/aranya-policy-lang/src/lang/parse.rs#L528-L537), we check for an int literal expression and construct a negative `Expression::Int` instead. This works for every integer except one, `i64::MIN` (-9223372036854775808). Because the positive version of that is unrepresentable by an `i64`, it can't be parsed.

Change the parse grammar and code to parse negative integers directly. That is, the string parsed by `parse::<i64>()` should have the `-` already there so we can parse `i64::MIN`.

nikki-aranya:
Following up: This is a lot more difficult than we thought due to the current parser setup. It's better to leave this as an edge case rather than come up with a hacky solution that technically works. Can revisit.

PR #508, opened 2025-12-05
policy-lang: add custom error for `+` and `-` ops

jdygert-spok:
Add custom errors to prevent confusion when using the old math operators. This approach seemed to work fine. I didn't add unary negate since I think it would conflict with negative literals and bring back #17.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -130,17 +130,13 @@ proc-macro2 = "1.0.80"
 syn = "2.0.59"
 quote = "1.0"
 
-[profile.dev]
-panic = "abort"
-opt-level = 1
-
 [profile.dev.package]
+aranya-quic-syncer.opt-level = 1
+aranya-runtime.opt-level = 1
+aranya-model.opt-level = 1
+
 insta.opt-level = 3
 similar.opt-level = 3
-
-[profile.dev-std]
-inherits = "dev"
-panic = "unwind"
 
 [profile.release]
 lto = "thin"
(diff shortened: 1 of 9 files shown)


crates/aranya-policy-lang/tests/data/match_expression/empty_match.snap
```diff
@@ -7,4 +7,4 @@ Syntax error: line 3 column 1:  --> 3:1
 3 | }
   | ^---
   |
-  = expected greater_than, less_than, greater_than_or_equal, less_than_or_equal, equal, not_equal, and, or, dot, substruct, cast, or is
+  = expected add, subtract, greater_than, less_than, greater_than_or_equal, less_than_or_equal, equal, not_equal, and, or, dot, substruct, cast, or is
```
jdygert-spok: This isn't ideal :/ but maybe OK.

ssweetney-spok (approved):

crates/aranya-policy-lang/src/lang/parse.rs
```diff
@@ -678,6 +678,20 @@ impl ChunkParser<'_> {
...
+                        return Err(ParseError::new(
+                            ParseErrorKind::InvalidOperator,
+                            String::from("found `+`, addition now uses functions `add` or `saturating_add`"),
+                            Some(op.as_span())
+                        ));
+                    }
+                    Rule::subtract => {
+                        return Err(ParseError::new(
+                            ParseErrorKind::InvalidOperator,
+                            String::from("found `-`, subtraction now uses functions `sub` or `saturating_sub`"),
```
gknopf-aranya: Would be useful to recommend the correct code so the policy writer doesn't need to look up these functions somewhere.

crates/aranya-policy-lang/tests/data/old_subtraction.policy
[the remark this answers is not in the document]
jdygert-spok: Wouldn't hurt, but subtraction is the one that would be more likely to fail given negative int literals.

PR #511, opened 2025-12-08
473 sync failures

benz-spideroak:
For this I added some code to make it easier to find minimal failing tests. I then found and fixed a failing test. I also improved sped up add_command.

### DIFF
--- a/crates/aranya-quic-syncer/src/lib.rs
+++ b/crates/aranya-quic-syncer/src/lib.rs
@@ -12,9 +12,8 @@ use std::{
 
 use aranya_crypto::{Csprng as _, Rng};
 use aranya_runtime::{
-    COMMAND_RESPONSE_MAX, ClientError, ClientState, Command as _, MAX_SYNC_MESSAGE_SIZE, PeerCache,
-    StorageError, SubscribeResult, SyncError, SyncRequestMessage, SyncRequester, SyncResponder,
-    SyncType,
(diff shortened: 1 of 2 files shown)


crates/aranya-runtime/src/sync/responder.rs
```diff
@@ -286,14 +295,11 @@ impl<A: Serialize + Clone> SyncResponder<A> {
     ) -> Result<Vec<Location, SEGMENT_BUFFER_MAX>, SyncError> {
         let mut have_locations = vec::Vec::new(); //BUG: not constant size
```
jdygert-spok: Not feedback but relevant. It might be worth filtering `have_location` so that it doesn't have any ancestor-descendant pairs. I expect that to be common and would speed up the stuff below.

benz-spideroak:
> > I don't think so.
> 
> Does this mean that the implementation didn't match the specs so these bugfixes are not affecting the specs because they are getting the implementation in line with the specs? Or does this mean these behaviors are not captured in any specs? Trying to evaluate if we should add more documentation to an existing syncer spec or a new spec but I don't have enough context to make that determination.
> 
> cc: @djohnson-spideroak, @jdygert-spok, @mooreoak

I'm not sure what you're saying. A spec captures what the functionality is intended to do. This isn't changing anything about how the sync protocol works. It's optimizing the code to improve some slow paths.

gknopf-aranya:
Looks good to me. I'll defer to @jdygert-spok for a final review.

PR #512, opened 2025-12-08
policy-compiler: Return a compile error when defining a struct that uses undefined types

### LINKED ISSUE #497: policy-compiler: fields with struct and enum types are not validated
This code compiles but shouldn't.

```
fact Foo[]=>{ s struct Unknown, e enum Unknown }
struct Bar { s struct Unknown, e enum Unknown }
```

ssweetney-spok:
closes #497

### DIFF
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -1690,9 +1690,10 @@ impl<'a> CompileState<'a> {
             .iter()
             .flat_map(|ffi_mod| ffi_mod.structs.iter().map(|def| def.name.clone()));
         let command_idents = self
-            .ffi_modules
+            .policy
+            .commands
             .iter()
-            .flat_map(|ffi_mod| ffi_mod.structs.iter().map(|def| def.name.clone()));
+            .map(|def| def.identifier.name.clone());
 
         let struct_idents = self
             .policy


ssweetney-spok (commented):
Have to revert 401528d0d0042efc0cfeeee09dc54e1879151749. This doesn't work for struct field insertions without compiling the structs in sorted order

jdygert-spok:
> Have to revert [401528d](https://github.com/aranya-project/aranya-core/commit/401528d0d0042efc0cfeeee09dc54e1879151749). This doesn't work for struct field insertions without compiling the structs in sorted order

Forgot about those :thinking: So I guess the rule will be that recursive structs are fine but recursive field inserts aren't.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -261,6 +266,15 @@ impl<'a> CompileState<'a> {
         for item in items {
             match item {
                 StructItem::Field(field) => {
+                    if field
+                        .field_type
+                        .as_struct()
+                        .is_some_and(|other_ident| other_ident.name == identifier.name)
```
gknopf-aranya: Is `other_indent` the field ident? Wondering if a better name would be `field_ident`.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -242,14 +242,19 @@ impl<'a> CompileState<'a> {
...
+    /// Insert a struct definition while preventing duplicates of the struct fields.
+    // Duplicate struct names are detected in [Self::list_structs].
     pub fn define_struct(
         &mut self,
         identifier: Ident,
         items: &[StructItem<FieldDefinition>],
     ) -> Result<(), CompileError> {
-        if self.m.struct_defs.contains_key(&identifier.name) {
-            return Err(self.err(CompileErrorType::AlreadyDefined(identifier.to_string())));
+        if !self.m.struct_defs.contains_key(&identifier.name) {
```
gknopf-aranya: A compiler bool seems like a better way to keep track of whether `list_structs` was called before `define_struct`. Checking if the `struct_defs` contains a certain key is less direct. Or perhaps we can check if `struct_defs` is empty (which it should be if `list_structs` wasn't called first).

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1635,7 +1671,58 @@ impl<'a> CompileState<'a> {
         Ok(())
     }
 
+    /// Adds entries for the Struct, Effect, Fact, Command, and FFI Struct defintions
+    /// to [CompileTarget::struct_defs] before fully compiling.
+    fn list_structs(&mut self) -> Result<(), CompileError> {
+        let idents = self
```
gknopf-aranya: This is a fairly long chain of idents that I found difficult to read. One thought is to chain idents together like this so it self-documents which idents are being chained together:
```
let effects = ...
let facts = ...
let ffi_modules = ...
...
idents.chain(effects).chain(facts).chain(ffi_modules)...
```

crates/aranya-policy-compiler/src/compile/target.rs
```diff
@@ -109,6 +111,19 @@ impl CompileTarget {
             TypeKind::Never => Some(0),
         }
     }
+
+    // Filter out the definitions that aren't fully compiled
+    fn filter_defs(
```
gknopf-aranya: This won't be needed if we use different maps to keep track of exists vs. def info.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -242,14 +242,19 @@ impl<'a> CompileState<'a> {
         Ok(())
     }
 
-    /// Insert a struct definition while preventing duplicates of the struct name and fields
+    /// Insert a struct definition while preventing duplicates of the struct fields.
+    // Duplicate struct names are detected in [Self::list_structs].
```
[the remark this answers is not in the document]
ssweetney-spok: I didn't make this a doc comment since the link would be broken unless you run the docs with the flag to show private items.

crates/aranya-policy-compiler/src/compile.rs
ssweetney-spok: So maybe `field_type_ident`

ssweetney-spok:
I'd prefer to go with #513 over the approach I used in this PR of listing the types before fully defining them. It handles some error cases that this one doesn't

```
struct Foo { x struct Bar }
struct Bar { x struct Foo }
```

What do others think?

crates/aranya-policy-compiler/src/compile/target.rs
```diff
@@ -30,7 +30,9 @@ pub(crate) struct CompileTarget {
     /// Fact schemas
     pub fact_defs: BTreeMap<Identifier, FactDefinition>,
     /// Struct schemas
-    pub struct_defs: BTreeMap<Identifier, Vec<ast::FieldDefinition>>,
+    ///
+    /// A [None] value for an entry indicates that the type exists but is only partially compiled.
+    pub struct_defs: BTreeMap<Identifier, Option<Vec<ast::FieldDefinition>>>,
```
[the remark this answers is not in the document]
chip-so: IMO using "Some" and "None" to denote some more specific kind of state means you should probably create an enum to describe that state. Though I do appreciate the composability of `Option<T>`.

jdygert-spok:
I think it is uncontroversial that definition order should not matter and field insertion cannot be recursive. The latter means that we will need the topo sort at least for the field insertions.

I am unsure whether we should allow conditional field recursion, e.g. `struct List { x int, next option[List] }`. I suppose it's reasonable to disallow it for now and later loosen the restriction if needed.

To be clear, that would mean going with #513.

PR #513, opened 2025-12-08
policy-compiler: Validate fields in user defined types

### LINKED ISSUE #497: policy-compiler: fields with struct and enum types are not validated

ssweetney-spok:
closes #497

- Also Allows users to define structs out of order

```
struct Fum { b struct Bar, f struct Foo }
struct Bar { f struct Foo }
struct Foo {}
```

### DIFF
--- a/crates/aranya-policy-compiler/src/tests.rs
+++ b/crates/aranya-policy-compiler/src/tests.rs
@@ -1917,13 +1917,7 @@ const FAKE_SCHEMA: &[ModuleSchema<'static>] = &[ModuleSchema {
         }],
         return_type: ffi::Type::Bool,
     }],
-    structs: &[ffi::Struct {
-        name: ident!("TestStruct"),
-        fields: &[ffi::Arg {
-            name: ident!("enabled"),
-            vtype: ffi::Type::Bool,
-        }],
-    }],
+    structs: &[],
     enums: &[],
 }];
 


crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1625,13 +1625,70 @@ impl<'a> CompileState<'a> {
...
+        let sorted_idents = {
+            let mut buf = Vec::new();
+
+            loop {
+                let mut next = ts.pop_all();
+                if next.is_empty() {
+                    // Check for cyclic dependencies
+                    // https://docs.rs/topological-sort/0.2.2/topological_sort/struct.TopologicalSort.html#method.pop_all
+                    if !ts.is_empty() {
+                        let msg = String::from("Found cyclic dependencies when compiling structs");
```
apetkov-so: We should list the structs that have cyclic deps.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1625,13 +1625,70 @@ impl<'a> CompileState<'a> {
...
+
+        let sorted_idents = {
+            let mut buf = Vec::new();
+
+            loop {
+                let mut next = ts.pop_all();
+                if next.is_empty() {
+                    // Check for cyclic dependencies
+                    // https://docs.rs/topological-sort/0.2.2/topological_sort/struct.TopologicalSort.html#method.pop_all
+                    if !ts.is_empty() {
```
apetkov-so: Is this logic right? The doc says: "If pop_all returns an empty vector and len is not 0, there is cyclic dependencies."

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -3331,3 +3331,45 @@ fn test_action_command_persistence() {
...
+            struct Foo { s string }
+
+        "#,
+    ];
+
+    let invalid_cases = [r#"
+        struct Fum { b struct Bar, f struct Foo }
+        struct Bar { f struct Foo }
+        struct Foo { fum struct Fum } // cycle
+    "#];
```
apetkov-so: I assume `TopologicalSort` will detect indirect cycles too, e.g. if Fum didn't reference Foo, but Bar still does.

crates/aranya-policy-compiler/src/compile.rs
ssweetney-spok: I believe so. The wording confused me at first but the "len" is referring to the number of items in the `TopologicalSort` collection.

The idea is that `pop_all` returns the items that no item in `TopologicalSort` collection depend on. If `pop_all` returns an empty vec but there are still items left in `TopologicalSort`, then that means there must be a cycle because all the items left in the collection has another item in the collection that depends on it.

crates/aranya-policy-compiler/src/tests.rs
ssweetney-spok: That's right, assuming this crate has implemented the algorithm correctly.

jdygert-spok:
I think we only need the topo sort if we want to add a check for unconditionally recursive types:

```
// error
struct Foo { x struct Bar }
struct Bar { x struct Foo }

// ok
struct Foo { x struct Bar }
struct Bar { x option[struct Foo] }

// ok
struct Foo { x struct Bar }
struct Bar { x int }
```

I think that's a "nice to have" but isn't necessary since they are impossible to actually create.

We could tweak this PR easily to make it work with `option` by just not adding the dependency for optional fields, but It'd be more complex for `result` since you don't know at that point whether one of the sides doesn't cycle.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1715,6 +1700,64 @@ impl<'a> CompileState<'a> {
...
+            // this struct has no dependencies or no other struct depends on it.
+            let _ = ts.insert(&struct_def.identifier.name);
+        }
+
+        let sorted_idents = {
+            let mut buf = Vec::new();
+
+            loop {
+                let mut next = ts.pop_all();
+                if next.is_empty() {
```
gknopf-aranya: It doesn't seem like order is important for checking for cycles. Delaying the sort until ordering is actually needed could be an optimization.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1715,6 +1700,64 @@ impl<'a> CompileState<'a> {
...
+                    // Check for cyclic dependencies
+                    // https://docs.rs/topological-sort/0.2.2/topological_sort/struct.TopologicalSort.html#method.pop_all
+                    if !ts.is_empty() {
+                        let msg =
+                            format!("Found cyclic dependencies when compiling structs: {ts:#?}");
```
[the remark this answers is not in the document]
ssweetney-spok: It took more effort `topo_sort` to get the output we wanted than I expected. Iterating over the edges (multiple `hashset`s) was tedious and it seemed easier to keep track of the current path in DFS.

https://github.com/aranya-project/aranya-core/pull/513/changes/edd57a1206df9f63570bd090ff3209b10c1de1ca

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -3284,3 +3335,46 @@ fn test_action_command_persistence() {
         assert_eq!(err, expected);
     }
 }
+
+#[test]
+fn test_structs_listed_out_of_order() {
+    let valid_cases = [
+        r#"
+            effect Fi { fum struct Fum }
+            struct Fum { b struct Bar, f struct Foo }
```
gknopf-aranya: May get better test coverage if one field's struct is defined after while another field's struct is defined before the current struct.

crates/aranya-policy-compiler/src/compile/lower.rs
```diff
@@ -42,7 +43,8 @@ impl CompileState<'_> {
     /// - the fields defined in the struct are present, and have the correct types
     /// - there are no duplicate fields
     fn lower_struct_literal(&mut self, s: &NamedStruct) -> Result<thir::NamedStruct, CompileError> {
-        let Some(struct_def) = self.m.struct_defs.get(&s.identifier.name).cloned() else {
+        let Some(State::Compiled(struct_def)) = self.m.struct_defs.get(&s.identifier.name).cloned()
```
gknopf-aranya: Is this compiling the struct with `State::Compiled()`? Is the difference that this code used to not compile the struct first?

crates/aranya-policy-compiler/src/compile/target.rs
```diff
@@ -11,6 +11,27 @@ use aranya_policy_module::{
...
+        match value {
+            State::Compiling => None,
+            State::Compiled(c) => Some(c),
+        }
+    }
+}
+
+const _: () = {
+    type StructDef = ast::StructDefinition;
+    assert!(size_of::<Vec<StructDef>>() == size_of::<State<Vec<StructDef>>>());
```
gknopf-aranya: How are these the same size when one is wrapped in an extra `State` enum?

crates/aranya-policy-compiler/src/compile/target.rs
```diff
@@ -109,6 +132,20 @@ impl CompileTarget {
             TypeKind::Never => Some(0),
         }
     }
+
+    // Filter out the definitions that aren't fully compiled
```
gknopf-aranya: This comment says the function is filtering out definitions that aren't fully compiled, but the comment above `debug_assert` indicates all the definitions should be fully compiled by the time this method is called. Is this method supposed to filter or verify all the definitions are compiled?

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -2919,6 +2921,55 @@ fn test_function_arguments_with_undefined_types() {
     }
 }
 
+#[test]
+fn test_structs_with_undefined_types() {
+    let cases = [
+        (
+            r#"
+            fact Foo[]=>{ s struct Unknown }
```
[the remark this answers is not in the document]
ssweetney-spok: Good call

crates/aranya-policy-compiler/src/compile/target.rs
ssweetney-spok: Null pointer optimization. It's kind of documented [here](https://doc.rust-lang.org/std/option/#representation).

It's not exclusive to `Option`. I believe you just need an enum with a variant that contains no data and a "pointer" type

crates/aranya-policy-compiler/src/compile/topo.rs
```diff
@@ -0,0 +1,156 @@
...
+        let all_nodes: IndexSet<&T> = graph
+            .iter()
+            .flat_map(|(node, deps)| std::iter::once(node).chain(deps.iter()))
+            .collect();
+
+        let mut state: IndexMap<&T, State> =
+            all_nodes.iter().map(|&n| (n, State::Unvisited)).collect();
+        let mut result = Vec::with_capacity(all_nodes.len());
+        // Keep track of the current path so we can print the cycle in case of an error.
+        let mut path = Vec::new();
```
gknopf-aranya: Can the size of this be estimated or upper-bounded so `with_capacity()` can be used?

crates/aranya-policy-compiler/src/compile/target.rs
```diff
@@ -30,7 +51,7 @@ pub(crate) struct CompileTarget {
     /// Fact schemas
     pub fact_defs: BTreeMap<Identifier, FactDefinition>,
     /// Struct schemas
-    pub struct_defs: BTreeMap<Identifier, Vec<ast::FieldDefinition>>,
+    pub struct_defs: BTreeMap<Identifier, State<Vec<ast::FieldDefinition>>>,
```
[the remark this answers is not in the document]
ssweetney-spok: [Added](https://github.com/aranya-project/aranya-core/pull/513/changes/bc966112c6e035b845c13cd80b7f53f75a897fbe#diff-e2b429adfc01f0f40d066e326bd3bb29a8f43a41819c8ceae5b44de1d60e6c31R198-R199)

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1567,28 +1582,128 @@ impl<'a> CompileState<'a> {
         Ok(())
     }
 
-    fn define_interfaces(&mut self) -> Result<(), CompileError> {
+    /// Adds entries for the Struct, Effect, Fact, Command, and FFI Struct defintions
+    /// to [CompileTarget::struct_defs] before fully compiling.
```
ssweetney-spok:
```suggestion
    /// to [CompileState::structs] before fully compiling.
```

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1395,9 +1410,9 @@ impl<'a> CompileState<'a> {
                         .assume("duplicates are prevented by compile_struct")?;
                 }
                 StructItem::StructRef(ref_name) => {
-                    let struct_def = self.m.struct_defs.get(&ref_name.name).ok_or_else(|| {
-                        self.err(CompileErrorType::NotDefined(ref_name.to_string()))
-                    })?;
+                    let Some(struct_def) = self.m.struct_defs.get(&ref_name.name) else {
+                        return Err(self.err(CompileErrorType::NotDefined(ref_name.to_string())));
+                    };
```
ssweetney-spok: revert

crates/aranya-policy-compiler/src/compile/topo.rs
```diff
@@ -0,0 +1,141 @@
...
+impl Display for CycleError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // This outputs a comma separated list of the nodes in each cycle/SCC.
+        // A direct path is not shown.
+        let cycles: Vec<_> = self
+            .0
+            .iter()
+            .map(|cycle| {
+                let mut buf = "[]".to_owned();
+                buf.insert_str(1, &cycle.as_slice().join(", "));
```
chip-so: Why not `format!("[{}]", cycle.as_slice.join(", "))`?

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -193,6 +198,8 @@ struct CompileState<'a> {
     function_signatures: BTreeMap<Identifier, FunctionSignature>,
     /// Builtin functions which have special behavior when compiling a function call.
     builtin_functions: BTreeMap<Identifier, BuiltinHandler>,
+    /// The identifiers of all user-defined and auto-generated structs.
+    structs: BTreeSet<&'a Identifier>,
```
chip-so: Why does this live in `CompileState`?

ssweetney-spok: After thinking about it more, this and `list_structs` aren't _really_ necessary for this PR. It's mostly carry over from the [initial approach](https://github.com/aranya-project/aranya-core/pull/512) to solve this problem without having to sort the type definitions. That initial approach worked for simple cases where types had fields like `field_name struct Foo` but didn't work for struct field insertions.

Although, `list_structs` is the only place where duplicate type names are checked at the moment. I did try removing it and noticed a failing test case:

```
// Should give an "already defined" error.
struct Foo {}
fact Foo[]=>{}
```

chip-so: Ah, I see you've just changed `list_structs()` to better describe its purpose. 👍

chip-so (approved):

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1567,28 +1576,128 @@ impl<'a> CompileState<'a> {
...
+                return Err(self.err(CompileErrorType::AlreadyDefined(ident.to_string())));
+            }
+        }
+
+        Ok(())
+    }
+
+    fn sorted_type_definitions(
+        &self,
+    ) -> Result<impl Iterator<Item = UserType<'a>> + use<'a>, CompileError> {
```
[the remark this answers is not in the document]
ssweetney-spok: It seems like we should always be able to compile FFI structs first so they don't need to be here

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1567,28 +1576,128 @@ impl<'a> CompileState<'a> {
...
-        for effect in &self.policy.effects {
-            let fields: Vec<StructItem<FieldDefinition>> = effect
-                .items
+        for effect_def in &self.policy.effects {
+            let deps = effect_def.items.iter().filter_map(|item| match item {
+                // { +Foo }
+                StructItem::StructRef(ident) => Some(&ident.name),
+                // { field_name struct Foo }
+                StructItem::Field(field) => field.field_type.as_struct().map(|ident| &ident.name),
+            });
```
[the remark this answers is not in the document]
ssweetney-spok: That'd be ideal but I wasn't sure how to get around the `EffectFieldDefinition` and `FieldDefinition` differences.

I _think_ (using `repr`) that we can safely cast `&EffectFieldDefinition` to `&FieldDefinition` but that doesn't work for `&StructItem<EffectFieldDefinition>` and `&StructItem<FieldDefintion>`

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1567,28 +1576,128 @@ impl<'a> CompileState<'a> {
...
+        for command_def in &self.policy.commands {
+            let deps = command_def.fields.iter().filter_map(extract_struct_ident);
+
+            topo.insert(&command_def.identifier.name, deps);
+        }
+
+        let sorted_idents = topo.sort().map_err(|err| self.err(err.into()))?;
+
+        // TODO(Steve): Store hashmaps in `aranya_policy_ast::ast::Policy`'s fields to make this more efficient.
+        let sorted_defs = sorted_idents.into_iter().filter_map(|ident| {
```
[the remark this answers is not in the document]
jdygert-spok: Not sure, i think FFI structs might end up in the global namespace which does seem pretty bad.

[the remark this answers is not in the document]
ssweetney-spok: > How about we construct a HashMap<&str, UserType<'_>> while we are inserting the dependencies above ..

This turned out nicely since it makes it easier to get the span information (except for FFI struct defs) now when finding a duplicate type

crates/aranya-policy-compiler/src/compile.rs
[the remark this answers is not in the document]
ssweetney-spok: Added

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -1915,7 +1917,13 @@ const FAKE_SCHEMA: &[ModuleSchema<'static>] = &[ModuleSchema {
...
         return_type: ffi::Type::Bool,
     }],
-    structs: &[],
+    structs: &[ffi::Struct {
+        name: ident!("TestStruct"),
+        fields: &[ffi::Arg {
+            name: ident!("enabled"),
+            vtype: ffi::Type::Bool,
+        }],
+    }],
```
ssweetney-spok:
```suggestion
    structs: &[],
```

Revert. Don't need this

PR #516, opened 2025-12-10
Give a trace error when the action is not found

chip-so:
Found this while intentionally testing a bad action. It gave me no traces whatsoever, just the `InternalErrro`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -131,7 +131,7 @@ checksum = "c1fd03a028ef38ba2276dce7e33fcd6369c158a1bca17946c4b1b701891c1ff7"
 
 [[package]]
 name = "aranya-afc-util"
-version = "0.17.0"
+version = "0.18.0"
 dependencies = [
  "aranya-afc-util",
  "aranya-crypto",
@@ -158,7 +158,7 @@ dependencies = [
 
 [[package]]
 name = "aranya-capi-codegen"
-version = "0.5.0"
+version = "0.6.0"
 dependencies = [
  "aho-corasick",
  "anyhow",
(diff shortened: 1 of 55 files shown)


crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -648,11 +648,10 @@ impl<E: aranya_crypto::Engine> Policy for VmPolicy<E> {
...
         let VmAction { name, args } = action;
 
-        let def = self
-            .machine
-            .action_defs
-            .get(&name)
-            .ok_or(EngineError::InternalError)?;
+        let def = self.machine.action_defs.get(&name).ok_or_else(|| {
+            error!("action `{name}` not found");
+            EngineError::InternalError
```
gknopf-aranya: Are there any other errors in the runtime we want to add tracing to?

chip-so: All of the other ones in `call_action()` at least had accompanying `error!()` calls. I could do a broader search.

gknopf-aranya: Took a quick look as well and most of them do appear to have tracing already. May have found a few errors in the engine, storage, and sync protocol that don't have tracing:
```
CommandPlacement::OffGraph => {
    // Immediate check failure.
    return Err(EngineError::Check);
}
```
```
// Storage ID is the id of the init command by definition.
if self.storage_id.as_base() != command.id().as_base() {
    return Err(ClientError::InitError);
}
```
```
if update.commands.is_empty() {
    return Err(StorageError::EmptyPerspective);
}
```
```
S::PartialSync => {
    return Err(SyncError::NotReady);
}
```

gknopf-aranya (approved):

crates/aranya-runtime/src/vm_policy.rs
chip-so: I'm happy to add some tracing here, but these are not parts of the code I'm that familiar with. I guess I'll make an attempt and we'll discuss?

Issue #469, opened 2025-10-30
policy-lang: spans include trailing whitespace and comments


jdygert-spok:
I don't think there is a simple way to fix this. Maybe we just manually trim them?

PR #524, opened 2026-01-02
Improve parser error rendering

ssweetney-spok:
Improve error messages encountered during parsing.

## Example

**before:**
```
Reserved identifier: line 1 column 5: this
```

**after:**
```
error: Reserved identifier
  |
1 | enum this { A }
  |      ^^^^ found reserved identifier: `this`
```

Differences between other error types can be found in the snapshot tests: crates/aranya-policy-lang/tests/data

## TODO:
~~- [ ] Audit annotate-snippets~~ Added exemption

### DIFF
--- a/crates/aranya-policy-lang/src/lang/parse.rs
+++ b/crates/aranya-policy-lang/src/lang/parse.rs
@@ -5,8 +5,8 @@ use aranya_policy_ast::{
     EnumDefinition, EnumReference, ExprKind, Expression, FactField, FactLiteral, FieldDefinition,
     ForeignFunctionCall, FunctionCall, Ident, IfStatement, InternalFunction, LetStatement,
     MapStatement, MatchArm, MatchExpression, MatchExpressionArm, MatchPattern, MatchStatement,
-    NamedStruct, Param, Persistence, ReturnStatement, Statement, StmtKind, Text, TypeKind,
-    UpdateStatement, VType, Version, ident,
+    NamedStruct, Param, Persistence, ResultPattern, ResultTypeKind, ReturnStatement, Statement,
+    StmtKind, Text, TypeKind, UpdateStatement, VType, Version, ident,
 };
 use buggy::BugExt as _;
 use pest::{
@@ -205,14 +205,13 @@ impl ChunkParser<'_> {
     /// Parse a type token (one of the types under Rule::vtype) into a
     /// Parse a type token into a VType.
     fn parse_type(&self, token: Pair<'_, Rule>) -> Result<VType, ParseError> {
-        self.parse_type_inner(token, TypeStyle::Unknown, true, None)
+        self.parse_type_inner(token, TypeStyle::Unknown, None)
     }
 
     fn parse_type_inner(
         &self,
         token: Pair<'_, Rule>,
-        mut style: TypeStyle,
-        allow_option: bool,
+        style: TypeStyle,
         outer_span: Option<Span<'_>>,
     ) -> Result<VType, ParseError> {
         let pest_span = token.as_span();
@@ -244,23 +243,26 @@ impl ChunkParser<'_> {
                         );
                     }
                 }
-                match (style, is_old) {
-                    (TypeStyle::Unknown, true) => style = TypeStyle::Old,
-                    (TypeStyle::Unknown, false) => style = TypeStyle::New,
-                    (TypeStyle::Old, true) | (TypeStyle::New, false) if allow_option => {}
-                    _ => {
-                        let outer_span = outer_span.assume("outer span was passed in")?;
-                        let outer_ast_span = self.to_ast_span(outer_span)?;
-                        return Err(ParseError::new(
-                            error::InvalidNestedOption {
-                                outer: outer_ast_span,
-                                inner: span,
-                            },
-                            String::from("Unable to use `optional T` for nested optional types."),
-                            Some(self.to_ast_span(outer_span)?),
-                        ));
-                    }
+
+                // Disallow mixing old and new optional syntax.
+                if style == TypeStyle::Old || (is_old && style == TypeStyle::New) {
+                    let outer_span = outer_span.assume("outer span was passed in")?;
+                    let outer_ast_span = self.to_ast_span(outer_span)?;
+                    return Err(ParseError::new(
+                        error::InvalidNestedOption {
+                            outer: outer_ast_span,
+                            inner: span,
+                        },
+                        String::from("Unable to use `optional T` for nested optional types."),
+                        Some(self.to_ast_span(outer_span)?),
+                    ));
                 }
+
+                let style = if is_old {
+                    TypeStyle::Old
+                } else {
+                    TypeStyle::New
+                };
                 let mut pairs = token.clone().into_inner();
                 let token = pairs.next().ok_or_else(|| {
                     self.to_ast_span(token.as_span()).map_or_else(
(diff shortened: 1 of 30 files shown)


crates/aranya-policy-lang/tests/data/keyword_collision/fact_key_name.snap
```diff
@@ -2,4 +2,7 @@
 source: crates/aranya-policy-lang/tests/tests.rs
 expression: err
 ---
-Reserved identifier: line 1 column 8: int
+error: Reserved identifier
+  |
+1 | fact x[int int]=>{}
```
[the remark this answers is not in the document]
jdygert-spok: The second int is the type, so it is valid anyways.

crates/aranya-policy-lang/src/lang/parse/error.rs
```diff
@@ -77,64 +101,176 @@ impl Display for ParseErrorKind {
...
+            kind: Box::new(kind),
             message,
-            location,
+            span,
+            line_offset: None,
         }
     }
 
+    pub(crate) fn into_report(self) -> ReportCell {
+        let Self {
```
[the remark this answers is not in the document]
ssweetney-spok: I see what you mean. I've only added suggestions for 2 error types so far and it already felt awkward trying to refactor this function.

crates/aranya-policy-lang/tests/data/match_expression/block_without_subexpression.snap
```diff
@@ -2,9 +2,11 @@
...
-4 |             1
-  |             ^---
-  |
-  = expected identifier, action_call, publish_statement, let_statement, check_statement, match_statement, if_statement, finish_statement, map_statement, create_statement, update_statement, delete_statement, emit_statement, return_statement, or debug_assert
+  = note:  --> 4:13
+            |
+          4 |             1
+            |             ^---
+            |
+            = expected identifier, action_call, publish_statement, let_statement, check_statement, match_statement, if_statement, finish_statement, map_statement, create_statement, update_statement, delete_statement, emit_statement, return_statement, or debug_assert
```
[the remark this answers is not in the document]
jdygert-spok: This error ought to be "invalid block expression", since that's the only expression that starts with `{`, but we can't really do that since pest doesn't offer a cut operator. Another reason to move off it eventually.

crates/aranya-policy-lang/tests/data/old_addition.snap
```diff
@@ -0,0 +1,19 @@
+---
+source: crates/aranya-policy-lang/tests/tests.rs
+expression: err
+---
+error: Invalid operator
+  |
+2 |       return 3 + 4
+  |  ____________^
```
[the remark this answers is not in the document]
ssweetney-spok: Fixed this so that the message refers to operator instead of the whole expression.

> Looks like it's pointing to the whole span of 3 + 4, which unfortunately includes all the trailing whitespace due to https://github.com/aranya-project/aranya-core/issues/469, so it looks weird here.

Ah, that makes sense.

crates/aranya-policy-lang/src/lang/parse.rs
```diff
@@ -35,66 +35,71 @@ type FieldsAndSources = (Vec<(Ident, Expression)>, Vec<Ident>);
 
 /// Captures the iterator over a Pair's contents, and the span
 /// information for error reporting.
-struct PairContext<'a> {
+struct PairContext<'a, 'b> {
     pairs: RefCell<Pairs<'a, Rule>>,
     span: Span<'a>,
+    to_ast_span: Box<dyn Fn(Span<'a>) -> Result<ast::Span, ParseError> + 'b>,
```
ssweetney-spok: This couples `PairContext` closely with `ChunkParser` now.

Maybe instead of this, `PairContext` should just hold a reference to `ChunkParser`? 🤔

jdygert-spok: I think now that we aren't using the pest errors directly, we don't actually need to pass the chunk offset into the parser, since we can just fix it up at the caller instead (i.e. `parse_policy_document`.

ssweetney-spok: Gotcha, that makes sense

ssweetney-spok: Removing the chunk offset is trickier than I thought it'd be cause we build up the `ast::Policy` object by parsing each chunk. After each chunk is parsed we need a way to know if an `AST item`'s span needs to be adjusted or not. I thought about doing something like:

` enum SpanType { adjusted(Span), non_adjusted(Span) }` but that didn't feel great.

I think we could get rid of the chunk offset logic by writing a parser that we can feed the entire policy document to.

jdygert-spok: Oh right, because we parse into the AST object. We'd need `parse_chunk` to return something like `Vec<TopLevelItem>` if we wanted to fix up the spans there. Or I guess we could parse into a new AST object and merge it with the existing one after fixup :/

chip-so: If we want to swap it to a vec of `TopLevelItem` I'm fine with that. The current structure made sense to me for ordering the compilation but it was a fairly arbitrary decision.

crates/aranya-policy-lang/tests/data/invalid_text/escaped_nul_byte.snap
```diff
@@ -1,5 +1,9 @@
 ---
 source: crates/aranya-policy-lang/tests/tests.rs
+assertion_line: 99
 expression: err
 ---
-Invalid string: line 1 column 9: string contained nul byte
+error: Invalid string
+  |
+1 | let x = "a\x00b"
+  |         ^^^^^^^^ string contained nul byte
```
chip-so: I know this wasn't part of this PR, but I think in a user-facing error we should spell out "null". Or if we really want to be ISO 646 about it, say "NUL".

crates/aranya-policy-lang/tests/data/error_line_number_in_chunks/two_chunk_custom_error_with_patch.snap
```diff
@@ -0,0 +1,17 @@
...
+15 |         return x + y
+   |                  ^ found `+`
+   |
+help: you might have meant to use an arithmetic function
+   |
+15 -         return x + y
+15 +         return saturating_add(x, y)
+   |
+15 -         return x + y
+15 +         return add(x, y)
```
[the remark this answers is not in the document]
ssweetney-spok: Ah, that's a good point. `add` and `saturating_add` do have different return types. 

It's likely that users will encounter another error if they just copy `return add(x, y)` as is so maybe `check_unwrap` should be in the suggestion

crates/aranya-policy-lang/tests/data/invalid_text/escaped_nul_byte.snap
ssweetney-spok: Changed to "null"

crates/aranya-policy-lang/tests/data/neg_number_out_of_range.snap
```diff
@@ -0,0 +1,8 @@
+---
+source: crates/aranya-policy-lang/tests/tests.rs
+expression: err
+---
+error: Invalid number
+  |
+2 | let small_num = -9223372036854775809
+  |                 ^^^^^^^^^^^^^^^^^^^^ number too small to fit in target type
```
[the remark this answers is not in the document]
ssweetney-spok: https://github.com/aranya-project/aranya-core/pull/524/commits/edab6d2ea7e7d21ffbf2e3e310ad04e06bf8edc8

crates/aranya-policy-lang/src/lang/parse/error.rs
```diff
@@ -1,23 +1,71 @@
...
+        write!(
+            f,
+            "Invalid policy version {found}, supported version is {required}"
+        )
+    }
+}
+
+/// Invalid usage of option/optional syntax
+///
+///(crate) `optional option[T]`, `optional optional T`, etc.
```
jdygert-spok: `(crate)` probably a typo

crates/aranya-policy-lang/src/lang/parse.rs
```diff
@@ -35,45 +34,50 @@ type FieldsAndSources = (Vec<(Ident, Expression)>, Vec<Ident>);
 
 /// Captures the iterator over a Pair's contents, and the span
 /// information for error reporting.
-struct PairContext<'a> {
-    pairs: RefCell<Pairs<'a, Rule>>,
-    span: Span<'a>,
+struct PairContext<'i, 'p> {
+    pairs: RefCell<Pairs<'i, Rule>>,
+    span: Span<'i>,
+    to_ast_span: Box<dyn Fn(Span<'i>) -> Result<ast::Span, ParseError> + 'p>,
```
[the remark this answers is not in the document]
ssweetney-spok: Yea, that should work. I'd be OK with that since I wasn't fond of adding the `PairContext::to_ast_span` field

chip-so (approved):

PR #526, opened 2026-01-05
chore: update tooling and deps

jdygert-spok:
cargo audit update needed for new cvss format. rkyv update needed for vulnerability.


jdygert-spok:
> We dont need a Cargo.toml update for rkyv? I see the lockfile updated

Not really, although I guess it'd help by force downstream users of the library crates to a newer version, if they already have a lockfile with an older version.

PR #527, opened 2026-01-05
chore: force update rkyv

jdygert-spok:
This will ensure downstream crate users with existing lockfiles will use the newer version too.

(Update for [RUSTSEC-2026-0001](https://rustsec.org/advisories/RUSTSEC-2026-0001.html))


ssweetney-spok (approved):

chip-so (approved):

gknopf-aranya:
@chip-so, @jdygert-spok looks like the embedded repo no longer builds with this change:
https://github.com/aranya-project/aranya-core/actions/runs/20724320714/job/59495958164?pr=527

PR #528, opened 2026-01-05
Close #525: Require that all variables be used.

### LINKED ISSUE #525: Don't allow unused variables.
Require that all local variables, including return values are used. Check nested scopes too.

```
function foo() int {
  var x = 5
  return 0 // x is unused
}
```

```
function foo() int {
  let a = 0
  if a > 0 {
    let b = 0 // b is unused
  }
  return a
}
```

```
function foo() int {
  return bar() // OK, value is consumed, even though there's no identifier binding
}
```

apetkov-so:
This is important for handling `result` values - we don't want them to be ignored.


crates/aranya-policy-compiler/src/tracer/analyzers/unused_result_analyzer.rs
```diff
@@ -0,0 +1,109 @@
...
+
+use aranya_policy_ast::{Identifier, ident};
+use aranya_policy_module::{Instruction, ModuleV0};
+
+use super::{Analyzer, AnalyzerStatus};
+use crate::tracer::TraceError;
+
+/// Ensures that all defined variables are used.
+#[derive(Clone)]
+pub struct UnusedResultAnalyzer {
```
ssweetney-spok: Should probably be renamed since it affects more than just `Result`s

crates/aranya-policy-compiler/src/tracer/analyzers/unused_result_analyzer.rs
```diff
@@ -0,0 +1,109 @@
...
+                    )));
+                }
+            }
+
+            // Def creates a must-use obligation (unless it's a predefined variable)
+            Instruction::Def(ident) => {
+                if !self.predefined_vars.contains(ident) {
+                    self.current_scope().insert(ident.clone());
+                }
+            }
```
ssweetney-spok: It'd be nice for this to work for `this` inside of a `seal` block as well. I think that'd be doable after https://github.com/aranya-project/aranya-core/pull/533

chip-so (commented):
Looking good, once concern.

crates/aranya-policy-compiler/src/tracer/analyzers/unused_var_analyzer.rs
```diff
@@ -0,0 +1,109 @@
...
+            }
+
+            // Pop scope and check for unused variables in that scope
+            Instruction::End => {
+                let Some(scope) = self.scope_stack.pop() else {
+                    return Ok(AnalyzerStatus::Ok);
+                };
+                if let Some(var) = scope.iter().next() {
+                    return Ok(AnalyzerStatus::Failed(format!(
+                        "unused variable: `{}`",
```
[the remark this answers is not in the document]
apetkov-so: Changed to IndexSet to preserve order, and reporting all unused vars.

crates/aranya-policy-compiler/src/tracer/analyzers/unused_var_analyzer.rs
```diff
@@ -0,0 +1,109 @@
...
+                    )));
+                }
+            }
+
+            // Def creates a must-use obligation (unless it's a predefined variable)
+            Instruction::Def(ident) => {
+                if !self.predefined_vars.contains(ident) {
+                    self.current_scope().insert(ident.clone());
+                }
+            }
```
apetkov-so: That would break a lot of test code that uses `return todo()`, though. Should built-ins like `this` and `envelope` be required to be used? @chip-so

crates/aranya-policy-compiler/src/tracer/analyzers/unused_var_analyzer.rs
```diff
@@ -0,0 +1,110 @@
...
+            Instruction::Get(ident) => {
+                self.remove_from_scopes(ident);
+            }
+
+            // At function exit points, check that all obligations in all scopes are satisfied
+            Instruction::Return | Instruction::Exit(_) => {
+                for scope in &self.scope_stack {
+                    if let Some(var) = scope.iter().next() {
+                        return Ok(AnalyzerStatus::Failed(format!(
+                            "unused variable: `{}`",
```
chip-so: `Instruction::End` lists multiple variables but this doesn't.

crates/aranya-policy-compiler/src/tracer/analyzers/unused_var_analyzer.rs
ssweetney-spok: > We do still have something that maps instruction locations to spans though, right?

Oh right. There is the codemap

chip-so (dismissed):

crates/aranya-policy-compiler/src/tracer/analyzers/unused_var_analyzer.rs
apetkov-so: The tracer already prints row/col numbers:

```
Trace `baz` policy: unused variable(s): `x`, `z` at row 7 col 17:
return y
```

We should report each var individually, but that requires some refactoring to return more than one error. I think we should do that in a different ticket.

ssweetney-spok (dismissed):

chip-so (dismissed):

crates/aranya-policy-compiler/src/tracer/analyzers/unused_var_analyzer.rs
```diff
@@ -0,0 +1,114 @@
+use aranya_policy_ast::{Identifier, ident};
+use aranya_policy_module::{Instruction, ModuleV0};
+use indexmap::IndexSet;
+
+use super::{Analyzer, AnalyzerStatus};
+use crate::tracer::TraceError;
+
+/// Ensures that all defined variables are used.
+#[derive(Clone)]
+pub struct UnusedVarAnalyzer {
```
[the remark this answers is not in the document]
ssweetney-spok: How would #477 address this case?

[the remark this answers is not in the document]
apetkov-so: This is harder than I thought. The approach I tried was to iterate over the the failures vector, extract the identifiers, and for each one, check the successful paths for a Get instruction with that identifier. If found, remove the var from the list  - it is used.

But that's fragile because the same identifier can be declared multiple times (in different scopes) in the same function. Just because we found `Get(n)`, doesn't mean it was the correct `n`. We have a scope_stack, but that's per analyzer, and it's gone by the time we get to post_analyze.

chip-so (dismissed):

crates/aranya-policy-compiler/src/tracer/analyzers/unused_var_analyzer.rs
```diff
@@ -0,0 +1,114 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl UnusedVarAnalyzer {
+    pub fn new() -> Self {
+        let mut predefined_vars = IndexSet::new();
+        predefined_vars.insert(ident!("this"));
+        predefined_vars.insert(ident!("envelope"));
```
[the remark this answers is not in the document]
apetkov-so: > Envelope isn't a keyword, so `let envelope = 42` is valid and would be missed by this.

Predefined vars like 'envelope' and 'this' are now added only for command contexts.

crates/aranya-policy-compiler/src/cfg/unused.rs
```diff
@@ -0,0 +1,104 @@
...
+//! neither `a` nor `b` is flagged.
+//!
+//! This works because the policy language doesn't allow shadowing: on any
+//! given path a name is bound at most once, so every reachable `Get(name)`
+//! reads the same definition. (Full liveness with kill-sets would only be
+//! needed if dead stores were possible.)
+//!
+//! The forward search uses a visited-set, so loops (e.g. the backward `Jump` in
+//! a `map`) terminate. The `map ... as` binding is set by `QueryNext`, not
+//! `Def`, so it is intentionally not checked here.
```
chip-so: Does this mean that an unused `map` iterator variable won't be detected by this?

chip-so (dismissed):

crates/aranya-policy-compiler/src/cfg/unused.rs
apetkov-so: Yes, we didn't check `map` iter idents. But when I added a test case for a `map`, that uncovered another issue: the tracer began to crash due to OOM - it didn't have a visited set to prevent infinite loops.

Fixed the tracer, and added test case for `map`.

crates/aranya-policy-compiler/src/cfg/graph.rs
jdygert-spok, line 1: I was surprised to see that this doesn't take `Block` and `End` into account.

Now I see that both the existing `ValueAnalyzer` and the unused var check don't handle this case correctly:

```
let x = {
    let x = 42
    : 0
}
return x
```

One could argue this is bad code but it is currently allowed since it doesn't actually shadow. See test case `inner_block_redefine`.

apetkov-so: Well, the CFG itself ignores `Block`/`End` because they are just scope markers, they don't affect control flow.
But you're right that scope does matter for cases like this.

apetkov-so: FWIW, we can't do `let x = { let x = ... }` - we get an AlreadyDefined error.  If we change the inner ident to something else, the test passes, i.e. we get a warning about the ident being unused.

jdygert-spok: What are you running to get an `AlreadyDefined` error during compilation? The only error I get for this is "set twice" from the `ValueAnalyzer` in validation.

apetkov-so: My bad, I misinterpreted the "set twice" error as "defined twice". So the issue is that we need to track the scope of idents in the cfg?

PR #541, opened 2026-01-21
syncer: remove server address from protocol

### LINKED ISSUE #540: syncer: remove "return address" from sync protocol
This is a transport-specific concern. Most transports have a way to access the peer's client address. If this is different from their server address, then the transport should handle that some way.

jdygert-spok:
Closes #540.


jdygert-spok:
Corresponding `aranya` changes: https://github.com/aranya-project/aranya/pull/693

PR #543, opened 2026-01-21
syncer: add graph ID to more of protocol

### LINKED ISSUE #538: syncer: include graph id in hello sync and push sync protocols
Include the graph id in the sync protocol for hello and push sync.

A previous syncer implementation was looking up the graph id based on the PSK used by the QUIC connection.
This is no longer possible with the mTLS implementation since the same certs are used for different graphs.

Including the graph ID in the sync protocol has the following benefits:
- We can sync multiple teams over the same QUIC connection.
- The transport and crypto implementations can be changed without breaking the sync protocol since the protocol doesn't have a dependency on the transport implementation.
- Graph ID provides relevant context for the graph command addresses already in the protocol.

cc: @benz-spideroak, @djohnson-spideroak, @jdygert-spok, @mooreoak

jdygert-spok:
Closes #538.

### DIFF
--- a/crates/aranya-quic-syncer/src/lib.rs
+++ b/crates/aranya-quic-syncer/src/lib.rs
@@ -236,19 +236,19 @@ where
                 .expect("Failed to extend received data from slice");
         }
         // An empty response means we're up to date and there's nothing to sync.
-        if !received_data.is_empty() {
-            if let Some(cmds) = syncer.receive(&received_data)? {
-                received = cmds.len();
-                let mut trx = client.transaction(storage_id);
-                client.add_commands(&mut trx, sink, &cmds)?;
-                client.commit(&mut trx, sink)?;
-                client.update_heads(
-                    storage_id,
-                    cmds.iter().filter_map(|cmd| cmd.address().ok()),
-                    heads,
-                )?;
-                self.push(storage_id)?;
-            }
+        if !received_data.is_empty()
+            && let Some(cmds) = syncer.receive(&received_data)?
+        {
+            received = cmds.len();
+            let mut trx = client.transaction(storage_id);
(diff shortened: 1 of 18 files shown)


crates/aranya-quic-syncer/src/lib.rs
gknopf-aranya: Looks like graph ID and storage ID are used interchangeably here. At the protocol level, graph seems to be the right abstraction since we are syncing graphs. Storage relates to how the graph is stored on disk which is an implementation detail of the daemon/runtime.

jdygert-spok: Yeah, I was thinking during this that we should change it in many places. But I would rather do that on its own PR. I stuck to existing names here.

gknopf-aranya: We could at least update it in the sync protocol API here. The other internal changes could happen later without a breaking change.

crates/aranya-runtime/src/sync/dispatcher.rs
gknopf-aranya: `Poll` sync should also specify the graph ID

gknopf-aranya: Oh, it's already there in `SyncRequest`

gknopf-aranya (approved):

gknopf-aranya:
Broke embedded repo. Guessing because of the breaking change in the sync protocol API.

mooreoak (dismissed):

gknopf-aranya (dismissed):

djohnson-spideroak (dismissed):

djohnson-spideroak (approved):

PR #549, opened 2026-01-21
Policy Runner

### LINKED ISSUE #548: Create Policy Tester
We need some tool that can test policy as-written and allow the user to see and check the effects produced.

chip-so:
Still have some cleanup and features to implement, but this should be working. See the [README](https://github.com/aranya-project/aranya-core/blob/8fd3fdc472049e356626fc097c2f51d0c89175de/crates/aranya-policy-runner/README.md).

Closes #548

### DIFF
--- a/crates/aranya-policy-runner/src/lib.rs
+++ b/crates/aranya-policy-runner/src/lib.rs
@@ -42,26 +42,12 @@ enum OutputDestination {
     File(fs::File),
 }
 
-impl Clone for OutputDestination {
-    fn clone(&self) -> Self {
-        match self {
-            Self::Stdout => Self::Stdout,
-            Self::File(f) => {
-                // As far as I can tell, this can only fail if the OS refuses to duplicate the
-                // underlying FD (e.g. you've hit the ulimit). So assuming this will succeed seems
-                // more or less on the same level as assuming a memory allocation will succeed.
-                Self::File(f.try_clone().expect("cannot clone file"))
-            }
-        }
-    }
-}
-
 /// A `RunSchedule` keeps track of which action thunks are associated
 /// with which file, so they can be run in sequence.
 struct RunSchedule<'a> {
-    pub file_path: &'a Path,
-    pub preamble_values: Vec<Value>,
-    pub thunk_range: Range<usize>,
+    file_path: &'a Path,
+    preamble_values: Vec<Value>,
+    thunk_range: Range<usize>,
 }
 
 /// The core Policy Runner object
@@ -83,7 +69,7 @@ struct RunSchedule<'a> {
 /// | `with_deterministic_rng(bool)` | Use a deterministic RNG implementation. |
 /// | `with_marker(bool)` | Print a marker between run files that shows which run file is producing the effects. |
 /// | `with_validator(bool)` | Run the compiler validator before execution. |
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 #[must_use]
 pub struct PolicyRunner {
     working_directory: WorkingDirectory,


crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -179,7 +179,7 @@ impl fmt::Display for StatementContext {
 type BuiltinHandler = fn(&mut CompileState<'_>) -> Result<(), CompileError>;
 
 /// The "compile state" of the machine.
-struct CompileState<'a> {
+pub struct CompileState<'a> {
```
gknopf-aranya: Is this `pub` because the policy tester needs it? I wonder if we should keep the policy tester inside the compiler crate so compiler internals don't need to be exposed via the crate's API.
This depends on whether the policy tester is a CLI tool only or both a lib and a CLI tool. A CLI is easier to add to an existing crate. A lib probably should have a standalone crate.

crates/aranya-policy-runner/Cargo.toml
```diff
@@ -0,0 +1,42 @@
+[package]
+name = "aranya-policy-runner"
+description = "The Aranya Policy Runner"
```
gknopf-aranya: Is runner or tester a better name? This is referred to as the policy tester in the PR description.

crates/aranya-policy-runner/README.md
```diff
@@ -0,0 +1,75 @@
...
+```
+Usage: policy-tester [OPTIONS] <POLICY> [RUNS]...
+
+Arguments:
+  <POLICY>   The policy file to load
+  [RUNS]...  The run files to load
+
+Options:
+  -w, --working-directory <WORKING_DIRECTORY>
+          The working directory for all data stored
```
gknopf-aranya: Could mention that default is current directory

crates/aranya-policy-runner/README.md
```diff
@@ -0,0 +1,75 @@
...
+          Suppress trace output and other diagnostics
+      --validator
+          Run the validator on the policy compilation (NOT YET IMPLEMENTED)
+  -h, --help
+          Print help
+```
+
+## Examples
+
+A simple policy is included under `examples`, along side several run
```
gknopf-aranya:
```suggestion
A simple policy is included under `examples`, alongside several run
```

crates/aranya-policy-runner/README.md
gknopf-aranya: The examples in this README.md show what happens when tests are run in isolation. Would be helpful to cover the case where a test is run on an existing working directory.

crates/aranya-policy-runner/examples/policy.md
gknopf-aranya: How do we know whether this example policy covers all the edge-cases we want the policy tester to be able to handle?

crates/aranya-policy-runner/src/bin/policy-runner/main.rs
```diff
@@ -0,0 +1,274 @@
...
+
+use aranya_policy_runner::{
+    EchoSink, RunSchedule, create_vmpolicy, get_runfile_preamble_values, load_policy, parse_runfile,
+};
+
+#[derive(Parser, Debug)]
+struct Args {
+    /// The working directory for all data stored
+    #[arg(long, short)]
+    working_directory: Option<String>,
```
gknopf-aranya:
```suggestion
    working_directory: Option<PathBuf>,
```

crates/aranya-policy-runner/src/bin/policy-runner/main.rs
```diff
@@ -0,0 +1,274 @@
...
+// Loads the Device ID from disk or generates a new one
+fn load_device_id(working_directory: &WorkingDirectory) -> anyhow::Result<DeviceId> {
+    let id_path = working_directory.device_id();
+
+    match fs::read(&id_path) {
+        Ok(buf) => {
+            tracing::debug!("loaded Device ID from '{}'", id_path.display());
+            let bytes = buf
+                .try_into()
+                .map_err(|_| anyhow!("Stored Device ID is not exactly 32 bytes"))?;
```
gknopf-aranya: Would be better to get the size of `DeviceId` rather than hardcoding 32 in the error message.

crates/aranya-policy-runner/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+
+[lib]
+
+[dependencies]
+anyhow = { workspace = true }
+aranya-afc-util = { version = "0.18.0", path = "../aranya-afc-util" }
```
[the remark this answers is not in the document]
chip-so: Can you do that in the same workspace where the crates are defined? They aren't dependencies of this workspace, they're just part of it.

crates/aranya-policy-runner/src/bin/policy-runner/main.rs
```diff
@@ -0,0 +1,274 @@
...
+            let head = storage.get_head()?;
+            (storage.get_linear_perspective(head)?, Some(storage))
+        }
+        None => {
+            tracing::debug!("creating new graph");
+            (provider.new_perspective(PolicyId::new(0)), None)
+        }
+    };
+
+    for schedule in run_schedules {
```
gknopf-aranya: Adding a run method on schedule would make this more object oriented and avoid the nested for loops.
```
for schedule in schedules {
    schedule.run()
}
```

crates/aranya-policy-runner/README.md
```diff
@@ -0,0 +1,75 @@
...
+Usage: policy-tester [OPTIONS] <POLICY> [RUNS]...
+
+Arguments:
+  <POLICY>   The policy file to load
+  [RUNS]...  The run files to load
+
+Options:
+  -w, --working-directory <WORKING_DIRECTORY>
+          The working directory for all data stored
+  -d, --deterministic-rng <DETERMINISTIC_RNG>
```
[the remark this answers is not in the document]
chip-so: I'm not really sure what I was thinking here. Thinking about it now, I'm not sure there even needs to be a specified seed. Does anyone have strong opinions about making the seed fixed for this option?

crates/aranya-policy-runner/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+A simple policy is included under `examples`, along side several run
+files which initialize and manipulate the graph. If you are in the
+`aranya-policy-runner` directory, you can replace `policy-runner` with
+`cargo run --`.
+
+### Initialize a graph
+
+```
+$ policy-runner -w wd examples/policy.md examples/init
```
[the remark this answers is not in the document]
chip-so: Yeah, I discussed this with @mooreoak and I begrudgingly agree. I'm thinking `.prun`?

crates/aranya-policy-runner/examples/add_raw_device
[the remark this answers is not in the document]
chip-so: Yes, I have not yet written any tests.

crates/aranya-policy-runner/src/io.rs
```diff
@@ -0,0 +1,93 @@
...
+{
+    type QueryIterator = NullFactIterator;
+
+    fn fact_insert(
+        &mut self,
+        _name: aranya_policy_vm::Identifier,
+        _key: impl IntoIterator<Item = FactKey>,
+        _value: impl IntoIterator<Item = FactValue>,
+    ) -> Result<(), MachineIOError> {
+        error!("Cannot use facts in preamble");
```
gknopf-aranya: Since this policy tester is not intended for production use, it would be acceptable to panic and exit right away if this happens rather than returning an error.

crates/aranya-policy-runner/src/bin/policy-runner/main.rs
```diff
@@ -0,0 +1,274 @@
...
+#[derive(Parser, Debug)]
+struct Args {
+    /// The working directory for all data stored
+    #[arg(long, short)]
+    working_directory: Option<String>,
+    /// Use a deterministic RNG with the given seed (NOT YET IMPLEMENTED)
+    #[arg(long, short)]
+    deterministic_rng: Option<String>,
+    /// Add a marker to the output when execution moves to a new run file
+    #[arg(long)]
```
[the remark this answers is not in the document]
chip-so: I made a judgement call about which options should only be long ones. My rationale is that these are probably less common options and I don't yet want to commit to giving them a short version.

crates/aranya-policy-runner/src/bin/policy-runner/main.rs
```diff
@@ -0,0 +1,274 @@
...
+    marker: bool,
+    /// Suppress trace output and other diagnostics
+    #[arg(long, short)]
+    quiet: bool,
+    /// Run the validator on the policy compilation (NOT YET IMPLEMENTED)
+    #[arg(long)]
```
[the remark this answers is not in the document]
chip-so: [Ditto](https://github.com/aranya-project/aranya-core/pull/549#discussion_r2719011029).

crates/aranya-policy-runner/src/bin/policy-runner/main.rs
```diff
@@ -0,0 +1,274 @@
...
+                "crypto engine secret key loaded from '{}'",
+                key_path.display()
+            );
+            let key = Import::import(buf.as_slice()).context("unable to import key from file")?;
+            Ok(key)
+        }
+        Err(err) if err.kind() == io::ErrorKind::NotFound => {
+            tracing::debug!("generating crypto engine secret key");
+            let key = K::random(&mut Rng);
+            let bytes = key
```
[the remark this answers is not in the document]
chip-so: I'm not worried about it. If anyone uses these secrets in production, something has gone seriously wrong.

crates/aranya-policy-runner/src/runfile.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub do_things: Vec<PolicyRunnable>,
+}
+
+enum Mode {
+    None,
+    Preamble,
+    Do,
+}
+
+pub fn parse_runfile(path: impl AsRef<Path>) -> anyhow::Result<RunFile> {
```
gknopf-aranya: This should probably be part of the `RunFile` impl. E.g. `RunFile::new()`.

crates/aranya-policy-runner/src/bin/policy-runner/main.rs
chip-so: Sure, and I'd have to pass in references to four local variables to do that. Feels like pointless refactoring to me. (And IMNSHO, "more object oriented" is not a useful goal in an of itself)

crates/aranya-policy-runner/src/io.rs
chip-so: I suppose. This use of `error!()` is meant to be consistent with how it's used in the rest of runtime/VM code (e.g. in `VmPolicyIO`).

crates/aranya-policy-runner/src/bin/policy-runner/main.rs
```diff
@@ -0,0 +1,274 @@
...
+use aranya_policy_vm::Identifier;
+use aranya_runtime::{
+    ActionPlacement, GraphId, Policy as _, PolicyId, Sink as _, Storage as _, StorageProvider as _,
+    VmAction, VmPolicy,
+    linear::{LinearStorageProvider, libc::FileManager},
+};
+use clap::Parser;
+
+use aranya_policy_runner::{
+    EchoSink, RunSchedule, create_vmpolicy, get_runfile_preamble_values, load_policy, parse_runfile,
```
gknopf-aranya: It seems like rather than importing these functions manually they could be grouped and imported via a type. If there aren't any cleaner abstractions, `PolicyRunner` is a way to group them together.

crates/aranya-policy-runner/src/bin/policy-runner/main.rs
```diff
@@ -0,0 +1,274 @@
...
+    let working_directory = WorkingDirectory::new(args.working_directory.as_deref().unwrap_or("."));
+    working_directory.make_dirs()?;
+    let keystore = fs_keystore::Store::open(working_directory.keystore())?;
+    let device_id = load_device_id(&working_directory)?;
+
+    let eng = Mutex::new(load_crypto_engine(&working_directory)?);
+    let runfiles: Vec<_> = args
+        .runs
+        .into_iter()
+        .map(|rf| parse_runfile(rf).expect("could not parse run file"))
```
gknopf-aranya: What's the reason for panicking here while returning errors elsewhere?

crates/aranya-policy-runner/src/runfile.rs
```diff
@@ -0,0 +1,95 @@
...
+        let line = line.trim_end();
+        let line_no = i
+            .checked_add(1)
+            .expect("overflow in calculating line number");
+        match line {
+            "preamble:" => mode = Mode::Preamble,
+            "do:" => mode = Mode::Do,
+            l => match mode {
+                Mode::None => {
+                    return Err(anyhow!("Line without 'preamble' or 'do'"));
```
gknopf-aranya: Line number missing from this error

crates/aranya-policy-runner/src/io.rs
```diff
@@ -0,0 +1,93 @@
...
+
+impl Iterator for NullFactIterator {
+    type Item = Result<(Vec<FactKey>, Vec<FactValue>), MachineIOError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        None
+    }
+}
+
+pub struct PreambleIO<'o, CE, KS> {
```
gknopf-aranya: No module or struct docs for `PreambleIO`

crates/aranya-policy-runner/src/runfile.rs
```diff
@@ -0,0 +1,95 @@
...
+                        Ok(expr) => {
+                            match expr.kind {
+                                ExprKind::FunctionCall(_) => {
+                                    do_things.push(PolicyRunnable::Action(full_expression));
+                                }
+                                ExprKind::NamedStruct(_) => {
+                                    do_things.push(PolicyRunnable::Command(full_expression));
+                                }
+                                _ => {
+                                    return Err(anyhow!(
```
gknopf-aranya: Could define an error type for the runfile parser rather than `anyhow!`

crates/aranya-policy-runner/src/policy.rs
```diff
@@ -0,0 +1,154 @@
...
+        CommandContext::Policy(PolicyContext {
+            name: ident!("preamble"),
+            id: Id::default(),
+            author: Id::default(),
+            version: Id::default(),
+        }),
+    );
+    rs.set_pc_by_label(&Label::new(ident!("preamble"), LabelType::Function))?;
+    match rs.run() {
+        Ok(exit_reason) => assert_eq!(exit_reason, ExitReason::Normal),
```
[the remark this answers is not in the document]
chip-so: Yep. `ExitReason` should never be anything other than `Normal`, so this should be impossible to fail. I thought this was more self-documenting than just putting `Ok(_) => {}`. This is more or less at the same level as putting `unreachable!()` on an impossible match case. I'll add a comment to explain this more clearly.

crates/aranya-policy-runner/src/policy.rs
```diff
@@ -0,0 +1,154 @@
...
+
+type CE = DefaultEngine;
+type KS = fs_keystore::Store;
+
+pub struct RunSchedule {
+    pub file_path: PathBuf,
+    pub thunk_ids: Vec<usize>,
+}
+
+pub const FFI_MODULES: [ModuleSchema<'static>; 6] = [
```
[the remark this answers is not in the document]
chip-so: The compiler knows but the runtime doesn't. And we have to define them in the same sequence in both places. So they are all defined.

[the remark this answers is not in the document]
chip-so: The compiler could understand the subset of FFIs needed and encode that into the compiled module, which is validated and loaded by the runtime/VM. This is an enhancement we've thought about for a while now, but we haven't bothered implementing because every Aranya application we've built includes both the compiler and the runtime into the same solution (the embedded project is the exception here, but policy compilation still happens as part of the build process). The only way they could be mismatched is if you edit the list of FFIs badly and recompile. I have mismatched them in development and tracking that down is annoying, but it's not an error you can easily ship because it breaks everything.

crates/aranya-policy-runner/src/runfile.rs
```diff
@@ -0,0 +1,95 @@
...
+                                        "runnable must be an action call or command struct @ line {line_no}: {l}"
+                                    ));
+                                }
+                            }
+                            partial_expression.clear();
+                        }
+                        Err(pe) if pe.kind == ParseErrorKind::Syntax => {
+                            // Not valid syntax, wait for the next line
```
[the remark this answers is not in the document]
chip-so: The parser tries to collect multiple lines into something that _does_ parse correctly, to support items that span multiple lines. We don't know until we reach EOF whether a parse error comes from something badly written, or if it is just not yet complete.

crates/aranya-policy-runner/src/bin/policy-runner/main.rs
```diff
@@ -0,0 +1,274 @@
...
+            globals.push((name, value));
+        }
+    }
+
+    let (machine, run_schedule) = load_policy(&args.policy, globals, runfiles)?;
+
+    let crypto_engine = load_crypto_engine(&working_directory)?;
+    let vm_policy = create_vmpolicy(machine, crypto_engine, keystore, device_id)?;
+
+    execute_schedule(vm_policy, &working_directory, run_schedule, args.marker)?;
```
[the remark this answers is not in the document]
chip-so: This has been fully refactored.

crates/aranya-policy-runner/src/bin/policy-runner/main.rs
```diff
@@ -0,0 +1,274 @@
...
+}
+
+fn main() -> anyhow::Result<()> {
+    let args = Args::parse();
+
+    if !args.quiet {
+        tracing_subscriber::fmt::init();
+    }
+
+    let working_directory = WorkingDirectory::new(args.working_directory.as_deref().unwrap_or("."));
```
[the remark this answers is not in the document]
chip-so: For the record, I disagree that shortening this makes it clearer and I haven't changed it.

crates/aranya-policy-runner/src/bin/policy-runner/main.rs
```diff
@@ -0,0 +1,70 @@
+use std::path::PathBuf;
+
+use anyhow::anyhow;
+use aranya_policy_runner::PolicyRunner;
+use clap::Parser;
+
+/// The Aranya Policy Runner runs sequences of actions on policy files.
```
ssweetney-spok: It runs commands too from what I can tell.

(Looking at crates/aranya-policy-runner/examples/hello.run)

crates/aranya-policy-runner/src/bin/policy-runner/main.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    if args.runs.is_empty() {
+        return Err(anyhow!("Please specify one or more run files"));
+    }
+
+    if !args.quiet {
+        tracing_subscriber::fmt::init();
+    }
+
+    let runner = PolicyRunner::new_from_path(args.policy)?
```
gknopf-aranya: I would usually only call methods that modify a builder if setting it to something other than the default.
```
let builder = PolicyRunnerBuilder::new();
if args.disable_validator {
    builder.disable_validator();
}
...
let validator = builder.build();
```

crates/aranya-policy-runner/src/io.rs
jdygert-spok: It looks like no, since it is compiled within a policy with no facts defined. I'd use either `unreachable!` or `bug!`.

crates/aranya-policy-runner/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+name = "aranya-policy-runner"
```
gknopf-aranya: TODO to publish this once the API is stabilized?

crates/aranya-policy-runner/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+                self.working_directory.base_dir().display()
+            );
+        }
+        self.working_directory
+            .make_dirs()
+            .context("Could not set up working directory")?;
+
+        self.inner_logic()
+            .inspect_err(|e| eprintln!("Execution failed: {e}"))
+            .ok();
```
gknopf-aranya: This is causing the error not to be propagated to the caller.

crates/aranya-policy-runner/src/runfile.rs
```diff
@@ -0,0 +1,445 @@
...
+            Ok(ExitReason::Yield) => unreachable!("Cannot yield in functions"),
+            Err(err) => {
+                return Err(err.into());
+            }
+        }
+        Ok(rs
+            .scope()
+            .locals()
+            .map(|(n, v)| (n.clone(), v.clone()))
+            .collect())
```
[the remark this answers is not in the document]
ssweetney-spok: Wouldn't that still work if we published some placeholder struct to cause a yield?

crates/aranya-policy-runner/src/rng.rs
```diff
@@ -0,0 +1,34 @@
...
+impl SwitchableRng {
+    /// Create a new RNG using [`aranya_crypto::Rng`].
+    pub fn new_default() -> Self {
+        Self::Default
+    }
+
+    /// Create a new RNG using `rand`'s [`StdRng`] implementation (which
+    /// is based on ChaCha12).
+    pub fn new_deterministic() -> Self {
+        Self::Deterministic(RefCell::new(StdRng::from_seed([0u8; 32])))
```
gknopf-aranya:
```suggestion
        Self::Deterministic(RefCell::new(StdRng::from_seed(DEFAULT_SEED)))
```

crates/aranya-policy-runner/src/bin/policy-runner/main.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// The marker includes the run file path.
+    #[arg(long)]
+    marker: bool,
+    /// Suppress trace output and other diagnostics.
+    ///
+    /// This just turns off the default `tracing` subscriber. Fatal
+    /// errors will still be printed.
+    #[arg(long, short)]
+    quiet: bool,
+    /// Run the validator on the policy compilation.
```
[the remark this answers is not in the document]
chip-so: The validator is kind of uselessly noisy at the moment. It's useful, but I wouldn't enable it by default until we can give it some more time in the oven.

crates/aranya-policy-runner/src/runfile.rs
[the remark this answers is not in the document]
chip-so: > I'd consider it a bug that we don't pop from the scope on the outermost return.

I think normally we would, but the VM treats a return with an empty call stack as an exit. It is kind of gross that I'm exploiting that, I'll agree.

> Wouldn't that still work if we published some placeholder struct to cause a yield?

You mean like a sentinel? I think that would work. I could probably also directly modify the produced instruction sequence to turn the return into an exit. I was going to do that until I recalled the function exit behavior (but it's probably also a bad idea).

crates/aranya-policy-runner/src/bin/policy-runner/main.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    if !args.quiet {
+        tracing_subscriber::fmt::init();
+    }
+
+    let runner = PolicyRunner::new_from_path(args.policy)?
+        .with_working_directory(args.working_directory)
+        .with_deterministic_rng(args.deterministic_rng)
+        .with_marker(args.marker)
+        .with_validator(args.validator)
```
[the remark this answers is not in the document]
gknopf-aranya: A builder pattern seems most useful in situations where the defaults are generally used as-is, but the caller needs flexibility to modify some of the attributes of the object without manually specifying all of them. If we're planning to use all the input args, then initializing a struct may make more sense:
```
let t = Type {
    args.opt1,
    args.opt2,
    ...
};
```

crates/aranya-policy-runner/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+Options:
+  -w, --working-directory <WORKING_DIRECTORY>
+          The working directory for all data stored
+  -d, --deterministic-rng <DETERMINISTIC_RNG>
+          Use a deterministic RNG with the given seed (NOT YET IMPLEMENTED)
+      --marker
+          Add a marker to the output when execution moves to a new run file
+  -q, --quiet
+          Suppress trace output and other diagnostics
```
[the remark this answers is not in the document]
chip-so: > I would probably always output to stdout for the effects and stderr for logging, and just expect the caller to redirect if they want either in a file.

I think that's a reasonable approach (I certainly like the Unix-ness of it). But a `-o` option is quite common.

crates/aranya-policy-runner/src/bin/policy-runner/main.rs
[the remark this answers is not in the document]
chip-so: I think if you were using it programmatically, you would often leave most of these at their defaults. It's just that the command-line tool's argument parser _also_ understands these options and has its own defaults so they're all specified.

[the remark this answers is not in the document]
chip-so: I think the better solution here is to have a configuration object that uses the builder pattern but in the CLI we also implement something like `From<Args>`.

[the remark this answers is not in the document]
chip-so: Well... every field in `PolicyRunner` is a configuration option, so it frankly seems kind of pointless to have a separate configuration object. I'm gonna leave it.

crates/aranya-policy-runner/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+
+        if self.working_directory.is_temporary {
+            // Clean up temporary working directory
+            tracing::debug!(
+                "Removing temporary working directory {}",
+                self.working_directory.base_dir().display()
+            );
+            self.working_directory.delete()?;
+        }
```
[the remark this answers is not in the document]
chip-so: So there's a subtle but important difference with `Drop` - it removes the ability to propagate errors caused by the deletion process. Which are not likely, but I think they're important to surface. I'll just leave a note.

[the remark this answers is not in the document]
chip-so: Well, I'll implement `Drop` as well for panic handling. But I'll keep the existing logic to surface errors in the normal case.

crates/aranya-policy-runner/src/policy.rs
```diff
@@ -0,0 +1,142 @@
...
+
+/// A `RunSchedule` keeps track of which action thunks are associated
+/// with which file, so they can be run in sequence.
+pub struct RunSchedule<'a> {
+    pub file_path: &'a Path,
+    pub preamble_values: Vec<Value>,
+    pub thunk_range: Range<usize>,
+}
+
+// NOTE(chip): It is important that these are the same FFIs in the same
```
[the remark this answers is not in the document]
chip-so: Shoving this off into a later task: #652.

crates/aranya-policy-runner/README.md
```diff
@@ -0,0 +1,104 @@
...
+repeatable feedback.
+
+## Usage
+
+```text
+policy-runner [OPTIONS] <POLICY> [RUNS]...
+```
+
+At its most basic, you must specify a policy document and one or more
+["run files"](#run-files). The run files specify the actions and command
```
apetkov-so: This should be "commands" (plural), no?

crates/aranya-policy-runner/README.md
```diff
@@ -0,0 +1,104 @@
...
+that are run on the policy. As effects are committed, they are printed
```
apetkov-so: Did you mean emitted?

crates/aranya-policy-vm/src/machine.rs
```diff
@@ -402,6 +406,11 @@ where
         self.pc
     }
 
+    /// Access the scope manager to read or change defined variables.
+    pub fn scope(&mut self) -> &ScopeManager<'_> {
```
jdygert-spok:
```suggestion
    /// Access the scope manager to read defined variables.
    pub fn scope(&self) -> &ScopeManager<'_> {
```

crates/aranya-policy-runner/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Fail to retrieve a device
+
+This example tries to retrieve a device with a random ID. Unless you're
+exceedingly lucky, this device shouldn't be found and it produces an
+error effect.
+
+```text
+$ policy-runner -w wd examples/policy.md examples/get_raw_device_not_found.run
+DeviceNotFound { device_id: BRJpYqJh3ZSdCtmB4XkXQvPCwQdANmDaeASvgGtTxk9X }
```
jdygert-spok, line 84: I just get `Error: panic` when I run this.

crates/aranya-policy-runner/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+aranya-policy-compiler = { version = "0.22.0", path = "../aranya-policy-compiler" }
+aranya-policy-lang = { version = "0.14.1", path = "../aranya-policy-lang" }
+aranya-policy-vm = { version = "0.21.0", path = "../aranya-policy-vm" }
+aranya-runtime = { version = "0.22.0", path = "../aranya-runtime", features = ["std", "libc"] }
+clap = { version = "4.5", features = ["derive"] }
+postcard = { workspace = true }
+rand = { workspace = true, features = ["std_rng"] }
+thiserror = { workspace = true }
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true, features = ["fmt", "env-filter"] }
```
jdygert-spok: nit: separate local deps

```suggestion
aranya-afc-util = { version = "0.23.0", path = "../aranya-afc-util" }
aranya-crypto = { version = "0.14.0", path = "../aranya-crypto", features = ["fs-keystore", "memstore"] }
aranya-crypto-ffi = { version = "0.21.0", path = "../aranya-crypto-ffi" }
aranya-device-ffi = { version = "0.21.0", path = "../aranya-device-ffi" }
aranya-envelope-ffi = { version = "0.21.0", path = "../aranya-envelope-ffi" }
aranya-idam-ffi = { version = "0.21.0", path = "../aranya-idam-ffi" }
aranya-perspective-ffi = { version = "0.21.0", path = "../aranya-perspective-ffi" }
aranya-policy-compiler = { version = "0.22.0", path = "../aranya-policy-compiler" }
aranya-policy-lang = { version = "0.14.1", path = "../aranya-policy-lang" }
aranya-policy-vm = { version = "0.21.0", path = "../aranya-policy-vm" }
aranya-runtime = { version = "0.22.0", path = "../aranya-runtime", features = ["std", "libc"] }

anyhow = { workspace = true }
clap = { version = "4.5", features = ["derive"] }
postcard = { workspace = true }
rand = { workspace = true, features = ["std_rng"] }
thiserror = { workspace = true }
tracing = { workspace = true }
tracing-subscriber = { workspace = true, features = ["fmt", "env-filter"] }
```

crates/aranya-policy-runner/README.md
jdygert-spok: Same for `hello.run` and `get_raw_device.run`

crates/aranya-policy-runner/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+/// | Option | Meaning |
+/// |--------|---------|
+/// | `with_working_directory(Option<PathBuf>)` | Set a working directory (`Some`) or use a temporary directory (`None`) |
+/// | `with_deterministic_rng(bool)` | Use a deterministic RNG implementation. |
+/// | `with_marker(bool)` | Print a marker between run files that shows which run file is producing the effects. |
+/// | `with_validator(bool)` | Run the compiler validator before execution. |
+#[derive(Debug, Clone)]
+#[must_use]
+pub struct PolicyRunner {
+    working_directory: WorkingDirectory,
```
jdygert-spok: This doesn't seem to need a `Clone` impl, and it seems a bit dubious since you'll clone the `WorkingDirectory` which gets wiped when you drop one of them.
Could just remove this impl plus the `impl Clone for OutputDestination`.

crates/aranya-policy-runner/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+    }
+}
+
+/// A `RunSchedule` keeps track of which action thunks are associated
+/// with which file, so they can be run in sequence.
+struct RunSchedule<'a> {
+    pub file_path: &'a Path,
+    pub preamble_values: Vec<Value>,
+    pub thunk_range: Range<usize>,
+}
```
jdygert-spok, line 51: nit: pub fields but they aren't actually exposed.

crates/aranya-policy-runner/README.md
chip-so: Yep. :/

crates/aranya-policy-runner/README.md
chip-so: Yep. :/

crates/aranya-policy-runner/README.md
chip-so: Works for me. Did you run the init command in the first example?

jdygert-spok: doh. That's what I get for not reading.

crates/aranya-policy-runner/src/lib.rs
chip-so: The goal was to allow a runner to be "forked" by cloning it, so that a base configuration could be set, then customized. But you're right about the `WorkingDirectory` and the utility of cloning it is questionable at best. I'll remove it.

PR #550, opened 2026-01-22
chore: update to rust 1.90

jdygert-spok:


gknopf-aranya (approved):

mooreoak (approved):
Were these changes due to updates to clippy?

jdygert-spok:
Yes, the main change was to use the newly stabilized if-let chains (with a clippy lint telling us to do it). There are also some new methods and lints around alignment/multiple checks.

PR #558, opened 2026-02-09
runtime: remove refcell usage for MachineIO

jdygert-spok:


crates/aranya-runtime/src/vm_policy/io.rs
```diff
@@ -49,21 +45,16 @@ where
 
 /// Implements the `MachineIO` interface for [VmPolicy](super::VmPolicy).
 pub struct VmPolicyIO<'o, P, S, CE, FFI> {
-    facts: &'o RefCell<&'o mut P>,
-    sink: &'o RefCell<&'o mut S>,
-    engine: &'o Mutex<CE>,
-    ffis: &'o [FFI],
+    pub facts: &'o mut P,
```
apetkov-so: Why did these become public?

PR #563, opened 2026-02-13
crypto: use shared rng and make engine shared

jdygert-spok:
Update to `spideroak-crypto@0.7.0` for aranya-project/crypto#67 and make the crypto engine shared too.

### DIFF
--- a/crates/aranya-fast-channels/src/mutex.rs
+++ b/crates/aranya-fast-channels/src/mutex.rs
@@ -11,12 +11,11 @@ use core::{
     sync::atomic::{AtomicU32, Ordering},
 };
 
-use cfg_if::cfg_if;
-
 use crate::util::const_assert;
 
 // A mutex that can NOT be used in shared memory.
-cfg_if! {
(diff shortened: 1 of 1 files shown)


crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -194,7 +193,7 @@ macro_rules! vm_effect {
 /// A [Policy] implementation that uses the Policy VM.
 pub struct VmPolicy<CE> {
     machine: Machine,
-    engine: Mutex<CE>,
+    engine: CE,
```
jdygert-spok: No mutex needed on crypto engine inside `VmPolicy`.

jdygert-spok:
This PR is almost all find-replace that results from changing the signatures. I commented on the notable changes above.

jdygert-spok:
I want to add blanket impls like `Engine for &E where E: Engine`, but I'm not sure what to do about `fn destroy(self)`. Nothing implements or uses it though, so... could remove it.

PR #566, opened 2026-02-17
suggestions: cache visited segments

gknopf-aranya:
- Fixes some typos
- Check visited before push
- Refactors similar code

### DIFF
--- a/crates/aranya-runtime/src/storage/mod.rs
+++ b/crates/aranya-runtime/src/storage/mod.rs
@@ -306,44 +306,9 @@ pub trait Storage {
 
         let (visited, queue) = buffers.get();
         push_queue(queue, start)?;
-
-        while let Some(loc) = queue.pop() {
-            debug_assert!(
-                loc.max_cut >= address.max_cut,
-                "Invariant: we only enqueue locations with at least the target max cut"
-            );
-
-            if !visited.visit(loc.segment) {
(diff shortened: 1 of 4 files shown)


crates/aranya-runtime/src/storage/mod.rs
```diff
@@ -416,45 +381,76 @@ pub trait Storage {
...
+/// `Ok(None)` if the queue is exhausted.
+///
+/// See `aranya-docs/docs/graph-traversal.md` for the traversal algorithm specification.
+fn traverse<S, T>(
+    storage: &S,
+    target_max_cut: MaxCut,
+    visited: &mut TraversalVisited,
+    queue: &mut TraversalQueue,
+    mut check: impl FnMut(&S::Segment) -> Result<Option<T>, StorageError>,
+) -> Result<Option<T>, StorageError>
```
jdygert-spok: I'm hesitant to abstract this out when we only have 2 uses (see rule of 3).

gknopf-aranya: It's a lot of lines of code though that should stay in sync. And if a 3rd case is added, will it be refactored? For 2 cases of 2-5 lines of repeated code it seems okay not to refactor. But, this code staying in sync partially helps to ensure correctness of the algorithm.

PR #581, opened 2026-02-22
Critical bug fix: Ensure RawSecret::Seed zeroes out sensitive bytes

Sajjon:
# Summary
This PR hardens secret handling in` aranya-crypto` by ensuring temporary secret material is zeroized when dropped, and by removing an unprotected raw-seed representation from `RawSecret`.

If we add the unit test below on `main` it fails, but it should not - i.e. this unit test demonstrates that `RawSecret::Seed` does not zeroize the seed bytes. This PR introduces a fix for this critical bug and an updated version of this unit test now passes.

<details>
<summary>Test of `RawSecret::Seed` zeroize - CLICK ME</summary>

```rust
use core::{mem::MaybeUninit, ptr};

use aranya_crypto::{default::DefaultCipherSuite, engine::RawSecret};

#[test]
#[ignore = "disabled pending seed zeroization behavior change"]
fn raw_secret_seed_should_zeroize_on_drop() {
    const SECRET: [u8; 64] = [0xA5; 64];

    let mut slot = MaybeUninit::new(RawSecret::<DefaultCipherSuite>::Seed(SECRET));
    let secret = slot.as_mut_ptr();

    // Capture a pointer directly to the seed field bytes.
    let seed = unsafe {
        match &mut *secret {
            RawSecret::Seed(seed) => seed as *mut [u8; 64],
            _ => unreachable!("constructed as `RawSecret::Seed`"),
        }
    };

    // Drop the enum in place and inspect the same memory.
    //
    // SAFETY:
    // - `secret` points to initialized `RawSecret` storage in `slot`.
    // - `drop_in_place(secret)` drops exactly once.
    // - `seed` points into the same storage, and we only read bytes.
    let after_drop = unsafe {
        ptr::drop_in_place(secret);
        seed.read()
    };

    // This assertion currently FAILS on main, demonstrating that
    // `RawSecret::Seed([u8; 64])` is not scrubbed on drop.
    assert_eq!(after_drop, [0u8; 64]);
}
```

</details>

## Changes
- Make `Ciphertext` impl `ZeroizeOnDrop`
- Added new `RawSeed([u8; 64])` type in [engine.rs](crates/aranya-crypto/src/engine.rs).
- RawSeed implements:
  * ZeroizeOnDrop
   * Drop that zeroizes the seed bytes
   * conversions `From<[u8; 64]>` and `From<RawSeed> for [u8; 64]`
- Changed `RawSecret::Seed` from `[u8; 64]` to `RawSeed`.

## Compatibility / API impact
This is a low-level API change:
`RawSecret::Seed([u8; 64])` is now `RawSecret::Seed(RawSeed)`.

### Migration examples:
Constructing: `RawSecret::Seed(seed.into())`
Extracting bytes: `let bytes: [u8; 64] = raw_seed.into();`


crates/aranya-crypto/src/engine.rs
```diff
@@ -135,6 +136,68 @@ pub trait RawSecretWrap<E: Engine> {
...
+    }
+
+    /// Returns the seed bytes, consuming `self`.
+    ///
+    /// This method replaces the internal storage with zeros before `self` is
+    /// dropped, so the original in-struct buffer is not left containing the seed.
+    ///
+    /// # Security
+    ///
+    /// The returned `[u8; 64]` is still secret material and is **not**
```
gknopf-aranya: If the goal is to make sure this data is always zeroized on drop, then it would be ideal if the API only returns types that are automatically zeroized without any reliance on the caller to zeroize correctly.

PR #584, opened 2026-02-23
policy-module: split off `ConstValue` and move `Value` to the vm

jdygert-spok:
`ConstValue` holds only the values which can be constructed at policy compile time and are stored in the module.

### DIFF
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -324,13 +324,14 @@ impl<'a> CompileState<'a> {
                     }
                 }
                 StructItem::StructRef(field_type_ident) => {
-                    let other =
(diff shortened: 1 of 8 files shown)


jdygert-spok:
> This looks fine, but I'm not sure I understand the need for it. Is this so that the global values can be easier to export through the ifgen?

Yep, it helps with that. In general it statically guarantees that we can't put a fact, id, or identifier in a global or an `Instruction::Const`, making them also no longer part of the serialized module format.

chip-so (dismissed):

apetkov-so (dismissed):

ssweetney-spok (dismissed):
This should also make [const expressions](https://github.com/aranya-project/aranya-core/issues/589) easier to implement I believe

PR #593, opened 2026-02-26
policy-ifgen: generate constants for globals

### LINKED ISSUE #582: Expose policy globals in policy-ifgen generated interface
## Summary

Update `policy-ifgen` to include policy globals (top-level `let` bindings) in the generated Rust interface, so downstream code can reference authoritative values instead of duplicating them.

## Motivation

The policy defines default role ranks as globals:

```policy
let admin_role_rank = 800
let operator_role_rank = 700
let member_role_rank = 600
```

These values are currently duplicated as Rust constants in multiple test files in `aranya` (annotated with `TODO(aranya-core#582)`). This is fragile — if the policy values change, the constants must be updated manually with no compile-time guarantee they stay in sync.

## Proposed Approach

Per discussion on [#577](https://github.com/aranya-project/aranya-core/issues/577), the globals are already compiled into the `globals` map in `Machine`/`Module`. The suggestion is to have `policy-ifgen` generate Rust constants from these compiled globals so they're available at compile time in the generated interface.

## Related

- #577 (closed) — original issue for exporting const values from the policy interface

## Current Workaround

Rank constants are manually redefined in downstream test code with `TODO(aranya-core#582)` comments.

jdygert-spok:
Closes #582.

### DIFF
--- a/crates/aranya-policy-compiler/src/compile.rs
+++ b/crates/aranya-policy-compiler/src/compile.rs
@@ -1936,6 +1936,8 @@ impl<'a> CompileState<'a> {
                 Some(e) => Some(Box::new(self.expression_value(e)?)),
                 None => None,
             })),
+            ExprKind::Ok(e) => Ok(ConstValue::Result(Ok(Box::new(self.expression_value(e)?)))),
+            ExprKind::Err(e) => Ok(ConstValue::Result(Err(Box::new(self.expression_value(e)?)))),
             _ => Err(self.err(CompileErrorType::InvalidExpression(e.clone()))),
         }
     }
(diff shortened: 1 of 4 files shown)


crates/aranya-policy-ifgen/tests/golden.rs
```diff
@@ -32,3 +32,8 @@ fn tictactoe() {
 fn structs() {
     dotest("structs");
 }
+
+#[test]
+fn constants() {
+    dotest("constants");
```
gknopf-aranya: Constants or globals? Can there be constants there are not global, or globals that are not constants?

chip-so (approved):

PR #595, opened 2026-02-26
aranya-crypto: remove bearssl re-export and feature

jdygert-spok:
The `aranya-crypto` bearssl feature just re-exports `spideroak_crypto::bearssl`. If we or someone else wants to use it for something we can still just use it from `spideroak_crypto`. Removing it here means that `aranya-core` CI doesn't need to build it and run tests for it.

### DIFF
--- a/crates/aranya-crypto/src/ciphersuite/mod.rs
+++ b/crates/aranya-crypto/src/ciphersuite/mod.rs
@@ -100,57 +100,6 @@ pub trait CipherSuite {
 
 #[cfg(test)]
 mod tests {
-    #[cfg(feature = "bearssl")]
-    mod bearssl {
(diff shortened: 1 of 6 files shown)


crates/aranya-crypto/src/ciphersuite/mod.rs
```diff
@@ -100,57 +100,6 @@ pub trait CipherSuite {
...
-            P384,
-        >);
-        test_ciphersuite!(p521, TestCs<
-            Aes256Gcm,
-            Sha256,
-            HkdfSha512,
-            DhKemP521HkdfSha512,
-            HmacSha512,
-            P521,
-        >);
```
jdygert-spok: TODO: `test_ciphersuite` is just a wrapper around tests from `spideroak_crypto`. Ensure those tests are called for all these types in `spideroak_crypto`.

PR #599, opened 2026-03-03
policy-compiler: Improve compiler error testing

ssweetney-spok:
Preliminary work for https://github.com/aranya-project/aranya-core/issues/572

The same as https://github.com/aranya-project/aranya-core/pull/489 but for the compiler crate

### DIFF
--- a/crates/aranya-policy-compiler/tests/data/results/result_err_not_defined.fail.snap
+++ b/crates/aranya-policy-compiler/tests/data/results/result_err_not_defined.fail.snap
@@ -1,8 +1,0 @@
----
-source: crates/aranya-policy-compiler/tests/tests.rs
-expression: error
----
-not defined: struct Nonexistent at line 1 col 1:
-	function f() result[int, struct Nonexistent] {
-    return Ok(1)
-}
--- a/crates/aranya-policy-compiler/tests/data/results/result_ok_not_defined.fail.snap
+++ b/crates/aranya-policy-compiler/tests/data/results/result_ok_not_defined.fail.snap
@@ -1,8 +1,0 @@
----
-source: crates/aranya-policy-compiler/tests/tests.rs
-expression: error
----
-not defined: struct Nonexistent at line 1 col 1:
-	function f() result[struct Nonexistent, int] {
-    return Err(1)
-}


ssweetney-spok:
The snapshots of the `Module`s for each "passing" case is making the diff bigger than it needs to be. I could omit it 🤔

crates/aranya-policy-compiler/tests/tests.rs
```diff
@@ -0,0 +1,138 @@
...
+        let ModuleData::V0(ModuleV0 {
+            labels,
+            action_defs,
+            command_defs,
+            fact_defs,
+            struct_defs,
+            enum_defs,
+            globals,
+            ..
+        }) = &self.0.data;
```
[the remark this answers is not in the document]
jdygert-spok: IMO it could be nice to have a large sample of compiled instructions to see how our changes affect the instructions.

crates/aranya-policy-compiler/tests/data/facts/update.pass.policy
```diff
@@ -0,0 +1,11 @@
+fact Foo[i int] => {a string}
+command Test {
+    seal { return todo() }
+    open { return todo() }
+    policy {
+        finish {
+            update Foo[i: 1] to {a: "updated"}
+            update Foo[i: 1]=>{a:"apple"} to {a: "orange"}
```
chip-so: I'm not sure about this test. This shouldn't work, but it's currently [defined as a runtime exception to change the same fact twice in a finish block](https://aranya-project.github.io/policy-book/reference/top-level/commands.html#finish-block). It's technically valid according to the compiler, but I don't know what this test actually proves.

crates/aranya-policy-compiler/tests/data/functions/arg_option_struct_not_defined.fail.policy
chip-so: I'm not sure that this proves anything the next test doesn't.

Cargo.toml
```diff
@@ -130,6 +130,10 @@ proc-macro2 = "1.0.80"
 syn = "2.0.59"
 quote = "1.0"
 
+insta = { version = "1", features = ["redactions", "yaml"] }
+rstest = { version = "0.26" }
+serde_json = { version = "1" }
```
jdygert-spok: `serde_json` is unused for the existing parser tests and the new compiler tests. The insta features are unused I think.

crates/aranya-policy-compiler/tests/tests.rs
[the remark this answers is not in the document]
ssweetney-spok: I'll bring back the instructions and keep them in the same file then.

crates/aranya-policy-compiler/tests/tests.rs
```diff
@@ -0,0 +1,205 @@
...
+            }
+        }
+
+        writeln!(f, "\nstruct_defs:")?;
+        for (name, fields) in struct_defs {
+            writeln!(f, "\t{name}:")?;
+            for field in fields {
+                writeln!(f, "\t\t{}: {}", field.identifier, field.field_type)?;
+            }
+        }
```
ssweetney-spok: Going to update this. Just noticed that field-less structs look odd

PR #603, opened 2026-03-04
clean up lifetimes

jdygert-spok:
Targeting #524. Many of the lifetimes were more restrictive than they needed to be.

### DIFF
--- a/crates/aranya-policy-lang/src/lang/parse.rs
+++ b/crates/aranya-policy-lang/src/lang/parse.rs
@@ -1731,9 +1731,9 @@ fn parse_policy_chunk(
     parse_policy_chunk_inner(chunk, &p, policy).map_err(|e| e.with_source(policy.text.clone()))
 }
 
-fn parse_policy_chunk_inner<'a>(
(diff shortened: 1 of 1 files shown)


crates/aranya-policy-lang/src/lang/parse.rs
```diff
@@ -35,13 +35,13 @@ type FieldsAndSources = (Vec<(Ident, Expression)>, Vec<Ident>);
...
-    span: Span<'a>,
-    to_ast_span: Box<dyn Fn(Span<'a>) -> Result<ast::Span, ParseError> + 'b>,
+struct PairContext<'i, 'p> {
+    pairs: RefCell<Pairs<'i, Rule>>,
+    span: Span<'i>,
+    to_ast_span: Box<dyn Fn(Span<'i>) -> Result<ast::Span, ParseError> + 'p>,
 }
 
-impl<'a, 'b> PairContext<'a, 'b> {
+impl<'i> PairContext<'i, '_> {
```
gknopf-aranya: Why call the lifetime `i` instead of `a` here?

Issue #530, opened 2026-01-07
policy-vm: test nested option handling


jdygert-spok:
Test that wrapping and unwrapping nested optionals works properly, and we can correctly distinguish e.g. `None` and `Some(None)`.

PR #632, opened 2026-03-25
policy-compiler: Improve error messages

### LINKED ISSUE #572: policy-compiler: improve error messages

ssweetney-spok:
resolves #572

old:
```
already defined: f at line 5 col 1:
	function f() int {}
```

new:
```
error: the name `f` is defined multiple times
  |
1 | function f(x int) int {
  |          - previous defintion here
...
5 | function f() int {}
  |          ^ re-defined here
```


crates/aranya-policy-compiler/tests/tests.rs
jdygert-spok, line 1: revert

crates/aranya-policy-compiler/src/compile/lower.rs
```diff
@@ -300,11 +347,12 @@ impl CompileState<'_> {
...
-                    return Err(self.err(CompileErrorType::InvalidType(
-                        "Cannot wrap option in another option".into(),
-                    )));
-                }
+                // We allow nested optional types.
+                // if matches!(inner_vtype.kind, TypeKind::Optional(_)) {
+                //     return Err(self.err(CompileErrorType::InvalidType(
+                //         "Cannot wrap option in another option".into(),
+                //     )));
+                // }
```
jdygert-spok, line 355: Oh, I had allowed this in the parser but I guess didn't allow it in the compiler. We should add a test for nested option in the compiler and/or vm. Just now remembering #530...

crates/aranya-policy-ast/src/tests.rs
```diff
@@ -108,15 +109,25 @@ fn test_result_type_matches() {
...
+    );
     let struct2 = named_struct!(
         "MyStruct",
         100,
         108,
         "field",
         109,
         114,
-        ExprKind::Int(42),
+        ExprKind::Int(IntLiteral::new(42, Span::empty())),
```
chip-so, line 130: Could we clean this up by making a `From<i64> for IntLiteral` so you can do `42.into()`?

crates/aranya-policy-compiler/src/compile/error.rs
```diff
@@ -1,164 +1,260 @@
...
+            expected,
+            expected_span: expected_span.filter(|s| !s.is_empty()),
+            found_type,
+            found_expr,
+        }
+    }
+}
+
+/// Invalid call color - Tried to make a function call to the wrong type of function.
+pub(crate) struct InvalidCallColor(pub InvalidCallColorKind, pub Span, pub Option<Span>);
```
chip-so, line 64: All of these tuple structs would be fine if they had a common structure, but it makes the implementations in `rendering.rs` kind of hard to follow. And besides, it feels stylistically inconsistent with some being tuples and other being regular (though I would be fine with a simple convention like, it can be a tuple if it only contains a `Span`).

crates/aranya-policy-compiler/src/compile/lower.rs
```diff
@@ -680,28 +755,24 @@ impl CompileState<'_> {
                 let t = self.lower_expression(t)?;
 
                 let name = t.vtype.as_struct().ok_or_else(|| {
-                    self.err(CompileErrorType::InvalidType(
-                        "Expression left of `.` is not a struct".into(),
+                    self.err(InvalidType::new(
+                        "struct T".to_owned(),
```
chip-so, line 759: `struct T` looks weird to me. `T` implies a generic inner type, but struct isn't generic. That's just the name of it (and if you had a struct named T this would be _very_ confusing). I'd just say "struct".

crates/aranya-policy-compiler/src/compile/types.rs
```diff
@@ -141,6 +144,14 @@ impl IdentifierTypeStack {
     }
 }
 
+pub(crate) struct IdentNotDefined;
```
chip-so, line 147: Why is this defined here and not with the other errors?

crates/aranya-policy-compiler/src/compile/types.rs
```diff
@@ -98,7 +101,7 @@ impl IdentifierTypeStack {
     /// Retrieve a type for an identifier. Searches lower stack items if a mapping is not
     /// found in the current scope.
     #[allow(clippy::result_large_err)]
```
chip-so, line 103: Seems like this can be removed.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -550,12 +551,8 @@ impl<'a> CompileState<'a> {
...
 
-    fn err(&self, err_type: CompileErrorType) -> CompileError {
-        self.err_loc(err_type, self.last_span)
-    }
-
-    fn err_loc(&self, err_type: CompileErrorType, span: Span) -> CompileError {
-        CompileError::from_span(err_type, span, self.m.codemap.as_ref())
+    fn err(&self, err_type: impl error::rendering::Error) -> CompileError {
+        CompileError::new(err_type, Some(self.policy.text.clone()))
     }
```
chip-so, line 556: So this clones the entire source text? I guess that's fine in this day and age, but it feels very wrong. 😅

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1960,16 +1947,30 @@ impl<'a> CompileState<'a> {
                         }
                         _ => None,
                     })
-                    .ok_or_else(|| self.err(CompileErrorType::InvalidExpression(e.clone()))),
-                _ => Err(self.err(CompileErrorType::InvalidExpression(e.clone()))),
+                    .ok_or_else(|| {
+                        // TODO(Steve): The policy book doesn't mention what a const context is. We should probably update the book to describe compile-time evaluation.
```
chip-so, line 1951: The book describes this context and its limitations in [Global Values](https://aranya-project.github.io/policy-book/reference/top-level/global-values.html), so I'd call it a "global value" context. There is also attributes, which follow the same rules, but it describes them as the same as global `let`.

(But I do think the TODO for describing this more fully in the book is a good one)

chip-so (commented):
Looked through all the code files. This is looking very good overall.

crates/aranya-policy-compiler/tests/tests.rs
jdygert-spok: reverted

crates/aranya-policy-ast/src/tests.rs
jdygert-spok: I have a PR to handle this generally by adding a `WithSpan` wrapper to replace each of the structs that just add a span, and conversion methods. #665

crates/aranya-policy-compiler/src/compile/lower.rs
apetkov-so: changed

crates/aranya-policy-compiler/src/compile/lower.rs
apetkov-so: Compiler already allows nested options. Removed comment and added tests.

crates/aranya-policy-compiler/src/compile/types.rs
apetkov-so: We could move it... Or just use the existing `NotDefined` error.

Edit: Actually, lower.rs:723 already maps it to NotDefined. This error (and UnifyTypeError) are generated from things that don't have spans.

crates/aranya-policy-ast/src/ast.rs
```diff
@@ -538,14 +570,42 @@ pub struct ForeignFunctionCall {
 }
 }
 
+/// A 64-bit signed integer
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct IntLiteral {
+    value: i64,
+    span: Span,
+}
```
jdygert-spok: Now we can use `type IntLiteral = WithSpan<i64>;`

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -122,21 +124,34 @@ macro_rules! typekind {
...
-            name: ident!("this").nowhere(),
-            ty: TypeKind::Struct(name).nowhere(),
+            name: Ident {
+                inner: ident!("this"),
+                span: Span::empty(),
+            },
+            ty: VType {
+                inner: TypeKind::Struct(name),
+                span: Span::empty(),
+            },
```
jdygert-spok: revert

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1145,17 +1163,25 @@ impl<'a> CompileState<'a> {
...
 
         self.identifier_types
-            .add_global(identifier.inner.clone(), vt.nowhere())
+            .add_global(
+                identifier.clone(),
+                VType {
+                    inner: vt,
+                    span: Span::default(),
+                },
+            )
```
jdygert-spok: More to revert here and below

jdygert-spok:
There's some lints to fix but otherwise I think this is good enough to ship and make any improvements in followups.

supply-chain/imports.lock
jdygert-spok, line 1: revert formatting regressions

jdygert-spok (approved):

crates/aranya-policy-compiler/src/compile/types.rs
```diff
@@ -242,15 +249,15 @@ pub(super) fn unify_pair(left: VType, right: VType) -> Result<VType, TypeUnifyEr
...
             })
         }
         (TypeKind::Result(left_result), TypeKind::Result(right_result)) => {
             let ok = unify_pair(left_result.ok.clone(), right_result.ok.clone())?;
             let err = unify_pair(left_result.err.clone(), right_result.err.clone())?;
             Ok(VType {
-                inner: TypeKind::Result(Box::new(aranya_policy_ast::ResultTypeKind { ok, err })),
-                span: aranya_policy_ast::Span::empty(), // TODO
+                inner: TypeKind::Result(Box::new(ResultTypeKind { ok, err })),
+                span: Span::empty(), // TODO
```
djohnson-spideroak, line 260: still TODO?

crates/aranya-policy-compiler/src/compile/lower.rs
```diff
@@ -1606,119 +1720,111 @@ impl CompileState<'_> {
...
                     thir::StmtKind::Create(thir::CreateStatement { fact })
                 }
-                (StmtKind::Update(s), StatementContext::Finish) => {
+                (StmtKind::Update(s), StatementContext::Finish(_)) => {
                     // ensure fact is mutable
                     let fact_def = self.get_fact_def(&s.fact.identifier)?;
                     if fact_def.immutable {
-                        return Err(
-                            self.err(CompileErrorType::Unknown(String::from("fact is immutable")))
+                        // TODO(Steve): This should probably be a new error type
```
djohnson-spideroak, line 1739: lets have an issue for this, or a general issue if there are other types we might want in the future.

djohnson-spideroak (approved):
Please make sure we are tracking the TODOs added in this PR

apetkov-so:
> Please make sure we are tracking the TODOs added in this PR

#681

crates/aranya-policy-compiler/src/compile/lower.rs
apetkov-so: #681

chip-so (approved):

crates/aranya-policy-compiler/src/compile.rs
apetkov-so: Maybe the err method can take a span and clone just a slice of the policy.

crates/aranya-policy-compiler/src/compile/error.rs
jdygert-spok: #682

crates/aranya-policy-compiler/src/compile.rs
jdygert-spok: #683

PR #633, opened 2026-03-26
sync: Split Graph Diffing Algorithms Into Module

nikki-aranya:
This splits the two main algorithms in the syncer (sampling commands in the requester, and finding missing segments in the responder) into their own separate `diff` module.

Changes also include beginning the migration to const generics on those new functions, and getting rid of the alloc::vec instances so it's properly no_alloc and using `TraversalBuffers` directly.


nikki-aranya:
Unsure what to do about the aranya-embedded CI check, I haven't touched core in a while

jdygert-spok:
Thanks for the update. Opened PR for that to be reviewed #645.

PR #636, opened 2026-03-31
Create a new public-core crate

### LINKED ISSUE #624: storage: stabilize LinearStorageProvider as the default storage implementation
The goal is to expose an API for using LinearStorageProvider while also still providing a way to create and use custom IO managers.

### LINKED ISSUE #628: core: public API crate
A new crate will be created to hold the ClientState, VmPolicy, LinearStorageProvider, and the sync protocol API. This crate will be the base of the public core API.

ssweetney-spok:
Closes #628
Closes #624 
Closes #644


chip-so (dismissed):

djohnson-spideroak (dismissed):
What do you think about eventually moving the minimal-core stuff into examples for this crate?

crates/aranya-public-core/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "aranya-public-core"
+description = "Stable public API for the Aranya runtime"
```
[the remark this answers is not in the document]
djohnson-spideroak: If thats available, I think we should go with that.

gknopf-aranya (dismissed):

crates/aranya-core/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+aranya-core = { path = ".", features = ["std", "libc"] }
+aranya-crypto = { version = "0.14.0", features = ["std", "alloc", "getrandom", "memstore"] }
+aranya-crypto-ffi = "*"
+aranya-device-ffi = "*"
+aranya-envelope-ffi = "*"
+aranya-idam-ffi = "*"
+aranya-perspective-ffi = "*"
+aranya-policy-compiler = "*"
+aranya-policy-lang = "*"
+aranya-policy-vm = "*"
```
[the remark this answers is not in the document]
djohnson-spideroak: I think we would have to skip that for aranya-crypto, it cant be defined as a path and cratesio version.

edit: I think its fine actually, I misread the Cargo.toml


```
     error: failed to load manifest for workspace member `/workspace/crates/aranya-core`                                                                                 
     referenced via `crates/*` by workspace at `/workspace/Cargo.toml`                                                                                                   
                                                                                                                                                                         
     Caused by:                                                                                                                                                          
       failed to parse manifest at `/workspace/crates/aranya-core/Cargo.toml`                                                                                            
                                                                                                                                                                         
     Caused by:                                                                                                                                                          
       Dependency 'aranya-crypto' has different source paths depending on the build target. Each dependency must have a single canonical source path irrespective of     
     build target.                                                                                                                                                       

```

crates/aranya-core/src/lib.rs
jdygert-spok, line 1: I would like to make the API less "flat", e.g. not re-exporting everything at top level but add more modules like maybe storage and policy.

I'd consider using inline modules here since they are simple, although the modules might be a good place for documentation which could make them larger.

crates/aranya-core/src/client.rs
```diff
@@ -0,0 +1,173 @@
...
+/// The Aranya client.
+///
+/// Wraps [`ClientState`] with the policy store fixed to [`VmPolicyStore`]
+/// and the storage provider fixed to [`LinearStorageProvider`].
+///
+/// - `CE` is the cryptographic engine (e.g., [`aranya_crypto::default::DefaultEngine`]).
+/// - `FM` is the IO manager for storage (e.g., [`aranya_runtime::linear::IoManager`]).
+pub struct Client<CE, FM: IoManager> {
+    inner: ClientState<VmPolicyStore<CE>, LinearStorageProvider<FM>>,
+}
```
jdygert-spok: Can this not just be a type alias like the others?

djohnson-spideroak: I didnt make this decision, so I can only go based on the code, but from what I understand:

A type alias would not show a concrete signature in the docs and it would bring the (currently hidden) PolicyStore impl for VmPolicyStore into the docs. 

The new struct also prevents newly added client functions from immediately becoming part of the public API, and provides more strict bounds on the types (IoManager vs StorageProvider).

I will defer to you on this though, if we want to try a type alias I can update the PR.

crates/aranya-core/examples/simple.rs
```diff
@@ -0,0 +1,380 @@
...
+use aranya_perspective_ffi::FfiPerspective as PerspectiveFfi;
+use aranya_policy_compiler::Compiler;
+use aranya_policy_lang::lang::parse_policy_document;
+use aranya_policy_vm::{Machine, Struct, Value, ffi::FfiModule as _, ident};
+
+// ---------------------------------------------------------------------------
+// Type Aliases
+// ---------------------------------------------------------------------------
+
+type CE = DefaultEngine<Rng, DefaultCipherSuite>;
```
chip-so: Use the `CS` alias here?

crates/aranya-core/src/client.rs
chip-so: FWIW, I thought ClientState was intended to be the public interface. But I don't know for sure, either.

crates/aranya-core/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+aranya-crypto = { path = "../aranya-crypto", features = ["std", "alloc", "getrandom", "memstore"] }
+aranya-crypto-ffi = { path = "../aranya-crypto-ffi" }
+aranya-device-ffi = { path = "../aranya-device-ffi" }
+aranya-envelope-ffi = { path = "../aranya-envelope-ffi" }
+aranya-idam-ffi = { path = "../aranya-idam-ffi" }
+aranya-perspective-ffi = { path = "../aranya-perspective-ffi" }
+aranya-policy-compiler = { path = "../aranya-policy-compiler" }
+aranya-policy-lang = { path = "../aranya-policy-lang" }
+aranya-policy-vm = { path = "../aranya-policy-vm" }
+postcard = { workspace = true, features = ["alloc"] }
```
chip-so: nit: Separate aranya deps and other deps here?

crates/aranya-core/src/client.rs
djohnson-spideroak: I think I am going to keep the wrapper, iirc we did decide that we wanted more strict storage bounds on this version. In the future, we could expose the raw ClientState if we need.

crates/aranya-core/examples/simple.rs
djohnson-spideroak: As in, rename the type, or use an existing alias? I thought CS was CipherSuite, with CE for the engine.

crates/aranya-core/src/client.rs
jdygert-spok: > provides more strict bounds on the types (IoManager vs StorageProvider).

Isn't that what e.g. `ClientTransaction` is also doing?

crates/aranya-core/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(not(any(test, doctest, feature = "std")), no_std)]
+
+mod client;
+
+#[doc(inline)]
+pub use aranya_runtime::{
+    Address, ClientError, CmdId, Command, GraphId, Session, Sink, Transaction, TraversalBuffer,
+    TraversalBuffers,
+};
```
jdygert-spok, line 26: Don't need transaction or session since we have a type alias for those.

crates/aranya-core/src/client.rs
djohnson-spideroak: What would be the point of the relaxed bounds on the client if we are also restricting the bounds on the ClientTransaction & related? I am guessing you would just include `LinearStorage<...>` in the type bounds for the less restrictive generics?

Either way, we decided we didn't want to spend the time trying to stabilize the storage interface for custom replacements of linear storage, but we still wanted to expose the IOManager. Let me know if you think we should just go with the relaxed bounds.

jdygert-spok: I don't see what makes this any less strict:

```rust
pub type Client<CE, FM: IoManager> = ClientState<VmPolicyStore<CE>, LinearStorageProvider<FM>>;
```

I don't think it's super important, but it just seemed weird that this is the only one we wrap.

djohnson-spideroak: What is your opinion? I can change it to just be a type alias, but we do lose the ability to selectively expose functions via the wrapper. If it isnt going to cause issues, and we dont have a reason to prefer one of the other, I think we can just keep it.

crates/aranya-core/examples/simple.rs
chip-so: As in, the next line defines the `CS` alias, so why not put it first and use it here?

jdygert-spok (approved):

crates/aranya-core/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+
+mod client;
+
+#[doc(inline)]
+pub use aranya_runtime::{
+    Address, ClientError, ClientState, CmdId, Command, GraphId, Session, Sink, Transaction,
+    TraversalBuffer, TraversalBuffers,
+};
+#[doc(inline)]
+pub use client::VmPolicyStore;
```
chip-so, line 28: I don't understand why this is exported in two places - here and in `policy` below.

chip-so (approved):

crates/aranya-core/src/lib.rs
djohnson-spideroak: I didnt catch that, it should probably just be client. I can resolve this in my followup PR for the engine.

PR #642, opened 2026-04-01
641 multiple recalls

### LINKED ISSUE #641: Support multiple recall blocks
- Commands should allow multiple recall blocks.
- Recall blocks can be named.
- Recall blocks can accept arguments.
- Extend `check` syntax to allow specifying recall block, e.g. `check ... or recall foo(1, "fail")`.
- This syntax is only allowed in `policy` blocks.
- Default unnamed block still allowed. It does not accept arguments.
- Default `check` statement still allowed, and continues to work in all contexts.
- Similar recall syntax for `check_upwrap`.

Discussion here: https://github.com/aranya-project/aranya-core/discussions/433

apetkov-so:
- Policies can define more than one recall block, and they can accept arguments.
- All blocks must be named.
- Added `recall` statement that invokes the named block.
- `check` no longer exits; it now has a required `else` clause with a terminal expression (e.g. `recall` or `return`)


apetkov-so:
We haven't extended the check_unwrap operator so far. Unlike the check statement, check_unwrap is a prefix operator, and we can't just stick `or recall` after it, e.g. `let x = check_unwrap ... or recall foo()`. Maybe we can change it to a postfix operator, but that's a breaking change. 🤔

apetkov-so:
I don't like all the `else todo()`s... We should probably land the unit type, action results first.

jdygert-spok:
Can't we continue to allow bare `check <cond>` in actions for now?

apetkov-so:
> Can't we continue to allow bare `check <cond>` in actions for now?

Not after we made the else clause mandatory. We should probably land actions-return-results first, then undo all those `else todo()`s. Edit: I mean replace them with `return Err(...)`.

Still no good way to check in open/seal though.

jdygert-spok:
We can keep the else mandatory in policy context, but just allow bare check in actions for now. Check in actions doesn't recall anyways, so it doesn't matter that there's no default recall block.

apetkov-so:
1. I think we should require the else clause in command policies. Otherwise we'd exit without invoking a recall since there is no longer a default recall block. This is because check failure compiles to `Exit(ExitReason::Check(None))`
2. And `check_unwrap` also exits with reason `Check(None)`, meaning no more default recall.

This needs more speccing.

crates/aranya-policy-compiler/src/compile/lower.rs
```diff
@@ -1042,6 +1046,53 @@ impl CompileState<'_> {
...
+    fn lower_recall_call(
+        &mut self,
+        fc: &FunctionCall,
+        cmd: &aranya_policy_ast::CommandDefinition,
+    ) -> Result<thir::FunctionCall, CompileError> {
+        // Find the matching recall block in the command definition
+        let recall_block = cmd
+            .recalls
+            .iter()
+            .find(|rb| rb.identifier.as_str() == fc.identifier.as_str())
```
chip-so: I guess it's the same but I'd do `identifier.inner` instead of `.as_str()` to make it more clear.

crates/aranya-policy-compiler/src/compile/lower.rs
```diff
@@ -1475,7 +1526,27 @@ impl CompileState<'_> {
...
-                    thir::StmtKind::Check(thir::CheckStatement { expression: et })
+
+                    // Lower else expression
+                    let else_expression = match &s.else_expression {
+                        Some(expr) => {
+                            let lowered = self.lower_expression(expr)?;
+                            if !matches!(lowered.vtype.inner, TypeKind::Never) {
+                                return Err(self.err_loc(
+                                    CompileErrorType::InvalidType(String::from(
+                                        "check else clause must be terminal (type Never)",
```
chip-so: I've got mixed feelings about describing "never" in an error message. I'm not sure if it would be nicer to the programmer to describe the (kind of type theoretically obtuse) never type in our documentation, or describe exactly the types of expressions which are valid (`return`, `recall`, `todo()`).

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -883,14 +901,19 @@ impl<'a> CompileState<'a> {
...
-                // branch to if the check succeeds is the
-                // instruction after that - current instruction + 2.
-                let next = self.wp.checked_add(2).assume("self.wp + 2 must not wrap")?;
-                self.append_instruction(Instruction::Branch(Target::Resolved(next)));
-                self.append_instruction(Instruction::Exit(ExitReason::Check));
+
+                // Reserve space for the branch instruction. We'll go back and update its target
+                // after we know where the success path begins.
+                let branch_wp = self.wp;
+                self.append_instruction(Instruction::Branch(Target::Resolved(0)));
```
chip-so: You should just insert a `Target::Unresolved` with an anonymous label. This kind of unknown target is why that mechanism exists.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1853,13 +1952,11 @@ impl<'a> CompileState<'a> {
 
         for function_def in &self.policy.functions {
             self.compile_function(function_def)?;
-            self.exit_statement_context();
```
chip-so, line 1857: Wait, so was this just getting called extra times and we didn't notice?

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1206,27 +1248,84 @@ impl<'a> CompileState<'a> {
...
 
+    /// Returns a unique recall block name for the given command and recall name.
+    /// The name follows the format: `<command>_recall_<name>`.
+    fn command_recall_name(
+        &self,
+        command: &ast::CommandDefinition,
+        recall_name: &Ident,
+    ) -> Result<Identifier, CompileError> {
+        format!(
+            "{}_recall_{}",
```
chip-so: I think the `_recall_` part here is redundant, since this should only show up in `LabelType::Recall` labels.

chip-so (commented):
Overall this looks really good.

crates/aranya-policy-compiler/src/compile.rs
apetkov-so: Yeah. 😳 `pop` on empty `statement_context` vec simply returns `None`; no panic. We never noticed because nothing reads the statement stack between function compilations, and each one (`compile_function`, `compile_finish_function`, `compile_action`) did their own context enter/exit logic. And this stack is only at compile-time, it doesn't affect the machine's stack.

chip-so (dismissed):

crates/aranya-policy-ast/src/ast.rs
```diff
@@ -28,6 +28,13 @@ where
     }
 }
 
+impl Spanned for Text {
+    fn span(&self) -> Span {
+        // TODO: we should add spans for text literals
+        Span::empty()
+    }
+}
+
```
jdygert-spok: Unused?

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1206,27 +1246,80 @@ impl<'a> CompileState<'a> {
         Ok(())
     }
 
+    /// Returns a unique recall block name for the given command and recall name.
+    /// The name follows the format: `<command>_<name>`.
+    fn command_recall_name(
```
jdygert-spok, line 1253: This is probably fine for now, but this breaks for a case like this:

```
command a { recall b_c() {} }
command a_b { recall c() {} }
```

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1188,6 +1212,22 @@ impl<'a> CompileState<'a> {
...
+        let context = self.get_statement_context()?.clone();
+        let n = match &context {
+            StatementContext::CommandPolicy(cmd) | StatementContext::CommandRecall(cmd) => {
+                Some(self.command_recall_name(cmd, &recall.identifier)?)
+            }
+            _ => None,
+        };
+        self.append_instruction(Instruction::Exit(ExitReason::Check(n)));
+        Ok(())
+    }
```
jdygert-spok: Let's store the command name in `thir::StmtKind::Recall` so we don't have to look in the context here.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1206,27 +1246,80 @@ impl<'a> CompileState<'a> {
...
+            // define args, like compile_function
+            for arg in recall_block.arguments.iter().rev() {
+                self.ensure_type_is_defined(&arg.ty)?;
+                self.append_var(arg.name.inner.clone(), arg.ty.clone())?;
+            }
+
+            self.compile_statements(&recall_block.statements, Scope::Same)?;
+            self.identifier_types.exit_function();
+            self.exit_statement_context();
+            self.append_instruction(Instruction::Exit(ExitReason::Normal));
```
jdygert-spok:
```suggestion
            let params = recall_block
                .arguments
                .iter()
                .cloned()
                .chain([param::this(command.identifier.clone()), param::envelope()])
                .collect::<Vec<_>>();

            self.enter_statement_context(StatementContext::CommandRecall(command.clone()));
            self.compile_function_like(
                &params,
                None,
                recall_block.span,
                &recall_block.statements,
                Label::new(full_name, LabelType::CommandRecall),
            )?;
            self.append_instruction(Instruction::Exit(ExitReason::Normal));
            self.exit_statement_context();
```

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1188,6 +1212,22 @@ impl<'a> CompileState<'a> {
         Ok(())
     }
 
+    fn compile_recall(&mut self, recall: thir::FunctionCall) -> Result<(), CompileError> {
```
jdygert-spok: It's interesting to consider an alternative way we could do the recall:

```
exit check
jump <recall 'Foo_bar'>
```

Then instead of calling into the vm by supplying the recall name from the check failure, you just resume the vm and it would then jump to the recall. That also means we could just make these temporary labels since they would only be internal.

crates/aranya-policy-lang/src/lang/parse/snapshots/aranya_policy_lang__lang__parse__tests__parse_policy_test.snap
jdygert-spok, line 1: extra file?

crates/aranya-policy-compiler/src/compile.rs
apetkov-so: Do you mean we could call other commands' recalls? That shouldn't be possible, because the recall statement takes the unmangled block name.

jdygert-spok: We would try to define the `a_b_c` label twice, causing a compile error.

crates/aranya-policy-compiler/src/compile.rs
jdygert-spok: Based on our rust call today we could just call the recall directly without exiting since we don't need to yield to the caller since we decided to always call the recall regardless of whether we are at the origin perspective.

crates/aranya-policy-ast/src/ast.rs
apetkov-so: removed

crates/aranya-policy-lang/src/lang/parse/snapshots/aranya_policy_lang__lang__parse__tests__parse_policy_test.snap
apetkov-so: removed

crates/aranya-policy-compiler/src/compile.rs
apetkov-so: Ok, #678.

crates/aranya-policy-compiler/src/compile.rs
apetkov-so: Added #679.

chip-so (dismissed):

chip-so (dismissed):

chip-so (dismissed):

crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -395,15 +400,21 @@ impl<CE: aranya_crypto::Engine> VmPolicy<CE> {
...
         this_data: Struct,
         envelope: Envelope<'_>,
+        recall_block_name: Identifier,
     ) -> Result<(), PolicyError>
     where
         M: MachineIO<MachineStack>,
     {
-        match rs.call_command_recall(this_data, envelope.into()) {
+        // NOTE Can recall actually fail?
+        match rs.call_command_recall(this_data, envelope.into(), recall_block_name) {
```
djohnson-spideroak: is this question resolved?

crates/aranya-policy-compiler/src/compile/lower.rs
```diff
@@ -1124,6 +1132,49 @@ impl CompileState<'_> {
...
+                self.err(NotDefined(note, fc.identifier.span))
+            })?;
+
+        let arg_defs = recall_block.arguments.as_slice();
+
+        let mut arguments = Vec::new();
+        for (i, (param, arg_e)) in arg_defs.iter().zip(fc.arguments.iter()).enumerate() {
+            let arg_te = self.lower_expression(arg_e)?;
+            if !arg_te.vtype.fits_type(&param.ty) {
+                let _ = i; // index unused beyond optional context
```
jdygert-spok: nit: remove `enumerate` ?

crates/aranya-policy-compiler/src/tests.rs
jdygert-spok, line 1: Is there a reason not to make these file tests?

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -197,3 +209,139 @@ fn test_validate_publish() {
...
+            seal { return todo() }
+            open { return todo() }
+            policy {
+                finish {}
+            }
+            recall default() {
+            }
+            recall foo() {
+            }
+        }"#,
```
jdygert-spok: `default` isn't special anymore.

crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -197,3 +209,139 @@ fn test_validate_publish() {
...
+
+#[test]
+fn test_check_statement() {
+    let text = r#"
+        action test_action() {
+            // Basic check statement
+            check true
+            
+            // Check with else clause
+            check true
```
jdygert-spok: Doesn't use `else` clause.

crates/aranya-policy-compiler/tests/data/commands/recall_duplicate_unnamed_already_defined.fail.snap
```diff
@@ -0,0 +1,9 @@
+---
+source: crates/aranya-policy-compiler/tests/tests.rs
+expression: error
+---
+error: unknown error
+   |
+10 | /     recall {
+11 | |     }
+   | |_____^ recall block 'default' for command 'Cmd' defined more than once
```
jdygert-spok: nit: this should be `AlreadyDefined` error.

crates/aranya-policy-compiler/tests/data/commands/recall_in_recall_invalid_context.fail.snap
```diff
@@ -0,0 +1,8 @@
+---
+source: crates/aranya-policy-compiler/tests/tests.rs
+expression: error
+---
+error: unknown error
+  |
+9 |         check true else recall foo()
+  |                         ^^^^^^^^^^^^ `recall` is only valid in command `policy` blocks
```
jdygert-spok: nit: this should not be unknown error. Maybe `InvalidExpression`?

crates/aranya-policy-compiler/tests/data/commands/recall_or_recall_invalid_context.fail.snap
jdygert-spok, line 1: no input file for this

crates/aranya-policy-compiler/tests/data/commands/recall_duplicate_unnamed_already_defined.fail.snap
jdygert-spok, line 1: no input file for this.

crates/aranya-policy-compiler/src/tests.rs
jdygert-spok: I think they might all be already covered by the added file tests.

crates/aranya-policy-vm/tests/vm.rs
jdygert-spok, line 1: Looks like there are some tests to update here. Or move to another PR?

crates/aranya-runtime/src/vm_policy.rs
apetkov-so: Removed the comment. Recalls can indeed panic or check-exit, e.g. as a result of `unwrap`, `todo`, or `check`, which are all allowed in check contexts.

crates/aranya-policy-compiler/tests/data/commands/recall_duplicate_unnamed_already_defined.fail.snap
apetkov-so: Removed. We don't have unnamed recalls anymore.

crates/aranya-policy-compiler/tests/data/commands/recall_or_recall_invalid_context.fail.snap
apetkov-so: Removed. The `or recall` clause was replaced with the `else` clause.

crates/aranya-policy-compiler/src/tests.rs
apetkov-so: All cases are already covered by other tests. Removed.

crates/aranya-policy-compiler/src/tests.rs
apetkov-so: Added else clause, and converted to policy tests

crates/aranya-policy-compiler/src/tests.rs
apetkov-so: All recall tests are now policy files.

chip-so (dismissed):

crates/aranya-policy-vm/tests/vm.rs
```diff
@@ -1566,6 +1616,99 @@ fn test_coalesce_or() -> anyhow::Result<()> {
     Ok(())
 }
 
+// Closes https://github.com/aranya-project/aranya-core/issues/530:
+// "Test that wrapping and unwrapping nested optionals works properly, and
+// we can correctly distinguish e.g. `None` and `Some(None)`."
+#[test]
+#[ignore = "blocked on aranya-project/aranya-core#632 lifting nested-option restriction"]
```
jdygert-spok: @apetkov-so what's the plan for these 3 ignored tests?

apetkov-so: Oops, that was for a different ticket. I'll remove it. But what do you mean by "3", there's only one?
Edit: Oh, the "rebase WIP" tests. I'll see if we still need them.

jdygert-spok (approved):

chip-so (approved):

PR #653, opened 2026-04-09
Braid convergence

benz-spideroak:
### Summary

- Replace the per-strand `is_ancestor` BFS (O(k) per location, where k = graph depth) with an incremental convergence map that identifies convergence points via a single BFS pre-pass, reducing ancestor detection to O(1) amortized
- Bound all braid data structures (`StrandHeap`, `BraidResult`, `ConvergenceMap`) with fixed-capacity `heapless` collections, eliminating unbounded heap allocation during braiding
- Add disk-backed overflow for both `BraidResult` and `ConvergenceMap` via a new `TempFile` abstraction, so large graphs spill to disk instead of OOMing
- Improve the test DSL with a `ConvergeAll` rule that syncs every client pair until convergence, replacing the previous fixed-round approach, and add `sync_client_zero` option to `GenerateGraph`

### Details

**Convergence map** (`convergence_map.rs`, new file): Walks backwards from both merge parents via BFS to find locations reachable from multiple paths. Keeps up to 3 blocks of ~170 entries in memory with LRU eviction to a temp file. During braiding, each prior location is checked against this map in O(1) instead of running a full BFS.

**Bounded braid result** (`braiding.rs`): Replaces `Vec<Location>` with a fixed-capacity `heapless::Vec` that spills to disk in 4KB pages when full. Provides reverse iteration across both in-memory and spilled entries.

**TempFile** (`storage/temp_file.rs`, new file): Platform-abstracted temporary file with `libc` (unlinked file in `/tmp`) and `testing` (in-memory `RefCell<Vec<u8>>`) backends.

**TraversalQueue extensions** (`storage/mod.rs`): Added `push_duplicate` and `pop_duplicates` for the convergence BFS, and new error variants `StrandHeapOverflow` and `ConvergenceRootOverflow`.


crates/aranya-runtime/src/storage/temp_file.rs
```diff
@@ -0,0 +1,135 @@
...
+    fn new() -> Result<Self, StorageError> {
+        use aranya_libc::{
+            self as libc, O_CLOEXEC, O_CREAT, O_DIRECTORY, O_EXCL, O_RDONLY, O_RDWR, Path, S_IRUSR,
+            S_IWUSR,
+        };
+
+        static COUNTER: core::sync::atomic::AtomicU64 = core::sync::atomic::AtomicU64::new(0);
+        let id = COUNTER.fetch_add(1, core::sync::atomic::Ordering::Relaxed);
+
+        let tmp_path = Path::new(b"/tmp\0");
```
chip-so, line 31: Is there a reason we don't want to use `std::env::temp_dir()`?

crates/aranya-runtime/src/storage/temp_file.rs
```diff
@@ -0,0 +1,135 @@
...
+            }
+        }
+        Ok(())
+    }
+}
+
+// --- testing (in-memory) backend ---
+
+#[cfg(all(feature = "testing", not(feature = "libc")))]
+pub struct TempFile {
```
chip-so, line 102: I find it a little weird that `TempFile` is two completely different implementations depending on feature flags (as opposed to e.g. `MemStore` vs `fs_keystore::Store` implementing `KeyStore`). I suppose there's nothing that prevents another implementation from being chosen at run time.

crates/aranya-runtime/src/client/convergence_map.rs
```diff
@@ -0,0 +1,394 @@
...
+/// Number of in-memory blocks.
+const NUM_BLOCKS: usize = 3;
+/// Maximum entries in the root index (3 × BLOCK_ENTRIES).
+/// Each root entry points to one spilled block, so this supports
+/// up to 510 × 170 = 86,700 convergence points before overflow.
+const ROOT_CAPACITY: usize = BLOCK_ENTRIES * 3;
+/// Size of one entry on disk: 3 × u64 = 24 bytes.
+const ENTRY_BYTES: usize = 24;
+/// Size of one block on disk.
+const BLOCK_BYTES: usize = BLOCK_ENTRIES * ENTRY_BYTES;
```
chip-so, line 19: I think the intent here is this fits within a 4K (4096 byte) block, but it's not really stated. And the fact that it's close but not exact (4080 bytes) means that if that's important, it probably needs some padding?

All of this feels rather arbitrary, though, and it would be great to have some explanation for the numbers chosen. And for things like `ENTRY_BYTES`, I think you could just use `size_of::<Entry>()` (or at least `size_of::<u64>() * 3`)?

crates/aranya-runtime/src/client/convergence_map.rs
```diff
@@ -0,0 +1,394 @@
...
+        buf[8..16].copy_from_slice(&(self.location.max_cut.0 as u64).to_le_bytes());
+        buf[16..24].copy_from_slice(&(self.count as u64).to_le_bytes());
+        buf
+    }
+
+    #[allow(clippy::unwrap_used)] // infallible: slices are exactly 8 bytes
+    fn from_bytes(buf: &[u8; ENTRY_BYTES]) -> Self {
+        let segment = u64::from_le_bytes(buf[0..8].try_into().unwrap()) as usize;
+        let max_cut = u64::from_le_bytes(buf[8..16].try_into().unwrap()) as usize;
+        let count = u64::from_le_bytes(buf[16..24].try_into().unwrap()) as usize;
```
chip-so, line 41: I'm assuming all of this is for serializing to disk, so it's probably fine to use `to`/`from_ne_bytes()`.

crates/aranya-runtime/src/storage/mod.rs
```diff
@@ -145,6 +183,50 @@ impl TraversalQueue {
         Ok(self.entries.swap_remove(self.partition))
     }
 
+    /// Pop the entry with the highest `max_cut`, removing all entries
+    /// at that exact location. Returns `(location, count)`.
```
chip-so, line 187: I think this doc comment was meant for the following `pop_duplicates()`.

crates/aranya-runtime/src/storage/mod.rs
```diff
@@ -145,6 +183,50 @@ impl TraversalQueue {
...
+        self.entries.iter().max_by_key(|loc| *loc)
+    }
+
+    pub fn pop_duplicates(&mut self) -> Result<Option<(Location, usize)>, StorageError> {
+        // Find the entry with the highest max_cut.
+        let Some((i, _)) = self.entries.iter().enumerate().max_by_key(|&(_, loc)| *loc) else {
+            return Ok(None);
+        };
+
+        let location = self.entries[i];
```
chip-so, line 202: It's not clear to me why the index `i` was found if the only use for it is to retrieve the entry. Seems like you could just do `let Some(location) = self.entries.iter().max_by_key(|loc| *loc) else { ... }`.

chip-so (commented):
While looking at the disk spill approach, it occurred to me that this is an ideal case for `rkyv` - excess entries are written to disk and read but not modified afterwards. If this could just serialize chunks of data into rkyv'd blocks, maybe some of this looks a lot simpler. Not suggesting a change, just food for thought.

crates/aranya-runtime/src/storage/mod.rs
benz-spideroak: Fixed

crates/aranya-runtime/src/storage/mod.rs
benz-spideroak: Fixed.

crates/aranya-runtime/src/client/convergence_map.rs
benz-spideroak: Fixed.

crates/aranya-runtime/src/client/convergence_map.rs
benz-spideroak: I changed it to `size_of::<u64>() * 3` and fixed the comment. My thought was to make sure it fit in 4K block even on a 64 bit system. I'm not sure how important that actually is. What do you think?

chip-so: I'd say it's somewhat important, if you're likely to have a lot of blocks (where "a lot" is like > 1000). It's unlikely the OS is going to read individual 4K sectors unless you ask it nicely, but it will likely be reading chunks with a granularity of 4K since that's both a common memory page size and a common disk sector size. So it's more about alignment than size. And in that respect, being close to but not quite 4096 is subtly sub-optimal because you will have some blocks that cross those 4K boundaries and potentially require a little more disk IO. The effect is probably subtle, and you'd probably have to bench it on a memory-constrained system to even see it. Personally, I'd write it in whatever size makes sense to you, and leave the micro-optimizations for someone else's particular use case.

chip-so (dismissed):

crates/aranya-runtime/src/storage/temp_file.rs
benz-spideroak: I updated this to use a trait and match `LinearStorage`.

crates/aranya-runtime/src/storage/spill.rs
```diff
@@ -0,0 +1,135 @@
...
+    fn new() -> Result<Self, StorageError> {
+        use aranya_libc::{
+            self as libc, O_CLOEXEC, O_CREAT, O_DIRECTORY, O_EXCL, O_RDONLY, O_RDWR, Path, S_IRUSR,
+            S_IWUSR,
+        };
+
+        static COUNTER: core::sync::atomic::AtomicU64 = core::sync::atomic::AtomicU64::new(0);
+        let id = COUNTER.fetch_add(1, core::sync::atomic::Ordering::Relaxed);
+
+        let tmp_path = Path::new(b"/tmp\0");
```
benz-spideroak: I don't think we can since we're no_std and aranya libc doesn't have a temp_dir.

benz-spideroak: I updated the way this works. Now there's a single file and we write the chunks within it.

chip-so: Then maybe it could be conditional on the `std` feature? Or just use the `TMPDIR` env var if it's available? It's probably unimportant in practice, but macOS, for example, provides a per-user temp directory in `TMPDIR` that `std::env::temp_dir()` uses. I'm not really sure why they prefer you'd use it, but clearly they do.

chip-so: Ah, I see, it's not picking a dir itself anymore. Sounds good.

crates/aranya-runtime/src/client/braiding.rs
```diff
@@ -182,32 +370,32 @@ mod strand_heap {
 
     /// A wrapper around a binary heap which is limited to one finalize command.
     pub struct StrandHeap<S> {
-        heap: BinaryHeap<Strand<S>>,
+        heap: heapless::BinaryHeap<Strand<S>, Max, STRAND_CAPACITY>,
```
jdygert-spok, line 362: This is over 100KB on the stack.

crates/aranya-runtime/src/client/braiding.rs
```diff
@@ -53,19 +47,198 @@ pub(super) fn last_common_ancestor<S: Storage>(
...
+
+        for loc in &self.mem {
+            let mut buf = [0u8; LOCATION_BYTES];
+            buf[0..8].copy_from_slice(&(loc.segment.0 as u64).to_ne_bytes());
+            buf[8..16].copy_from_slice(&(loc.max_cut.0 as u64).to_ne_bytes());
+            self.spill_file.write_at(offset, &buf)?;
+            offset = offset
+                .checked_add(LOCATION_BYTES)
+                .assume("spill offset must not overflow")?;
+        }
```
jdygert-spok: With `zerocopy` (to hide away the unsafe), we can just read and write `&[Location]` directly.

crates/aranya-runtime/src/client/braiding.rs
```diff
@@ -53,19 +47,198 @@ pub(super) fn last_common_ancestor<S: Storage>(
...
+    }
+}
+
+/// Iterator over braid results in forward order (reversed from push order).
+///
+/// Reads the in-memory buffer backwards first, then reads spilled
+/// blocks from disk in reverse.
+///
+/// Does not implement [`Iterator`] because iteration is fallible
+/// (disk reads may fail). Use [`next`](Self::next) directly.
```
jdygert-spok: The usual pattern for this is to make the iterator item a `Result`.

crates/aranya-runtime/src/client/braiding.rs
```diff
@@ -53,19 +47,198 @@ pub(super) fn last_common_ancestor<S: Storage>(
...
+    }
+
+    /// Get the next location in forward braid order.
+    pub fn next(&mut self) -> Result<Option<Location>, ClientError> {
+        // First yield from in-memory buffer (backwards).
+        if self.mem_pos > 0 {
+            self.mem_pos = self
+                .mem_pos
+                .checked_sub(1)
+                .assume("mem_pos > 0 checked above")?;
```
jdygert-spok: nit:

```suggestion
        if let Some(prev_mem_pos) = self.mem_pos.checked_sub(1) {
            self.mem_pos = prev_mem_pos;
```

Same below

crates/aranya-runtime/src/storage/spill.rs
```diff
@@ -0,0 +1,153 @@
...
+
+use crate::{StorageError, storage::Spill};
+
+// --- libc backend ---
+
+/// File-backed spill. Created unlinked inside a caller-supplied directory
+/// and cleaned up when the last clone of the handle is dropped.
+#[cfg(feature = "libc")]
+pub struct LibcSpill {
+    fd: alloc::sync::Arc<aranya_libc::OwnedFd>,
```
jdygert-spok: Unused arc?

crates/aranya-runtime/src/storage/mod.rs
```diff
@@ -15,6 +15,26 @@ use crate::{Address, CmdId, Command, PolicyId, Prior};
...
+
+/// Byte-addressable overflow storage for braid and convergence data.
+///
+/// Implemented by [`LibcSpill`] (file-backed) and [`MemSpill`] (in-memory);
+/// each backend has its own constructor signature (paths, etc.), matching
+/// how [`IoManager`](linear::io::IoManager) backends are constructed.
+/// Callers build a spill and pass it in.
+pub trait Spill {
+    /// Write `data` at the given byte offset.
+    fn write_at(&self, offset: usize, data: &[u8]) -> Result<(), StorageError>;
```
jdygert-spok: Let's make this `&mut self`.

crates/aranya-runtime/src/client/braiding.rs
```diff
@@ -251,22 +217,28 @@ impl<'a, F: ScratchFile> BraidIter<'a, F> {
...
     right: Location,
     lca: Location,
     buffer: &mut TraversalBuffer,
-) -> Result<BraidResult<F>, ClientError> {
+    make_spill: &MS,
+) -> Result<BraidResult<F>, ClientError>
+where
+    S: Storage,
+    F: Spill,
+    MS: Fn() -> Result<F, StorageError>,
```
chip-so, line 214: I'm curious what led you to this pattern instead of just taking an `impl Spill`. Lazy construction?

chip-so (dismissed):

crates/aranya-runtime/src/client/convergence_map.rs
```diff
@@ -0,0 +1,388 @@
...
+/// Incrementally-computed convergence map with disk-backed overflow.
+///
+/// Keeps up to 3 blocks of 256 entries in memory. When a block
+/// fills, the least-recently-accessed block is spilled to a temp
+/// file. An in-memory root index maps max_cut ranges to file
+/// offsets for O(1) block lookup.
+pub struct ConvergenceMap<'a, F> {
+    blocks: [Block; NUM_BLOCKS],
+    active_block: usize,
+    root: heapless::Vec<NodeEntry, ROOT_CAPACITY>,
```
jdygert-spok: This is also fairly large, ~32KB on the stack.

crates/aranya-runtime/src/client/braiding.rs
jdygert-spok: `BraidResult` and `ConvergenceMap` both need one at the same time.

crates/aranya-runtime/src/client/braiding.rs
benz-spideroak: This is now passed in.

crates/aranya-runtime/src/client/braiding.rs
benz-spideroak: Changed to an iterator.

crates/aranya-runtime/src/storage/spill.rs
benz-spideroak: Removed

crates/aranya-runtime/src/storage/mod.rs
benz-spideroak: Changed

crates/aranya-runtime/src/client/convergence_map.rs
benz-spideroak: This is now passed in.

crates/aranya-runtime/src/client/braiding.rs
benz-spideroak: Used zerocopy

crates/aranya-runtime/src/client/convergence_map.rs
```diff
@@ -386,3 +420,49 @@ impl<'a, F: Spill> ConvergenceMap<'a, F> {
...
+                block.last_accessed, 0,
+                "last_accessed must be reset to 0 on reuse"
+            );
+        }
+
+        // root must also be cleared.
+        assert!(cs2.root.is_empty());
+    }
+
+    /// Verify that `Block::clear` resets `last_accessed` (the A.2.1 fix).
```
chip-so: I don't understand what A.2.1 refers to here.

chip-so (dismissed):

chip-so (dismissed):

crates/aranya-runtime/src/client/buffers.rs
```diff
@@ -0,0 +1,69 @@
...
+/// field alongside this struct.
+///
+/// Internal graph-mutation helpers that require a singular
+/// [`TraversalBuffer`](crate::storage::TraversalBuffer) receive
+/// `&mut buffers.traversal.primary`.
+///
+/// Construct once per long-lived component and reuse across calls.
+pub struct RuntimeBuffers<S> {
+    pub traversal: TraversalBuffers,
+    pub braid: BraidBuffer<S>,
```
jdygert-spok: Let's make this field and `BraidBuffer` itself only `pub(crate)`.

benz-spideroak: Done. `BraidBuffer` (and its `new`) is `pub(crate)`, and the `braid` field on `RuntimeBuffers` is `pub(crate)`. Also dropped the `BraidBuffer` re-export from `aranya-core::lib`. Internal callers reach it via `crate::BraidBuffer` through `pub(crate) use buffers::BraidBuffer` in `client.rs`.

jdygert-spok (approved):

crates/aranya-runtime/src/storage/spill.rs
```diff
@@ -0,0 +1,147 @@
...
+    buf: Vec<u8>,
+}
+
+#[cfg(feature = "testing")]
+impl MemSpill {
+    /// Create an empty in-memory spill.
+    pub fn new() -> Result<Self, StorageError> {
+        Ok(Self { buf: Vec::new() })
+    }
+}
```
djohnson-spideroak, line 124: The docs say that MemSpill can be used in environments without a filesystem. I think if that is a goal we should remove the testing flag.

djohnson-spideroak (approved):

PR #654, opened 2026-04-13
core: reexport aranya-crypto members for public API

### LINKED ISSUE #625: crypto: stabilize the Engine and DefaultEngine implementations
Stabilize the Engine and DefaultEngine implementations an expose them via API.

djohnson-spideroak:
closes #625

### DIFF
--- a/crates/aranya-core/tests/crypto_reexports.rs
+++ b/crates/aranya-core/tests/crypto_reexports.rs
@@ -1,0 +1,100 @@
+//! Compile-only regression test: names every symbol re-exported from
+//! [`aranya_core::crypto`]. If a name is renamed, removed, or moved
+//! upstream, this test stops compiling — the breakage surfaces here
+//! at the facade rather than silently in downstream integrations.
+
+#![cfg(all(feature = "std", feature = "memstore"))]
+#![allow(dead_code)]
+
+// Each `use` forces the path to resolve at compile time. That is the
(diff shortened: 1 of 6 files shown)


crates/aranya-core/tests/crypto_reexports.rs
djohnson-spideroak, line 1: I wasnt sure if we actually wanted a file like this, but it seems useful. We can more quickly catch changes we might overlook otherwise if the API changes. We may even want to add compile time tests for the other exports as well.

jdygert-spok: Could probably use [cargo-public-api](https://crates.io/crates/cargo-public-api) instead.

djohnson-spideroak: Do we already use cargo-public-api? Also, it only has 150k downloads lifetime, do we trust it over just writing these cumbersome test files?

jdygert-spok: We used it some for `aranya`, we aren't using it in CI anywhere.

Writing all these manually seems tedious and error prone, and public-api would be sure to show us all the trait bounds on every type. e.g. if something stopped being `Sync`.

djohnson-spideroak:
closing in favor of https://github.com/aranya-project/aranya-core/pull/662

PR #655, opened 2026-04-14
Ignore advisory RUSTSEC-2026-0097

chip-so:
We aren't using the `log` or `thread_rng` features of `rand`, so this advisory doesn't apply to us.


djohnson-spideroak (approved):

jdygert-spok:
We do conditionally enable the thread rng. (In this version it is behind std intead of a separate feature). But we don't enable `rand/log` or use rand in any loggers so we should still be fine.

jdygert-spok (approved):

PR #659, opened 2026-04-16
runtime: forward `StorageProvider` and `PolicyStore` impls

### LINKED ISSUE #660: core: change client state to be able to accept PolicyStore and LinearStorage references
discussed in the call today 2026-04-16. Can do impl's on the references for both LinearStorage and PolicyStore so that they can be accepted as references. The goal is to allow the caller to manage where they are kept in memory.

jdygert-spok:
Closes #660.

### DIFF
--- a/crates/aranya-runtime/src/policy.rs
+++ b/crates/aranya-runtime/src/policy.rs
@@ -179,3 +179,71 @@ pub enum CommandPlacement {
     /// The command is being evaluated off-graph in an ephemeral session.
     OffGraph,
 }
+
+mod impls {
+    use alloc::boxed::Box;
+
+    use super::{PolicyError, PolicyId, PolicyStore, Sink};
+
+    impl<PS: PolicyStore> PolicyStore for &mut PS {
+        type Policy = PS::Policy;
+        type Effect = PS::Effect;
+
+        fn add_policy(&mut self, policy: &[u8]) -> Result<PolicyId, PolicyError> {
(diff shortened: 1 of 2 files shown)


chip-so (approved):

crates/aranya-runtime/src/policy.rs
```diff
@@ -179,3 +179,71 @@ pub enum CommandPlacement {
...
+        fn add_policy(&mut self, policy: &[u8]) -> Result<PolicyId, PolicyError> {
+            PS::add_policy(self, policy)
+        }
+
+        fn get_policy(&self, id: PolicyId) -> Result<&Self::Policy, PolicyError> {
+            PS::get_policy(self, id)
+        }
+    }
+
+    impl<S: Sink<Eff>, Eff> Sink<Eff> for &mut S {
```
chip-so, line 214: So if I'm understanding this correctly, the `Eff` in `S: Sink<Eff>` is actually the generic described _afterwards_? Is there a reason it's listed in that order and not the other way that would read more obviously?

djohnson-spideroak (approved):
closes #660

crates/aranya-runtime/src/policy.rs
jdygert-spok: No reason, just how I happened to write it. It seems less weird in the equivalent desugared form:
```rust
impl<S, Eff> Sink<Eff> for &mut S
    where S: Sink<Eff>
```

PR #662, opened 2026-04-17
core: reexport aranya_crypto and keystore

### LINKED ISSUE #625: crypto: stabilize the Engine and DefaultEngine implementations

djohnson-spideroak:
Reexport the Engine, DefaultCipherSuite, and keystore + related members.

### DIFF
--- a/crates/aranya-core/Cargo.toml
+++ b/crates/aranya-core/Cargo.toml
@@ -21,11 +21,16 @@ default = []
 libc = ["aranya-runtime/libc"]
 
 std = [
+	"aranya-crypto/getrandom",
 	"aranya-crypto/std",
 	"aranya-runtime/std",
 ]
 
 low-mem-usage = ["aranya-runtime/low-mem-usage"]
+
+custom-engine = []
+
+memstore = ["aranya-crypto/memstore"]
 
 [dev-dependencies]
 anyhow = "1"
@@ -46,4 +51,4 @@ aranya-policy-vm = { path = "../aranya-policy-vm" }
 [[example]]
 name = "simple"
 path = "examples/simple.rs"
-required-features = ["std", "libc"]
+required-features = ["std", "libc", "memstore"]
--- a/crates/aranya-core/src/lib.rs
+++ b/crates/aranya-core/src/lib.rs
@@ -6,6 +6,8 @@
 //!
 //! # Modules
 //!
+//! - [`crypto`] — cryptography engine and cipher suite.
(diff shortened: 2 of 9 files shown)


chip-so (approved):

crates/aranya-core/src/lib.rs
```diff
@@ -64,7 +64,69 @@ pub mod policy {
...
+    #[doc(inline)]
+    pub use aranya_crypto::{
+        CipherSuite, Engine, Rng, UnwrapError, WrapError,
+        default::{DefaultCipherSuite, DefaultEngine},
+    };
+    #[cfg(feature = "custom-engine")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "custom-engine")))]
+    #[doc(inline)]
+    pub use aranya_crypto::{
+        Csprng, Random,
```
jdygert-spok: `Csprng` and `Random` should be always provided.

jdygert-spok: Or at least, they should be _usable_. I could see wanting to gate the ability to implement `Csprng` :thinking:

crates/aranya-core/Cargo.toml
```diff
@@ -21,12 +21,17 @@ default = []
...
+	"aranya-crypto/getrandom",
 	"aranya-crypto/std",
 	"aranya-runtime/std",
 ]
 
 low-mem-usage = ["aranya-runtime/low-mem-usage"]
 
+custom-engine = []
+
+memstore = ["aranya-crypto/memstore"]
```
jdygert-spok, line 33: Should we just put this and the memory-backed `IoProvider` behind a `testing` feature?

chip-so: I think so. Or at least some feature flag. The policy runner doesn't even use it (though it could).

djohnson-spideroak: Is it insecure? Or just not used? I think if it could be used, we should keep it.

chip-so: It's not clear to me where an ephemeral key store _would_ be used, but I suppose it's not impossible.

djohnson-spideroak: I can imagine using/needing a more ephemeral device, so I think we can keep it in.

jdygert-spok (approved):

chip-so (approved):

PR #663, opened 2026-04-17
runtime: reuse segments when finding LCA

jdygert-spok:
Refactored `last_common_ancestor`. Unexpected cases now return a `Bug`. The segments are kept across loop iterations, and only loaded when moving to a new segment.

### DIFF
--- a/crates/aranya-runtime/src/client/braiding.rs
+++ b/crates/aranya-runtime/src/client/braiding.rs
@@ -1,5 +1,6 @@
 use alloc::vec::Vec;
 
+use buggy::{BugExt as _, bug};
 use tracing::trace;
 
 use crate::{ClientError, Location, Prior, Segment as _, Storage, storage::TraversalBuffer};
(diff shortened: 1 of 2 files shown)


chip-so (approved):

djohnson-spideroak (dismissed):

crates/aranya-runtime/src/client/braiding.rs
```diff
@@ -19,60 +20,34 @@ impl From<strand_heap::ParallelFinalize> for ClientError {
...
     trace!(%left, %right, "finding least common ancestor");
-    let mut left = left;
-    let mut right = right;
+    let mut left_seg = storage.get_segment(left)?;
+    let mut right_seg = storage.get_segment(right)?;
     while left != right {
-        let left_seg = storage.get_segment(left)?;
-        let right_seg = storage.get_segment(right)?;
         // The command with the lower max cut could be our least common ancestor
         // so we keeping following the command with the higher max cut until
```
apetkov-so: nit: keeping->keep

djohnson-spideroak (approved):

apetkov-so (approved):

PR #665, opened 2026-04-20
policy-ast: add `WithSpan` wrapper and `WithSpanExt` helpers

jdygert-spok:
`WithSpan` replaces each of the adhoc wrappers which just add a span, so now e.g. `type VType = WithSpan<TypeKind>`. `WithSpanExt` adds extension methods to wrap a type into a `WithSpan`.


jdygert-spok:
I recommend viewing at least the first commit separately since that one is just a large rename.

apetkov-so (approved):

djohnson-spideroak (approved):

PR #667, opened 2026-04-20
core: expose aranya-id

### LINKED ISSUE #627: aranya-id: stabilize
Stabilize the public API of aranya-id, ensuring external crates can use the rich ID features.

djohnson-spideroak:


crates/aranya-core/src/id.rs
jdygert-spok, line 1: We can just re-export the macro. It uses `$crate` to refer to `aranya_id` which works even if you don't directly depend on it. This works for declarative macros but not for proc macros, since they don't have anything equivalent to `$crate`.

jdygert-spok (approved):

chip-so (approved):

PR #669, opened 2026-04-21
policy: add unit type

jdygert-spok:
`unit` is a type and a literal value.

Adds `TypeKind::Unit`, `Value::Unit`, and `ExprKind::Unit`.

Unit is not handled specially in any way, like omitting it from the stack or determining what functions can be used as a statement.


mooreoak (commented):
I don't have strong feelings but this seems fine. If were going to use a word possibly "Nothing", or "Empty" or something like that would be more understandable to people who don't know what the unit type is.

jdygert-spok:
But it isn't nothing, it is one thing :)

apetkov-so:
I was expecting a unit type to be more special. Like not being allowed on the stack. If unit is not nothing, then all functions are pure! E.g. `function foo() unit {  return unit }`.

crates/aranya-policy-lang/src/lang/parse/policy.pest
```diff
@@ -152,6 +154,8 @@ foreign_function_call = { identifier ~ "::" ~ function_call }
 // Split "Result::Ok" into enum name and value name in order to resolve value to integer
 enum_reference = { identifier ~ "::" ~ identifier }
 
+unit_literal = { "unit" }
```
chip-so: How do you feel about making the literal `Unit` with a capital U, to match Ok/Err/Some/None?

crates/aranya-policy-vm/tests/vm.rs
```diff
@@ -2626,3 +2626,97 @@ fn test_match_patterns() -> anyhow::Result<()> {
...
+
+        command DoMatch {
+            fields {
+                n int
+            }
+            seal { return todo() }
+            open { return todo() }
+            policy {
+                match verify(this.n) {
+                    Ok(Unit) => {
```
chip-so, line 2843: In fact, here it is capitalized, which should be an error but I guess it gets interpreted as a variable binding? Or maybe it was intentional but just confusing?

crates/aranya-policy-lang/src/lang/parse/policy.pest
jdygert-spok: I could go either way. It's sort of a primitive like bool with true/false. But it also could match those like you said...

Semi related, the types option and result could be capitalized too. Maybe bytes and string too? IDK

crates/aranya-policy-vm/tests/vm.rs
jdygert-spok: oops

benz-spideroak (dismissed):

crates/aranya-policy-lang/src/lang/parse/policy.pest
jdygert-spok: Made literal uppercase

apetkov-so:
`unit` needs to be added to keywords.txt.

apetkov-so (dismissed):

crates/aranya-policy-vm/tests/vm.rs
```diff
@@ -2812,3 +2902,76 @@ fn test_unit() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn test_recall_with_args() -> anyhow::Result<()> {
```
apetkov-so, line 2907: Why do we need this test for unit types?

jdygert-spok: It wasn't added here, if you look at the full diff you can see it was already there. Not sure why it's showing as added on your comment.

chip-so (approved):

apetkov-so (approved):

PR #670, opened 2026-04-21
Check FFI structure when loading module

### LINKED ISSUE #652: policy-compiler/VM: Check FFI imports are stable between compile and VM.
Geoff had a suggested implementation here: #610. This is a basic step towards some kind of full policy schema compatibility check.

chip-so:
Currently a basic implementation that checks FFI order and names. Intended to be extensible to include everything else that constitutes the module's contract. This will probably subsume some existing fields like the fact schema defs.


crates/aranya-policy-vm/src/machine.rs
```diff
@@ -191,6 +194,19 @@ impl Machine {
...
+                progmem: m.progmem.into(),
+                labels: m.labels,
+                action_defs: m.action_defs,
+                command_defs: m.command_defs,
+                fact_defs: m.fact_defs,
+                struct_defs: m.struct_defs,
+                enum_defs: m.enum_defs,
+                codemap: m.codemap,
+                globals: m.globals,
+                contract: Some(m.contract),
```
chip-so, line 210: I created a new module version, V1. I don't think it was totally necessary, but it was a good exercise in what that looks like. It's kind of a pain in the ass. Rather than unpacking all of this, maybe it would make sense to have an abstract Module trait that exposes what the internals need?

crates/aranya-policy-compiler/src/compile/target.rs
```diff
@@ -27,19 +27,24 @@ pub(crate) struct CompileTarget {
     pub fact_defs: BTreeMap<Identifier, FactDefinition>,
     /// Mapping between program instructions and original code
     pub codemap: Option<CodeMap>,
+    /// Module contract
+    pub contract: ModuleContract,
```
chip-so, line 31: I had initially put the FFI list in `PolicyInterface` but I split it out because that didn't feel right. It seems to me that the runtime contract and the interface are not exactly the same thing, and the interface shouldn't describe details the public interface never sees.

mooreoak (dismissed):
Looks reasonable, I saw you comments on making it abstract to avoid so mu copying. That seems reasonable but how often are we really going to do this?

chip-so:
> Looks reasonable, I saw you comments on making it abstract to avoid so mu copying. That seems reasonable but how often are we really going to do this?

Probably not extremely often, and if we deprecate and remove old module versions, there probably won't be more than two or three we ever have to deal with. I don't expect it to change often, either.

But there are some parts of the code that have ossified around the "V0" internal module format, and to minimize changes I wrote a `From<ModuleV1>` to convert to V0, making it kind of a "canonical" representation. Having a canonical trait would lessen the churn for the tracer and a handful of tests, but you're probably right that it is not worth the effort.

crates/aranya-policy-module/src/module/contract.rs
```diff
@@ -0,0 +1,23 @@
...
+    Deserialize,
+    rkyv::Archive,
+    rkyv::Deserialize,
+    rkyv::Serialize,
+)]
+pub struct ModuleContract {
+    /// FFI module names
+    // TODO(chip): extend this to full schema, not just the name
+    pub ffis: Vec<Identifier>,
+    // TODO(chip): catalog every other public-facing thing in a module - actions, effects, and
```
apetkov-so, line 418: What's the difference between the module and the contract? Where would things like enums go? they are defined by the module, but could be used in the public interface.

chip-so: I think the "contract" is the public interface plus the internal fact schema and FFI signatures (plus any user-defined types they touch). The goal is that everything is internally and externally consistent when one policy is swapped out for another.

Maybe "contract" is a confusing term for this. Open to alternatives.

apetkov-so (dismissed):

apetkov-so (approved):

PR #673, opened 2026-04-24
ifgen build flag for import

### LINKED ISSUE #626: policy-ifgen: stabilize the policy interface generator
Provide a stable API/version of the policy interface generator.

jdygert-spok:


crates/aranya-policy-ifgen-build/src/imp.rs
```diff
@@ -135,6 +135,12 @@ pub fn generate_code(target: &PolicyInterface) -> String {
         }
     };
 
+    let import_ifgen = ifgen.map(|ifgen| {
+        let path =
+            syn::parse_str::<syn::Path>(ifgen).expect("ifgen path should be a valid Rust path");
```
djohnson-spideroak: changing this to not panic

crates/aranya-policy-ifgen-build/src/lib.rs
```diff
@@ -12,18 +12,26 @@ mod imp;
...
 
 /// Read policy from `input` and write Rust interface to `output`.
-pub fn generate(input: impl AsRef<Path>, output: impl AsRef<Path>) -> Result<()> {
-    generate_(input.as_ref(), output.as_ref())
+pub fn generate(
+    input: impl AsRef<Path>,
+    output: impl AsRef<Path>,
+    ifgen: Option<&str>,
+) -> Result<()> {
+    generate_(input.as_ref(), output.as_ref(), ifgen)
```
jdygert-spok, line 20: Wonder if we should make a config struct/builder now.

djohnson-spideroak: What else might we want to add to this call? I guess we might need to expand this for when we add import statements to the policy for importing other policy?

crates/aranya-policy-ifgen/tests/compile_aranya_core.rs
```diff
@@ -0,0 +1,13 @@
+//! Compile-check that goldenfiles generated with `Some("aranya_core::ifgen")`
+//! actually compile through the `aranya-core` re-export.
```
jdygert-spok, line 2: Worth noting that these tests mostly only show that `use aranya_core::ifgen as aranya_policy_ifgen;` gets generated. Since the true ifgen crate is still in scope, the macros _could_ be still using that. (Although I believe they aren't.)

jdygert-spok: I think what we probably want to do is move the aranya-core simple.rs example to a separate crate so we can ensure that example works without directly depending on the internal crates, and add ifgen usage to that. (or copy and modify if we want to keep the original example).

djohnson-spideroak: It also shows that it compiles. Do you think we need to make changes to avoid the possibility of accidentally using ifgen here instead of core-ifgen? 

I guess we could change the macro to generate the qualified core path instead of this shadow-then-use approach.

edit: your reply didnt show until I posted, I can make those changes.

crates/aranya-policy-ifgen-build/src/lib.rs
jdygert-spok: Nothing exactly in mind yet. If we added type mappings or custom attributes or other code generation they would go here.

djohnson-spideroak: I think we should do it, but I do want to avoid adding additional scope to this PR. I will create a ticket for a followup.

djohnson-spideroak: https://github.com/aranya-project/aranya-core/issues/674

djohnson-spideroak:
Aranya embedded PR here: https://github.com/aranya-project/aranya-embedded/pull/73

crates/aranya-core-example/src/main.rs
```diff
@@ -225,155 +199,106 @@ fn sync_graphs(
...
-        FileManager::new(&dir_b).context("create FileManager for device B")?,
-    );
+    let provider_a =
+        LinearStorageProvider::new(FileManager::new(&dir_a).context("FileManager for device A")?);
+    let provider_b =
+        LinearStorageProvider::new(FileManager::new(&dir_b).context("FileManager for device B")?);
 
     println!("== Device Setup ==");
 
-    // Step 1: Create Device A (owner)
```
chip-so, line 274: Not sure why the commentary was removed. Seems useful for an example.

crates/aranya-core/Cargo.toml
```diff
@@ -14,6 +14,7 @@ workspace = true
 [dependencies]
 aranya-crypto = { version = "0.14.0", path = "../aranya-crypto", default-features = false }
 aranya-id = { version = "0.2.2", path = "../aranya-id", default-features = false }
+aranya-policy-ifgen = { version = "0.22.0", path = "../aranya-policy-ifgen" }
```
jdygert-spok: Let's enable the serde feature here.

```suggestion
aranya-policy-ifgen = { version = "0.22.0", path = "../aranya-policy-ifgen", features = ["serde"] }
```

jdygert-spok:
Can't approve as author but LGTM

djohnson-spideroak (approved):
Approving in place of @jdygert-spok

chip-so (approved):

PR #675, opened 2026-04-29
sync: hide wire format

djohnson-spideroak:
Move the postcard-shaped wire types (SyncType, SyncHelloType, SubscribeResult, CommandMeta) into a pub(crate) wire module and expose a SyncIncoming::decode dispatch surface so the on-wire layout can evolve without breaking consumers.

### DIFF
--- a/crates/aranya-runtime/src/sync/mod.rs
+++ b/crates/aranya-runtime/src/sync/mod.rs
@@ -1,22 +1,28 @@
 //! Interface for syncing state between clients.
+//!
+//! Transports decode incoming bytes into a [`SyncIncoming`] and dispatch on
+//! its variants; the on-wire postcard layout is hidden behind that
+//! enumeration so it can evolve without breaking consumers.
+
+use core::time::Duration;
 
 use buggy::Bug;
+use heapless::Vec;
 use postcard::Error as PostcardError;
 use serde::{Deserialize, Serialize};
 
 use crate::{
     Address, MaxCut, Prior,
     command::{CmdId, Command, Priority},
-    storage::{MAX_COMMAND_LENGTH, StorageError},
+    storage::{GraphId, MAX_COMMAND_LENGTH, StorageError},
 };
 
-mod dispatcher;
 mod requester;
 mod responder;
+mod wire;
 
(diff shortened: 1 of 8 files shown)


crates/aranya-tcp-syncer/src/lib.rs
```diff
@@ -154,7 +154,7 @@ where
         stream.write_all(&self.return_address)?;
         stream.write_all(&buffer)?;
         stream.shutdown(Shutdown::Write)?;
-        let mut received_data = std::vec::Vec::new();
+        let mut received_data = Vec::new();
```
djohnson-spideroak, line 157: look into these changes. previously we were using mixed std Vec and heapless Vec.

crates/aranya-runtime/src/sync/mod.rs
```diff
@@ -130,3 +118,195 @@ impl<'a> Command for SyncCommand<'a> {
...
+        let (sync_type, remaining) = postcard::take_from_bytes::<SyncType>(data)?;
+        let consumed = data.len().saturating_sub(remaining.len());
+        // Postcard variant tags for SyncType (5 variants) all encode in 1
+        // varint byte. The body of each variant immediately follows.
+        let body = data.get(1..consumed).unwrap_or(&[]);
+        Ok(match sync_type {
+            SyncType::Poll { request } => Self::Poll {
+                session_id: request.session_id(),
+                raw: body,
+            },
```
jdygert-spok: This seems funky. We mostly ignore the `SyncRequestMessage`, manually extract the body, then that later gets fed back into `poll` which again deserializes it as `SyncRequestMessage`.

Let's just make `SyncIncoming::Poll` hold an opaque type like `PushIncoming` and feed that to `poll`.

crates/aranya-runtime/src/sync/mod.rs
jdygert-spok, line 1: We probably want all the variants of `SyncIncoming` and `SyncHello` to either hold an opaque type or at least be `#[non_exhaustive]`

jdygert-spok (commented):
LGTM other than my two notes

chip-so (approved):

jdygert-spok (approved):

PR #676, opened 2026-05-04
storage: use `rkyv` for serialization

jdygert-spok:
- Use `rkyv` for linear storage serialization
- Custom `NonEmpty` instead of `vec1` crate
- `TrieMap` for facts
- `usize` -> `u64` in some places
- `Read::Handle` lets us use e.g. `&ArchivedSegmentIndex` directly instead of deserializing


crates/aranya-runtime/src/policy.rs
```diff
@@ -35,11 +34,26 @@ impl From<core::convert::Infallible> for PolicyError {
...
+    PartialOrd,
+    Eq,
+    PartialEq,
+    serde::Serialize,
+    serde::Deserialize,
+    rkyv::Archive,
+    rkyv::Serialize,
+    rkyv::Deserialize,
+)]
+// #[archive(as = "Self")]
```
benz-spideroak: Why not use this instead of `DeserInfallible<T>`?

crates/aranya-runtime/src/storage/linear/triemap.rs
```diff
@@ -0,0 +1,322 @@
...
+                Slot::Leaf(_) => return Err(InvalidDepth),
+            }
+        }
+        Ok(TrieMapIter {
+            stack: vec![(path, slot)],
+        })
+    }
+
+    pub fn remove(&mut self, _keys: impl IntoIterator<Item: AsRef<[u8]>>) {
+        todo!()
```
benz-spideroak, line 154: Is this reachable? It appears it'll be called when `self.prior.is_none()`.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -377,8 +382,37 @@ impl File {
...
+    fn dump_root(&self, offset: i64, root: &Root) -> Result<(), StorageError> {
+        const SIZE: usize = size_of::<ArchivedRoot>();
+        let mut buf = [0u8; SIZE];
+        let mut writer = rkyv::ser::writer::Buffer::from(buf.as_mut_slice());
+        rkyv::api::serialize_using::<_, rkyv::rancor::Failure>(root, &mut writer)
+            .assume("can serialize root")?;
+        self.write_all(offset, &writer)
+    }
+
+    // TODO(jdygert): tests don't cover this.
```
benz-spideroak, line 387: We still need a test for this.

crates/aranya-runtime/src/policy.rs
jdygert-spok: rkyv 0.8 changed how it handles endianness so I needed to use `u64_le`. Made that change for each of these types.

crates/aranya-runtime/src/storage/linear/triemap.rs
jdygert-spok: I think it should be but this isn't covered by tests either.

crates/aranya-runtime/src/storage/linear/triemap.rs
```diff
@@ -150,8 +150,28 @@ impl TrieMap {
         })
     }
 
-    pub fn remove(&mut self, _keys: impl IntoIterator<Item: AsRef<[u8]>>) {
-        todo!()
+    /// Removes the value at `keys`, pruning any branches left empty.
+    ///
+    /// Does nothing if `keys` doesn't lead exactly to a leaf.
+    pub fn remove(&mut self, keys: impl IntoIterator<Item: AsRef<[u8]>>) {
```
jdygert-spok: I think we want to error with `InvalidDepth` if it doesn't lead to a leaf.

crates/aranya-runtime/src/storage/linear/mod.rs
```diff
@@ -876,18 +885,12 @@ impl<R: Read> Segment for LinearSegment<R> {
...
             .repr
             .max_cut
-            .checked_add(
-                self.repr
-                    .commands
-                    .len()
-                    .checked_sub(1)
-                    .assume("must not overflow")? as u64,
-            )
+            .checked_add(self.repr.commands.last_index() as u64)
```
mooreoak: `as` casts need safety justifications. (or you could use `try_into()`)

crates/aranya-runtime/src/storage/linear/triemap.rs
```diff
@@ -0,0 +1,419 @@
+use alloc::{collections::BTreeMap, vec, vec::Vec};
+use core::mem;
+
+use rkyv::{Archive, Deserialize, Serialize};
+
+use super::Bytes;
+
```
mooreoak: This really needs a comment describing what at triemap is. It's really hard to understand what all this slot stuff is without a understanding the theory or mental model of the structure.

crates/aranya-runtime/src/storage/linear/triemap.rs
```diff
@@ -0,0 +1,419 @@
...
+    fn empty() -> Self {
+        Self { stack: Vec::new() }
+    }
+}
+
+impl<'a> Iterator for TrieMapIter<'a> {
+    type Item = (Vec<&'a [u8]>, Option<&'a [u8]>);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
```
mooreoak: can this be a `while let (path, slot) = self.stack.pop()? {` to clear that it terminates?

mooreoak (changes requested):
This looks gennrealy good but I really need documentation about the triemap to review it,.

PR #688, opened 2026-05-20
runtime: breaking changes for rkyv support with u64

jdygert-spok:
Makes `SegmentIndex`, `MaxCut`, and `PolicyId` opaque wrappers of `u64`.

This will allow them to be `u64_le` wrappers later so they can implement `rkyv::Portable` and thus archive as themselves.

### DIFF
--- a/crates/aranya-runtime/src/client/transaction.rs
+++ b/crates/aranya-runtime/src/client/transaction.rs
@@ -682,7 +682,7 @@ mod test {
             let mut max_cuts = HashMap::new();
             let mut buffer = TraversalBuffer::new();
             for (max_cut, &id) in ids.iter().enumerate() {
-                let max_cut = MaxCut(max_cut);
+                let max_cut = MaxCut::new(max_cut as u64);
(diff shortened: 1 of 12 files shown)


crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -682,7 +682,7 @@ mod test {
             let mut max_cuts = HashMap::new();
             let mut buffer = TraversalBuffer::new();
             for (max_cut, &id) in ids.iter().enumerate() {
-                let max_cut = MaxCut(max_cut);
+                let max_cut = MaxCut::new(max_cut as u64);
```
djohnson-spideroak, line 685: potential panic?

jdygert-spok: There's a bunch of these added, it is only a potential issue if `usize = u128` which I don't think is worth worrying about currently.

djohnson-spideroak (approved):

benz-spideroak (approved):

PR #692, opened 2026-05-27
afc: lender type for in-memory state

jdygert-spok:
Gives the in-memory AFC state's seal/open ctx direct access to the channel while still letting it be revoked by the write side of the state. This improves the fast path to be on par with the shm-based state, performing only an atomic load to check validity. (It's actually generally better since _only_ revoking that channel will invalidate it, compared to the shm ctx cache being invalidated on any change.)


chip-so (dismissed):

crates/aranya-fast-channels/src/memory/lender.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    /// Accesses the shared data `S`.
+    pub fn shared(&self) -> &S {
+        &self.inner().shared
+    }
+
+    /// Accesses the exclusive data `X`.
+    ///
+    /// Returns `None` if access is currently lent via [`Self::lend`].
+    pub fn exclusive_ref(&self) -> Option<&X> {
```
mooreoak: can't a lend happen while this &X ref is live?

mooreoak (commented):
It would be good to have a comment with a state machine in here that talks about drops

```
Start -> STATE_UNSHARED
STATE_UNSHARED -- lend() -> STATE_SHARED
STATE_UNSHARED -- drop(Owner) -> free(inner) -> End
STATE_SHARED -- drop(Owner) -> STATE_CLOSED
STATE_SHARED -- drop(Loan) -> STATE_UNSHARED
STATE_CLOSED -- drop(Loan) -> free(inner) -> End
```
It that's probably not clear enough but something like that.

Over all I think this is reasonable but I really would like to have a bunch more comments in all parts of this change to explain what and why it is doing what it is doing.

crates/aranya-fast-channels/src/memory/lender.rs
jdygert-spok: Oops, right. I tacked this on at the end for the debug impl so I'll just remove it.

jdygert-spok:
Added lots of comments, a state diagram in mermaidjs, and changed name to `Lender` and `Loan` which seems more clear to me.

crates/aranya-fast-channels/src/memory/lender.rs
```diff
@@ -0,0 +1,186 @@
...
+}
+
+/// Only [`Lender`] has access to the data.
+const STATE_UNSHARED: u8 = 0;
+/// [`Lender`] has lent access to [`Loan`].
+const STATE_SHARED: u8 = 1;
+/// [`Lender`] has dropped and revoked access from a potential [`Loan`].
+const STATE_CLOSED: u8 = 2;
+
+impl<S, X> Lender<S, X> {
```
mooreoak, line 114: so it's intended that S and X are moved in to Lender and not taken by ref?

mooreoak (commented):
I like these changes but I still don't really understand the why of this. There also don't seem to be tests of the inverants here. I think this would be a good candidate for mutation testing.

crates/aranya-fast-channels/src/memory/lender.rs
jdygert-spok: Yes. Think of it as somewhat similar to `Arc<(S, Mutex<X>)>`. It lets multiple places have access to some data while being `'static`.

chip-so (dismissed):

crates/aranya-fast-channels/src/memory/lender.rs
```diff
@@ -0,0 +1,179 @@
...
+        sync::atomic::{AtomicBool, Ordering},
+    };
+
+    // BiArc is an Arc-like type which supports 1 or 2 handles.
+    pub struct BiArc<T>(NonNull<BiArcInner<T>>);
+
+    // SAFETY: BiArc is thread-safe.
+    unsafe impl<T: Send> Send for BiArc<T> {}
+    // SAFETY: BiArc is thread-safe.
+    unsafe impl<T: Sync> Sync for BiArc<T> {}
```
benz-spideroak: Would these be better as:

```
unsafe impl<T: Send + Sync> Send for BiArc<T> {}
unsafe impl<T: Send + Sync> Sync for BiArc<T> {}
```

So you can't get a situation where get_unconditional() is called on two different threads and both obtain &T concurrently.

chip-so (approved):

benz-spideroak (approved):

PR #697, opened 2026-05-29
runtime: improve skip list and find-needed-segments traversal

benz-spideroak:
Skip-list construction (storage/linear/mod.rs):
- Replace per-segment random skip targets with deterministic exponentially-spaced targets (N/2, 3N/4, 7N/8, ...), producing ~log2(N/10) entries per segment for O(log n) worst-case search.
- Build skip-list targets via a single backwards walk that reuses intermediate segments' skip lists to jump when possible.
- Sparse construction: skip building when an ancestor within MIN_SKIP_GAP (10) already carries a rich (len > 1) skip list. The check walks through merges via their LCA — the sole entry in an LCA-only list — so a rich anchor past a merge is still reachable.
- Merge segments walk from the LCA instead of the parent and always include the LCA as a skip entry, preserving the braiding invariant that the last skip entry is the LCA.
- All arithmetic uses checked_* + assume() per the crate's established pattern; no #[allow(arithmetic_side_effects)].

find_needed_segments skip optimization (sync/responder.rs):
- Before the main traversal, jump from head to approximately highest_have + SEGMENT_BUFFER_MAX using skip-list entries that stay at or above the target. Reduces per-round cost from O(n) to O(log n + SEGMENT_BUFFER_MAX), cutting total cost across sync rounds from O(n^2) to O(n log n).
- Measured 25x reduction in sync_reads on the 10K-command linear sync benchmark.

Test DSL extensions (testing/dsl.rs, testing/testdata/*):
- Add HelloSubscribe / HelloUnsubscribe rules and the process_hello_notifications cascade helper to model notification-driven sync in tests.
- Introduce SyncMethod enum with Poll, HelloSync, and None variants, plus HelloTopology (HubAndSpoke / Ring) for GenerateGraph. None distributes commands round-robin with no in-flight sync, useful for batched-convergence benchmarks.
- New test vectors: hello_sync, generate_graph_hello_sync, sync_all_at_once. Existing generate_graph reshaped to use the new SyncMethod schema.

### DIFF
--- a/crates/aranya-runtime/src/storage/linear/mod.rs
+++ b/crates/aranya-runtime/src/storage/linear/mod.rs
@@ -30,7 +30,6 @@ pub mod testing;
 use alloc::{boxed::Box, collections::BTreeMap, string::String, vec, vec::Vec};
 use core::ops::Bound;
 
-use aranya_crypto::{Rng, dangerous::spideroak_crypto::csprng::rand::Rng as _};
 use buggy::{Bug, BugExt as _, bug};
 use serde::{Deserialize, Serialize};
 use vec1::Vec1;
@@ -273,40 +272,6 @@ impl<FM: IoManager> StorageProvider for LinearStorageProvider<FM> {
         &mut self,
     ) -> Result<impl Iterator<Item = Result<GraphId, StorageError>>, StorageError> {
         self.manager.list()
-    }
-}
-
-impl<W: Write> LinearStorage<W> {
-    fn get_skip(
-        &self,
-        segment: <Self as Storage>::Segment,
-        max_cut: MaxCut,
-    ) -> Result<Option<Location>, StorageError> {
-        let mut head = segment;
-        let mut current = None;
-        'outer: loop {
-            if max_cut > head.longest_max_cut()? {
-                return Ok(current);
-            }
-            current = Some(head.first_location());
-            if max_cut >= head.shortest_max_cut() {
-                return Ok(current);
-            }
-            // Assumes skip list is sorted in ascending order.
-            // We always want to skip as close to the root as possible.
-            for skip in head.skip_list() {
-                if skip.max_cut <= max_cut {
-                    head = self.get_segment(*skip)?;
-                    continue 'outer;
-                }
-            }
-            head = match head.prior() {
-                Prior::None | Prior::Merge(_, _) => {
-                    return Ok(current);
-                }
-                Prior::Single(l) => self.get_segment(l)?,
-            }
-        }
     }
 }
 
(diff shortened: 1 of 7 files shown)


crates/aranya-runtime/src/storage/linear/mod.rs
```diff
@@ -570,43 +537,143 @@ impl<F: Write> Storage for LinearStorage<F> {
...
+        let skip_list = if let Some(walk_start) = walk_start {
+            let n = perspective.max_cut.get();
+
+            // If a nearby ancestor (within MIN_SKIP_GAP) already has a
+            // rich skip list, defer to it. `len() > 1` excludes the
+            // LCA-only list every merge segment carries, which isn't a
+            // strong enough anchor on its own. Walk through merges via
+            // their LCA.
+            let mut needs_skip_list = true;
+            {
```
mooreoak, line 563: Why do we have these {}?

crates/aranya-runtime/src/storage/linear/mod.rs
mooreoak, line 530: This function is really long, can we break it up it to more functions so it is easier to fallow?

crates/aranya-runtime/src/storage/linear/mod.rs
```diff
@@ -570,43 +537,143 @@ impl<F: Write> Storage for LinearStorage<F> {
...
+        let (walk_start, lca) = match perspective.prior {
+            Prior::None => (None, None),
+            Prior::Merge(_, _) => {
+                let lca = perspective.last_common_ancestor.assume("lca must exist")?;
+                (Some(lca), Some(lca))
+            }
+            Prior::Single(l) => (Some(l), None),
+        };
+
+        let skip_list = if let Some(walk_start) = walk_start {
```
mooreoak, line 554: I think this would be cleaner as a sub which has return points rather than a very large expression.

crates/aranya-runtime/src/storage/linear/mod.rs
```diff
@@ -570,43 +537,143 @@ impl<F: Write> Storage for LinearStorage<F> {
...
+            } else {
+                // Build skip targets at N/2, 3N/4, 7N/8, 15N/16, ...
+                // until the gap is <= MIN_SKIP_GAP.
+                //
+                // Loop invariant: 0 < boundary < n. boundary starts at n/2
+                // (positive because the n < MIN_SKIP_GAP case returned
+                // above), and each iteration adds at most gap/2 < gap, so
+                // boundary stays strictly below n.
+                let mut targets = vec![];
+                {
```
mooreoak, line 597: not sure why we need this bare block either. I think this is another candidate to become a function thou.

mooreoak (changes requested):
This is hard to review as is. The results sound great but I would like to see a few changes so it's easier to understand.  

This seems like two different efforts in on PR but the skip list changes and some unrelated HelloSync change. If it's not too much of a pain it would be nice to see those in different PRs.

As a doc comment or a small spec it would be nice to have a description of the algorithm being implemented. I think I understand it but I am not sure I have it right.

Some of these functions a very long containing many loops I think it would be easier to read if each of those loops was a function so instead of:
```
fn () {
  // do step comment
  loop {
    ...
  }

  // do step comment
  loop {
    ...
  }
  ...
}
``
we hand
```
fn () {
  // do step comment
 let (updated, state) = step_funcntion(local, state)?;

  // do step comment
 let (updated, state) = step_funcntion(local, state)?;
}
```

PR #698, opened 2026-06-01
runtime: add hello-sync DSL and refine GenerateGraph schema

benz-spideroak:
Testing scaffolding for hello sync, plus a richer distribution model for GenerateGraph.

DSL additions (testing/dsl.rs):
- HelloSubscribe / HelloUnsubscribe rules model peers that receive proactive sync hints from each other.
- process_hello_notifications cascade helper drains pending hello notifications and triggers responsive sync.
- SyncMethod enum (Poll, HelloSync, None) on GenerateGraph, plus HelloTopology (HubAndSpoke / Ring). None distributes commands round-robin with no in-flight sync — useful for batched-convergence scenarios.

Test vectors (testing/testdata/*):
- New: hello_sync, generate_graph_hello_sync, sync_all_at_once.
- Existing generate_graph reshaped to use the new SyncMethod schema.

### DIFF
--- a/crates/aranya-runtime/src/testing/dsl.rs
+++ b/crates/aranya-runtime/src/testing/dsl.rs
@@ -649,7 +649,7 @@ where
                                         });
                                         count += 1;
                                     }
-                                    x if x < sync_ceiling => {
+                                    _ => {
                                         let mut from = (client + 1) % clients;
                                         if !sync_client_zero && from == 0 {
                                             from += 1;
@@ -663,7 +663,6 @@ where
                                             max_syncs: 1,
                                         });
                                     }
-                                    _ => {}
                                 }
                             }
                             // Converge all clients by syncing every pair in both
--- a/crates/aranya-runtime/src/testing/dsl.rs
+++ b/crates/aranya-runtime/src/testing/dsl.rs
@@ -84,6 +84,65 @@ fn default_max_syncs() -> u64 {
     1
 }
 
+fn default_max_cascade_depth() -> u64 {
+    100
+}
+
+fn default_notify_interval() -> u64 {
+    1
+}
+
+/// Tracks per-subscriber state for hello sync debouncing.
+#[derive(Clone, Debug)]
+struct HelloSub {
+    /// Notify after this many graph changes.
+    notify_interval: u64,
+    /// Graph changes since last notification.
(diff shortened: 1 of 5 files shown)


crates/aranya-runtime/src/testing/dsl.rs
```diff
@@ -385,77 +580,233 @@ where
...
+                            assert!(
+                                clients >= min_clients,
+                                "There must be at least {min_clients} clients"
+                            );
+                            assert!(
+                                add_command_chance > 0,
+                                "There must be a positive command chance or it will never exit"
+                            );
+                            // Calculate the maximum number of syncs needed to send all commands.
+                            // We add 100 to account for extra syncs needed for merge commands.
```
chip-so, line 631: This feels imprecise. What are the odds this isn't enough?

crates/aranya-runtime/src/testing/dsl.rs
```diff
@@ -385,77 +580,233 @@ where
...
+                                        generated_actions.push(TestRule::ActionSet {
+                                            client,
+                                            graph,
+                                            key: 0,
+                                            value: rng.gen_range(0..10),
+                                            repeat: 1,
+                                        });
+                                        count += 1;
+                                    }
+                                    x if x < sync_ceiling => {
```
chip-so: Since the range is `0..sync_ceiling`, isn't this always true?

crates/aranya-runtime/src/testing/dsl.rs
benz-spideroak: The reason for this to not allow an infinite loop. We should never need close to 100. If we do hit it it's likely a bug. Can you think of a clearer way to express that?

crates/aranya-runtime/src/testing/dsl.rs
benz-spideroak: Fixed.

crates/aranya-runtime/src/testing/dsl.rs
chip-so: If you're confident it's enough buffer that it'll always work, it's fine with me.

chip-so (approved):

apetkov-so (approved):

PR #699, opened 2026-06-01
runtime: improve skip list and find-needed-segments traversal

benz-spideroak:
Implements the design changes from aranya-docs PR #199 (https://github.com/aranya-project/aranya-docs/pull/199).

Skip-list construction (storage/linear/mod.rs):
- Replace per-segment random skip targets with deterministic exponentially-spaced targets (N/2, 3N/4, 7N/8, ...), producing ~log2(N/10) entries per segment for O(log n) worst-case search.
- Build skip-list targets via a single backwards walk that reuses intermediate segments' skip lists to jump when possible.
- Sparse construction: skip building when an ancestor within MIN_SKIP_GAP (10) already carries a rich (len > 1) skip list. The check walks through merges via their LCA — the sole entry in an LCA-only list — so a rich anchor past a merge is still reachable.
- Merge segments walk from the LCA instead of the parent and always include the LCA as a skip entry, preserving the braiding invariant that the last skip entry is the LCA.

find_needed_segments skip optimization (sync/responder.rs):
- Before the main traversal, jump from head to approximately highest_have + SEGMENT_BUFFER_MAX using skip-list entries that stay at or above the target. Reduces per-round cost from O(n) to O(log n + SEGMENT_BUFFER_MAX), cutting total cost across sync rounds from O(n^2) to O(n log n).
- Measured 25x reduction in sync_reads on the 10K-command linear sync benchmark.

### DIFF
--- a/crates/aranya-runtime/src/storage/linear/mod.rs
+++ b/crates/aranya-runtime/src/storage/linear/mod.rs
@@ -279,9 +279,11 @@ impl<FM: IoManager> StorageProvider for LinearStorageProvider<FM> {
 const MIN_SKIP_GAP: u64 = 10;
 
 /// Skip-list target boundaries for a segment of length `n`: `n/2`, `3n/4`,
-/// `7n/8`, ..., halving the remaining gap each step until the gap to `n`
-/// drops to [`MIN_SKIP_GAP`]. Returned ascending; callers walk backwards
-/// and pop highest-first. Empty when `n < 2`.
+/// `7n/8`, ..., halving the remaining gap each step. Continues until the
+/// gap from the final boundary to `n` is ≤ [`MIN_SKIP_GAP`], so the walk
+/// from head to the first skip entry never exceeds the cheap-walk
+/// threshold. Returned ascending; callers walk backwards and pop
+/// highest-first. Empty when `n < 2`.
 fn skip_target_boundaries(n: u64) -> Result<Vec<MaxCut>, StorageError> {
     let mut targets = vec![];
     let mut boundary = n / 2;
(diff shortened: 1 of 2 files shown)


crates/aranya-runtime/src/storage/linear/mod.rs
```diff
@@ -274,38 +273,31 @@ impl<FM: IoManager> StorageProvider for LinearStorageProvider<FM> {
...
+        targets.push(MaxCut::new(boundary));
+        let gap = n
+            .checked_sub(boundary)
+            .assume("boundary < n by loop invariant")?;
+        if gap <= MIN_SKIP_GAP {
+            break;
         }
+        boundary = boundary
+            .checked_add(gap / 2)
+            .assume("boundary + gap/2 <= n <= u64::MAX")?;
```
chip-so, line 300: This does produce a final gap in the last element that is less than `MIN_SKIP_GAP`. e.g. `n = 100` produces `[50, 75, 87, 93]`. I'm not sure from the description if that's intentional.

Also I thought this would be a simpler way to express it:
```
let mut gap = n / 2;
let mut boundary = gap;
while boundary > 0 {
    targets.push(MaxCut::new(boundary));
    if gap < MIN_SKIP_GAP {
        break;
    }
    gap = gap / 2;
    boundary = boundary
        .checked_add(gap)
        .assume("boundary + gap/2 <= n <= u64::MAX")?;
}
```

But it produces a slightly different series of values, as the error between `gap` and the actual remaining gap increases due to iterative integer division.

benz-spideroak: I updated the comment to make it clearer. I want to continue until the final gap is less than MIN_SKIP_GAP

chip-so (approved):

mooreoak (approved):

PR #700, opened 2026-06-01
Close #648: Invoke recall blocks directly from policy, without roundtrip to runtime.

### LINKED ISSUE #678: Improve recall
https://github.com/aranya-project/aranya-core/pull/642#discussion_r3169701965

Update: we could simplify recalls further by calling them directly from the policy, without check-exiting. Currently, the runtime recalls only if the command is on-graph, but this check is not really necessary; it's okay to recall all commands, including ephemeral ones.

Changes:

1. The `recall` statement/expression compiles to a function-like call to the named recall block. This is followed by Exit(Check) so runtime still knows that command was recalled.
2. What should `evaluate_rule` do in the case of ExitReason::Check?
3. Remove the `recall_internal` method.

Effects emitted in recall contexts continue to be tagged as `recalled`.

apetkov-so:


chip-so (dismissed):
This really cleans things up, huh? :)

apetkov-so:
Fixing the broken tests is actually a bit tricky. Recall effects are not tagged as `recalled` because the RunState context is no longer updated (the VMmPolicy used to to that via `set_context`).

One fix is to add a SetContext instruction and emit that before calling the recall block. Ideally we should use a context stack in the runstate (for `finish` inside `recall`), but `finish` is terminal so we can get with a just clobbering `ctx`.

Edit: Found simpler fix. We just figure out if the function-like block we're about to call is a recall block, and set the context accordingly.

crates/aranya-policy-vm/src/machine.rs
```diff
@@ -596,6 +596,16 @@ where
...
+                    // Set context to Recall before invoking recall block - it affects emitted effects
+                    if self
+                        .machine
+                        .labels
+                        .iter()
+                        .any(|(label, &a)| a == n && label.ltype == LabelType::CommandRecall)
+                        && let CommandContext::Policy(c) = &self.ctx
+                    {
+                        self.ctx = CommandContext::Recall(c.clone());
+                    }
```
jdygert-spok: Not a huge fan of scanning the list of labels to modify the context at runtime. I think a recall instruction which transitions from policy to recall ctx would be fine.

Or we could replace `call <recall target>` with `recall <recall target>`. I think there are some clever things we could do with that. Instead of treating it as a full function call we can make it just a jump which resets the scope list to the top layer containing only `this` and `envelope`.

apetkov-so: I didn't understand the "or" part. Do you mean add this instruction either way, but implement it differently?

jdygert-spok: Yeah, I think either way we want a new instruction for swapping the context rather than relying on finding label type. The "or" part is just a way I'm thinking we could simplify it, unsure whether it's worth doing.

Option 1:
```
  get 'envelope'
  get 'this'
  recall # switches context
  call <my_recall> # function call

my_recall:
  def 'this'
  def 'envelope'
  def 'my_arg'
  ...
```

Option 2:
```
  recall <my_recall> # switch context, pop local stack to base, jump to target

my_recall:
  def 'my_arg'
  ...
```

apetkov-so: Why not a `setcontext` instruction? We could use it for all contexts, and remove RunState::set_context.

jdygert-spok: This will already let us remove that since the only usage is for converting to recall.

chip-so: I think I'd do option 1, but put the `recall` context switch at the head of the recall function itself rather than the call site. `recall` as a combo instruction is interesting, but I'd like to keep the instruction set more general if possible. More complex instructions are harder to maintain when the rest of the system changes (I think if we had to do that you'd just make a `recall2` and deprecate the old one, though).

Another idea: if you don't care about stack hygiene (and I don't think we do as recall is terminal), you could just do:
```
  get 'envelope'
  get 'this'
  get 'any_other_args'
  jump <my_recall>

my_recall:
  recall
  def 'any_other_args'
  ...
```

That is, just treat it as a continuation of the same scope with a new context. If you wanted to be clean about it, you could add a `clear` instruction that clears the stack.

chip-so (approved):

djohnson-spideroak (approved):

PR #706, opened 2026-06-10
Close #705: Actions can return results

### LINKED ISSUE #705: Return from actions
See spec: https://github.com/aranya-project/aranya-docs/pull/204
- Actions can have optional `result[unit, T]` return type
- bare `check` no longer allowed - `else` clause required
- `check_unwrap` is removed

apetkov-so:


chip-so (commented):
Looks good so far!

crates/aranya-runtime/src/client.rs
```diff
@@ -53,8 +53,16 @@ pub enum ClientError {
 
 impl From<PolicyError> for ClientError {
     fn from(error: PolicyError) -> Self {
+        // Check errors used to come only from commands, and were mapped to NotAuthorized. But now
+        // that actions can return errors (which are also surfaced as check errors) we need to
+        // distinguish between them.
```
jdygert-spok: Couldn't we check fail in actions already? Do we want those to also be `NotAuthorized`?

crates/aranya-runtime/src/policy.rs
```diff
@@ -20,8 +21,10 @@ pub enum PolicyError {
     Read,
     #[error("write error")]
     Write,
+    /// A check failed. Carries an optional error value: actions that return
+    /// `Err(value)` populate it; command/recall checks leave it `None`.
     #[error("check error")]
-    Check,
+    Check(Option<Value>),
```
jdygert-spok: Putting a VM Value in this error enum directly is a bit awkward, but I'm not sure how else to handle it. I had #695 but IDK whether that make sense if we want action error to trigger a sink revert. I guess we could do `Result<Result<(), Self::ActionError>, VmError>` for action return type? Hmm...

crates/aranya-runtime/src/client.rs
apetkov-so: You mean Exit(Check) on `return Err()`? I thought about that, but actions don't exit at all; that's to allow nesting (actions calling actions). Instead, the runtime detects when the outermost action returns and the call stack becomes empty, and exits normally.

I don't think all action errors should be interpreted as NotAuthorized, because we'd lose the ability to pass along the reason for the failure.

chip-so (dismissed):

crates/aranya-model/src/tests/ffi-policy.md
```diff
@@ -78,7 +78,12 @@ function Role_Device() string {
 function authorized_device_key_ids(device_keys struct DeviceKeyBundle) struct NewDevice {
     let got_device_id = idam::derive_device_id(device_keys.ident_pk)
 
-    check got_device_id == device_keys.device_id
+    check got_device_id == device_keys.device_id else return NewDevice {
+        device_id: device_keys.device_id,
+        ident_pk: device_keys.ident_pk,
+        sign_pk_id: device_keys.device_id,
+        sign_pk: device_keys.sign_pk,
+    }
```
chip-so: It seems like this causes it to succeed when it should fail?

crates/aranya-policy-compiler/tests/data/statements/finish_not_last_unknown.fail.policy
```diff
@@ -13,4 +13,7 @@ command Foo {
         }
         let a = 5
     }
+    recall reject() {
+        finish {}
```
chip-so: Why does this one have a `finish` block but the others don't?

crates/aranya-policy-lang/tests/data/if_statement.policy
```diff
@@ -1,22 +1,22 @@
 action test() {
     if 0 {
-        check 1
+        check 1 else return Err("check failed")
```
chip-so: Shouldn't this require a return type on the action?

crates/aranya-policy-compiler/tests/data/statements/finish_not_last_unknown.fail.policy
apetkov-so: Typo, removed.

crates/aranya-model/src/tests/ffi-policy.md
apetkov-so: Oops.

crates/aranya-policy-lang/tests/data/if_statement.policy
apetkov-so: Yeah, the if 0 is wrong too. This is just a parsing test, so no type checking. Should be fixed though.

chip-so (dismissed):

crates/aranya-runtime/src/policy.rs
apetkov-so: The error value is intended for the app, so it can display a meaningful message. The runtime doesn't care about the specific reason, it just knows the action failed and can revert the sink, if that's what we want.

crates/aranya-policy-ast/src/ast.rs
```diff
@@ -1249,6 +1249,9 @@ pub struct ActionDefinition {
     pub identifier: Ident,
     /// The arguments to the action
     pub arguments: Vec<Param>,
+    /// Optional `result[unit, E]` type the action returns. `None` means the action is infallible.
+    /// The success (ok) type is limited to `unit`.
+    pub return_type: Option<ResultTypeKind>,
```
jdygert-spok: Thoughts on making this just a VType here, and doing restrictions as necessary in the compiler instead of parser?

crates/aranya-policy-compiler/src/compile/lower.rs
```diff
@@ -683,6 +683,19 @@ impl CompileState<'_> {
...
+                    // terminal) returns the action's `result[unit, E]` type.
+                    StatementContext::Action(action) => {
+                        let Some(result_type) = &action.return_type else {
+                            let msg = "cannot return from an infallible action; declare a `result[unit, E]` return type";
+                            return Err(self.err(InvalidExpression(msg, expression.clone(), None)));
+                        };
+                        if !matches!(result_type.ok.inner, TypeKind::Unit) {
+                            let msg = "an action's success type must be `unit`";
+                            return Err(self.err(InvalidExpression(msg, expression.clone(), None)));
+                        }
```
jdygert-spok: This is a weird place to check the action's declared return type. It should probably be checked only once, in `compile_action`, rather than on every return within that action.

crates/aranya-policy-compiler/src/compile/lower.rs
```diff
@@ -1577,8 +1590,7 @@ impl CompileState<'_> {
                     StmtKind::Check(s),
                     StatementContext::Action(_)
                     | StatementContext::PureFunction(_)
-                    | StatementContext::CommandPolicy(_)
-                    | StatementContext::CommandRecall(_),
+                    | StatementContext::CommandPolicy(_),
```
jdygert-spok: Why disallow check in recall?

crates/aranya-policy-compiler/src/compile/lower.rs
```diff
@@ -1715,6 +1719,35 @@ impl CompileState<'_> {
...
+                            Some(statement.span),
+                        )));
+                    };
+                    // The success type is limited to `unit`.
+                    if !matches!(result_type.ok.inner, TypeKind::Unit) {
+                        return Err(self.err(UnknownError(
+                            "an action's success type must be `unit`".to_owned(),
+                            Some(result_type.ok.span),
+                        )));
+                    }
```
jdygert-spok: ditto on checking this in returns

crates/aranya-policy-vm/tests/vm.rs
```diff
@@ -1658,7 +1694,8 @@ fn test_envelope_in_policy_and_recall() -> anyhow::Result<()> {
             }
 
             recall default() {
-                check envelope.payload == this.test
+                // `check` is not allowed in a recall block; the recall body
+                // runs with `envelope` and `this` in scope and emits.
```
jdygert-spok: This no longer tests that `this` and `envelope` are exposed properly in the recall.

crates/aranya-policy-vm/tests/vm.rs
jdygert-spok, line 1: I thought we were going to change most tests usages of `check` to `debug_assert`.

crates/aranya-policy-compiler/src/compile/lower.rs
apetkov-so: Because we don't allow nested recalls. Since bare check is gone, the else clause would have to either return or recall.

jdygert-spok: We can still `check false else todo("")` but not sure if thats useful.

crates/aranya-policy-vm/tests/vm.rs
apetkov-so: changed to `test_fail`

crates/aranya-policy-compiler/tests/data/statements/finish_not_last_unknown.fail.snap
```diff
@@ -2,9 +2,7 @@
 source: crates/aranya-policy-compiler/tests/tests.rs
 expression: error
 ---
-error: unknown error
-   |
- 8 | /         finish {
- 9 | |             delete Blah[]
-10 | |         } // finish must be the last statement in policy
-   | |_________^ `finish` must be the last statement in the block
+error: a thing being referenced is not in scope
```
chip-so: I'm a little confused about this error, as it has nothing to do with the finish block and below it appears to be erroring on the `test_fail`.

crates/aranya-policy-compiler/src/compile/lower.rs
apetkov-so: Re-allowed. We now use `check ... else test_fail("...")`

crates/aranya-policy-compiler/tests/data/statements/finish_not_last_unknown.fail.snap
apetkov-so: updated test

crates/aranya-policy-vm/tests/vm.rs
apetkov-so: restored

crates/aranya-policy-ast/src/ast.rs
apetkov-so: Okay, return type is now VType. `result[unit, E]` or `unit` for infallible actions.

crates/aranya-policy-compiler/src/compile/lower.rs
apetkov-so: Moved type check to compile.

crates/aranya-policy-compiler/tests/data/actions/non_unit_success.fail.snap
```diff
@@ -2,7 +2,9 @@
 source: crates/aranya-policy-compiler/tests/tests.rs
 expression: error
 ---
-error: unknown error
+error: an action's success type must be `unit`
   |
-2 | action bad() result[string, string] {
-  |                     ^^^^^^ an action's success type must be `unit`
+2 | / action bad() result[string, string] {
```
chip-so: I feel like the prior version was a lot clearer. :/

chip-so: Not the "unknown error" part, of course. But the pointing directly at the offending type.

chip-so (dismissed):

crates/aranya-policy-compiler/tests/data/actions/non_unit_success.fail.snap
apetkov-so: Yeah, changed back to expression.span.

chip-so (dismissed):

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1162,15 +1170,34 @@ impl<'a> CompileState<'a> {
...
+            TypeKind::Unit => None,
+            TypeKind::Result(r) => {
+                if !matches!(r.ok.inner, TypeKind::Unit) {
+                    return Err(self.err(InvalidReturn {
+                        message: "an action's success type must be `unit`".to_owned(),
+                        span: r.ok.span,
+                    }));
+                }
+                Some(action_node.return_type.clone())
+            }
```
jdygert-spok, line 1186: This is probably fine but worth noting this makes `action f() Unit` and `function f() Unit` different in that the function could have `return Unit` but the action can't.

jdygert-spok: Oh wait, this isn't tied to that. But it is still true due to the way we check for result when lowering returns.

apetkov-so: Action return types are either `result[unit, T]` or nothing, can't use `unit`.

jdygert-spok: I see this now, the grammar has `result_t?`. I thought it had been changed to `vtype?`. It feels weird to me to hard code that in the parser but I won't block on it.

crates/aranya-policy-vm/tests/vm.rs
```diff
@@ -775,28 +775,28 @@ fn test_query_partial_key() -> anyhow::Result<()> {
...
             let f = unwrap query Foo[i:?, j:?]
         }
 
         action test_exists() {
-            check exists Foo[i:1, j:?]
-            check exists Foo[i:-1, j:?] == false
-            check !exists Foo[i:1, j:?] => {x:-1, s:?}
+            check exists Foo[i:1, j:?] else test_fail("check failed")
+            check exists Foo[i:-1, j:?] == false else test_fail("check failed")
+            check !exists Foo[i:1, j:?] => {x:-1, s:?} else test_fail("check failed")
```
jdygert-spok: In a follow up, let's make the message optional and remove all the `"check failed"`, they are just adding noise.

crates/aranya-policy-vm/tests/vm.rs
```diff
@@ -923,18 +924,22 @@ fn test_if_true() -> anyhow::Result<()> {
     let mut rs = machine.create_run_state(&mut io, ctx);
 
     let result = rs.call_action(name, [true])?;
-    assert_eq!(result, ExitReason::Check);
+    // The check fails, so the action runs its `else return Err(..)`, exiting
+    // Normal with the Err result left on the stack.
+    assert_eq!(result, ExitReason::Normal);
+    assert_eq!(rs.stack.len(), 1);
```
jdygert-spok: This no longer tests that the if branch was taken.

crates/aranya-policy-vm/tests/vm.rs
```diff
@@ -923,18 +924,22 @@ fn test_if_true() -> anyhow::Result<()> {
...
 #[test]
 fn test_if_false() -> anyhow::Result<()> {
     let text = r#"
-        action foo(x bool) {
+        action foo(x bool) result[unit, string] {
             if x == true {
-                check true == false
+                check true == false else return Err("check failed")
             }
+            return Ok(Unit)
```
jdygert-spok: ditto, not testing the branch wasn't taken.

crates/aranya-policy-vm/tests/vm.rs
```diff
@@ -1059,16 +1064,17 @@ fn test_match_alternation() -> anyhow::Result<()> {
             policy {}
         }
 
-        action foo(x int) {
+        action foo(x int) result[unit, string] {
             match x {
                 0 | 1 => {
-                    check false
+                    check false else return Err("check failed")
```
jdygert-spok: this test still works but might as well use test_fail.

crates/aranya-policy-vm/tests/vm.rs
```diff
@@ -2371,15 +2378,16 @@ fn test_struct_composition() -> anyhow::Result<()> {
...
+            "action f() {{ check {expr} else test_fail(\"check failed\") }}"
+        ));
         let mut io = TestIO::new();
         let ctx = dummy_ctx_action(ident!("f"));
         let mut rs = machine.create_run_state(&mut io, ctx);
         let exit = rs
             .call_action(ident!("f"), iter::empty::<Value>())
             .expect("action runs");
         assert_eq!(exit, ExitReason::Normal);
-        assert!(rs.stack.is_empty());
```
jdygert-spok, line 2382: Multiple tests lost this check which was useful for ensuring that we don't accidentally leave extra values on the stack.

crates/aranya-policy-vm/tests/vm.rs
```diff
@@ -2396,39 +2404,59 @@ fn test_boolean_operators() {
...
-    // `todo()` panics if it runs; a failing bare `check` exits with `Check(None)`.
-    // So `Panic` means `todo()` evaluated; `Check(None)` means the check failed
-    // without `todo()` running; `Normal` means the check passed.
+    // `todo()` panics if it runs. A failing `check` runs its `else return Err(..)`,
+    // exiting `Normal`. So `Panic` means `todo()` evaluated; `Normal` means it did
+    // not (the check either passed, or failed and returned Err).
     assert_eq!(run("true && todo()"), ExitReason::Panic); // todo runs in &&
-    assert_eq!(run("false && todo()"), ExitReason::Check); // && short-circuits → check fails
+    assert_eq!(run("false && todo()"), ExitReason::Normal); // && short-circuits → check fails → return Err
     assert_eq!(run("true || todo()"), ExitReason::Normal); // || short-circuits, check passes
```
jdygert-spok, line 2399: these two normal exits no longer test the boolean output, although `test_boolean_operators` is probably enough to cover it.

crates/aranya-runtime/src/client.rs
jdygert-spok: My concern is that if these both semantically mean "this operation was not authorized", it might be confusing that only one uses that error variant.

```
action foo() result[Unit, string] {
    check precondition else return Err("not valid")
}
```

vs

```
action foo() {
    publish Bar {}
}

command Bar {
    policy {
        check precondition else recall not_valid()
    }
    recall not_valid() { }
}
```

crates/aranya-runtime/src/policy.rs
jdygert-spok: I mean that this change is specifically tying the policy interface to the `aranya_policy_vm::Value` type, whereas currently it is generic for any policy.

crates/aranya-policy-vm/tests/vm.rs
apetkov-so: Simplified and merged with test_if_false.

crates/aranya-policy-vm/tests/vm.rs
apetkov-so: Merged with test_if_true

crates/aranya-policy-vm/tests/vm.rs
apetkov-so: May as well do it here.

crates/aranya-policy-vm/tests/vm.rs
apetkov-so: Restored.

crates/aranya-policy-vm/tests/vm.rs
```diff
@@ -2396,31 +2380,49 @@ fn test_boolean_operators() {
...
 
-    // `todo()` panics if it runs; a failing bare `check` exits with `Check(None)`.
-    // So `Panic` means `todo()` evaluated; `Check(None)` means the check failed
-    // without `todo()` running; `Normal` means the check passed.
+    // `todo()` panics if it runs. A failing `check` runs its `else return Err(..)`,
+    // exiting `Normal`. So `Panic` means `todo()` evaluated; `Normal` means it did
+    // not (the check either passed, or failed and returned Err).
     assert_eq!(run("true && todo()"), ExitReason::Panic); // todo runs in &&
-    assert_eq!(run("false && todo()"), ExitReason::Check); // && short-circuits → check fails
+    assert_eq!(run("false && todo()"), ExitReason::Panic); // && short-circuits → check fails
```
jdygert-spok: These don't distinguish whether the `todo()` or the `test_fail()` ran.

apetkov-so: Changed action to result. We don't check stack-empty anymore, but that's being checked in other tests.

crates/aranya-runtime/src/client.rs
apetkov-so: I don't really understand why check errors meant not-authorized in the first place. A check failure could indicate any number of reasons. Shouldn't `Check` map to `PolicyError::Check`?

crates/aranya-runtime/src/policy.rs
apetkov-so: I like the idea of an ActionReturn associated type. 🤔

crates/aranya-runtime/src/client.rs
jdygert-spok: Maybe authorized isn't exactly the right word, but these errors do always mean that the policy determined that an operation wasn't "allowed" or "valid".

I'm open to removing the variant from `ClientError` if we think it makes more sense to just be part of `PolicyError`.

Worth noting these aren't specifically "check failures" now anyways, they are just err returns or recalls. So I think we will want to move away from calling it a check error and toward something more like not authorized or invalid operation.

chip-so (dismissed):

crates/aranya-runtime/src/client.rs
```diff
@@ -53,18 +51,7 @@ pub enum ClientError {
...
-        // command: `check .. else recall` -> Check(None)
-        // action:  `check .. else return` -> Check(Some)
-        match error {
-            PolicyError::Check(err) => match err {
-                None => Self::NotAuthorized,
-                Some(v) => Self::PolicyError(PolicyError::Check(Some(v))),
-            },
-            _ => Self::PolicyError(error),
-        }
+        Self::PolicyError(error)
```
jdygert-spok: Nit: replace this impl by using `PolicyError(#[from] PolicyError),` in the enum definition above. (via the existing `thiserror::Error` derive)

crates/aranya-runtime/src/policy.rs
```diff
@@ -21,10 +20,8 @@ pub enum PolicyError {
     Read,
     #[error("write error")]
     Write,
-    /// A check failed. Carries an optional error value: actions that return
-    /// `Err(value)` populate it; command/recall checks leave it `None`.
-    #[error("check error")]
-    Check(Option<Value>),
+    #[error("policy error")]
```
jdygert-spok: nit: let's make this more descriptive. maybe "operation was rejected by the policy"

crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -645,11 +645,11 @@ impl<CE: aranya_crypto::Engine> Policy for VmPolicy<CE> {
                                 error!("expected action result value: {e}");
                                 PolicyError::InternalError
                             })?;
-                            // `Err(payload)` => the action failed with that value;
+                            // `Err(_)` => the action failed;
                             // anything else (`Ok`/placeholder success) succeeds.
-                            if let Value::Result(Err(payload)) = value {
+                            if matches!(value, Value::Result(Err(_))) {
```
jdygert-spok: Let's log the error payload too.

crates/aranya-policy-compiler/src/compile.rs
apetkov-so: It's VType in the AST, but I'm not sure how we would we get the result[...] syntax if the grammar specifies vtype.

chip-so (approved):

djohnson-spideroak (approved):

PR #711, opened 2026-06-26
tests: add branch coverage

djohnson-spideroak:
Add branch coverage makefile tasks, add coverage job


crates/aranya-crypto/src/apq.rs
```diff
@@ -659,7 +659,7 @@ impl<CS: CipherSuite> ReceiverPublicKey<CS> {
 
 #[cfg(test)]
 mod tests {
-    use spideroak_crypto::{ed25519::Ed25519, import::Import as _, kem::Kem, rust, signer::Signer};
+    use spideroak_crypto::{ed25519::Ed25519, kem::Kem, rust, signer::Signer};
```
djohnson-spideroak: These changes were needed to resolve a lint in nightly

djohnson-spideroak:
Splitting the report into 4: CAPI coverage, compiler, tools, and the rest of the codebase named core-prod. 

The following 4 crates are excluded from all reports:

- aranya-capi-codegen
- aranya-core-example
- aranya-model
- arnaya-policy-vm-explorer

scripts/coverage.sh
djohnson-spideroak, line 1: I was thinking of changing this to use a rust program or something to do the regex construction, theres a bit too much bash logic for my taste.

jdygert-spok: I agree this is a bit much for bash. I'd personally switch it to python.

Makefile.toml
```diff
@@ -49,6 +49,79 @@ command = "cargo"
 args = ["run", "-p", "aranya-core-example"]
 
 
+# Coverage
+[tasks.coverage-full]
+category = "test"
+description = "Whole-workspace branch coverage report (per-file table to terminal)"
+toolchain = "nightly"
```
jdygert-spok: I think we should be able to use `env.BOOTSTRAP = "1"` for these to avoid nightly so we don't have to worry about pinning it. (like #542)

jdygert-spok (approved):

mooreoak (approved):

PR #715, opened 2026-06-30
test: add tests to aranya-crypto for additional branch coverage

djohnson-spideroak:


crates/aranya-crypto/src/keystore/fs_keystore/error.rs
```diff
@@ -85,7 +85,12 @@ where
 
 impl From<Bug> for Error {
     fn from(err: Bug) -> Self {
-        <Self as keystore::Error>::other(err)
+        // Construct the `Bug` variant directly rather than routing
+        // through `Repr::new`. A `Bug` can only be created via
+        // `buggy::Bug::new`, which panics in debug builds, so a
+        // runtime `downcast_ref::<Bug>` branch in `Repr::new` could
+        // never be exercised under coverage.
+        Self(Repr::Bug(err))
```
djohnson-spideroak: This is one place where the implementation changed. Overall, I think its correct to remove the panic-during-debug code, especially if we wouldn't expect it to panic outside of debug.

crates/aranya-crypto/src/keystore/fs_keystore/error.rs
```diff
@@ -127,8 +132,6 @@ impl Repr {
                 cbor::de::Error::Semantic(x, y) => cbor::de::Error::Semantic(*x, y.clone()),
                 cbor::de::Error::RecursionLimitExceeded => cbor::de::Error::RecursionLimitExceeded,
             })
-        } else if let Some(err) = err.downcast_ref::<Bug>() {
-            Self::Bug(err.clone())
```
djohnson-spideroak, line 131: Bug now falls through to other.

crates/aranya-crypto/src/groupkey.rs
```diff
@@ -162,14 +161,8 @@ impl<CS: CipherSuite> GroupKey<CS> {
...
             // Not enough room in `dst`.
-            let required = self
-                .overhead()
-                .checked_add(plaintext.len())
-                .ok_or(Error::Bug(Bug::new(
-                    "overhead + plaintext length must not wrap",
-                )))?;
             return Err(Error::Seal(SealError::BufferTooSmall(BufferTooSmallError(
-                Some(required),
+                self.overhead().checked_add(plaintext.len()),
```
djohnson-spideroak, line 165: This change results in different error text when hitting the wrap. I can add the context back in if needed, but this shouldn't change behavior.

crates/aranya-crypto/src/keystore/fs_keystore/error.rs
djohnson-spideroak: I am going to revert this change (and the cascading change) and let this branch go uncovered/add an exemption. 

I will also open a ticket to evaluate where we use buggy to identify places we would rather have control flow over buggy.

crates/aranya-crypto/src/groupkey.rs
djohnson-spideroak: I am keeping this change, the only consumer is the display impl. The new behavior is that if it hits the wrap when calculating the hint, it returns BufferTooSmall(None) and displays properly without a hint. It should be impossible to hit this wrap anyways (you would need to alloc a buffer near the usize::MAX).

chip-so (approved):

mooreoak (approved):

PR #716, opened 2026-07-01
Lazy merges

benz-spideroak:


crates/aranya-runtime/src/client/session.rs
```diff
@@ -476,12 +481,310 @@ where
...
+    impl Command for SeqCommand {
+        fn priority(&self) -> Priority {
+            if self.finalize {
+                return Priority::Finalize;
+            }
+            match self.prior {
+                Prior::None => Priority::Init,
+                Prior::Single(_) => {
+                    let id = self.id.as_bytes();
+                    let priority = u32::from(*id.last().unwrap());
```
mooreoak, line 622: Why are we computing the priority from the id? Is it just becouse we want a random id?

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -73,6 +75,30 @@ impl<SP: StorageProvider, PS: PolicyStore> Transaction<SP, PS> {
...
+            self.phead = None;
+            let segment = storage.write(p)?;
+            self.heads
+                .insert(segment.head_id(), segment.head_location()?);
+        }
+        Ok(())
+    }
+
+    /// The transaction's accumulated frontier (committed + received-but-uncommitted
+    /// tips), as `LocatedAddress`. Call `flush` first so every accumulated command
```
mooreoak: Is the flush something the caller is suposed to do or something this is suposed to do?

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -931,11 +999,36 @@ mod test {
...
+        storage
+            .get_heads()
+            .unwrap()
+            .iter()
+            .map(|la| la.max_cut)
+            .max()
+            .unwrap()
+    }
+
+    /// The sole committed graph head; panics if the graph is multi-head.
```
mooreoak, line 1093: wouldn't it be better to make this return a Option than panic?

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -931,11 +999,36 @@ mod test {
...
-        let p = storage.get_fact_perspective(head).unwrap();
+        // Read the committed fact cache, which is the braided state across all
+        // graph heads (lazy merges keep the graph multi-head after commit, so
+        // there is no single head segment whose facts represent the whole graph).
+        let p = storage.fact_cache().unwrap();
         p.query(name, &[]).unwrap()
     }
 
+    /// Number of committed graph heads.
+    fn head_count(storage: &impl Storage) -> usize {
```
mooreoak, line 1078: what was the reason to not implement these on storage?

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -228,35 +258,35 @@ impl Write for Writer {
         }
     }
 
-    fn head(&self) -> Result<Location, StorageError> {
-        if self.root.generation == 0 {
-            bug!("not initialized")
+    fn heads(&self) -> Result<HeadSet, StorageError> {
+        if self.root.heads == u64::MAX {
+            return Err(StorageError::NoSuchStorage);
```
mooreoak: This seems like the wrong error here? This happens because it has not been inited yet right?

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -228,35 +258,35 @@ impl Write for Writer {
...
         }
-        Ok(self.root.head)
+        self.fetch_owned(self.root.heads)
+    }
+
+    fn fact_cache(&self) -> Result<u64, StorageError> {
```
mooreoak: we should use a new type patter here instead of a u65 so `struct Cache(u64)` or something should be the return type.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -265,11 +295,15 @@ impl Write for Writer {
 /// Section of control data for the file
 #[derive(Debug, Serialize, Deserialize)]
 struct Root {
-    /// Incremented each commit
+    /// Incremented each commit.
     generation: u64,
-    /// Commit head.
-    head: Location,
-    /// Offset to write new item at.
+    /// Offset of the appended `HeadSet` record (`u64::MAX` before first
```
mooreoak: It would be better if these were option types or if we did not build a Root untill commit.

crates/aranya-runtime/src/storage/head_set.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /// Whether `location` is one of the heads.
+    ///
+    /// Linear scan: `Location` is not the sort key, so binary search does not apply.
+    pub fn contains(&self, location: Location) -> bool {
+        self.heads.iter().any(|h| h.location() == location)
+    }
+
+    /// Insert keeping the set sorted and deduplicated. Returns
```
mooreoak: above in contains we say it is not sorted and here we say it is.

mooreoak (commented):
This is a big PR and I can't say I have the background to fully understand it. Given it's size I want to make sure we get it landed soon so we don't have to deal with too many merge conflicts. I am not sure I see any of my comments as blocking, if Jonathan Digert is ok with this PR i'll aprouve too.

crates/aranya-runtime/src/client.rs
```diff
@@ -234,15 +236,34 @@ where
...
+        // A new command needs a single parent: collapse the head set, which may
+        // run a braid (hence the buffers and spill).
+        let heads = storage.get_heads()?.clone();
+        let head = transaction::collapse_heads::<_, PS, F, _>(
+            storage,
+            &mut self.policy_store,
+            heads,
+            buffers,
+            &make_spill,
+        )?;
```
jdygert-spok, line 281: Haven't we already braided and stored the fact result in the fact_cache? Why do we need to re-braid here when performing an action?

benz-spideroak: We do have the final fact result. We don't have the LCA or the facts for the intermediate merges. We could optimize this to only compute the LCA if there are only two heads and then use the fact cache. I think that can probably be a separate improvement.

crates/aranya-runtime/src/client/transaction.rs
benz-spideroak: The diff doesn't make it obvious, but this is all test code. So it's a test assertion.

crates/aranya-runtime/src/client/transaction.rs
benz-spideroak: This is only used by test code. No need to add to the Storage trait until prod code needs it.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
benz-spideroak: Done

crates/aranya-runtime/src/client.rs
jdygert-spok: Oh I see. I was hoping there was a way we wouldn't need to add the buffers and make_spill to this method api. I wonder if we could instead just write the final fact result as a multi-headed merge.

mooreoak: I think we don't want to write any merge until it is required to avoid the extra nodes in the graph that creates.

benz-spideroak: We talked about N way merges to avoid this issue, but as a follow up improvement.

mooreoak (dismissed):

crates/aranya-runtime/src/client.rs
jdygert-spok: IMO it's worth thinking a bit about how the N way merges would work for this and whether they'd avoid needing the new params. If so, it might be worth doing together to avoid API churn.

mooreoak: I am still not a fan of n-way merges as it is not clear to me how they compose but I am totally open to having my mind changed.

crates/aranya-runtime/src/storage/head_set.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// Iterate over heads.
+    pub fn iter(&self) -> impl Iterator<Item = LocatedAddress> + '_ {
+        self.heads.iter().copied()
+    }
+
+    /// Whether `location` is one of the heads.
+    ///
+    /// The set is sorted by `LocatedAddress` (command id first), not by
+    /// `Location`, so looking up a location is a linear scan rather than a
+    /// binary search.
```
jdygert-spok: If the purpose of the sorting is to dedup on `push`, we could sort by `Location` first so we can binary search in `contains` too.

crates/aranya-runtime/src/storage/head_set.rs
```diff
@@ -0,0 +1,141 @@
...
+    pub fn contains(&self, location: Location) -> bool {
+        self.heads.iter().any(|h| h.location() == location)
+    }
+
+    /// Insert keeping the set sorted and deduplicated. Returns
+    /// `StorageError::HeadSetFull` if at capacity.
+    pub fn push(&mut self, head: LocatedAddress) -> Result<(), StorageError> {
+        // Use partition_point on the deref'd slice to find insertion position.
+        let idx = self.heads.partition_point(|h| h < &head);
```
jdygert-spok: Why `partition_point` instead of `binary_search`?

crates/aranya-runtime/src/storage/head_set.rs
```diff
@@ -0,0 +1,141 @@
...
+use crate::{
+    StorageError,
+    storage::{LocatedAddress, Location, MAX_HEADS},
+};
+
+/// A bounded, sorted set of graph heads.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct HeadSet {
+    heads: heapless::Vec<LocatedAddress, MAX_HEADS>,
+}
```
jdygert-spok, line 19: Seems like this gets passed around on the stack a lot which isn't great since it's 24KB.

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -86,97 +116,61 @@ impl<SP: StorageProvider, PS: PolicyStore> Transaction<SP, PS> {
...
-            return Ok(false);
-        };
-        if original_head != storage.get_head()? {
-            return Err(ClientError::ConcurrentTransaction);
+            let Some(original_heads) = self.original_heads.take() else {
+                return Ok(false);
+            };
+            if &original_heads != storage.get_heads()? {
+                return Err(ClientError::ConcurrentTransaction);
+            }
```
jdygert-spok: I would like to find another approach instead of storing and comparing the whole head set here. Can we use the fact cache offset or something?

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -86,97 +116,61 @@ impl<SP: StorageProvider, PS: PolicyStore> Transaction<SP, PS> {
...
 
-        // Write out current perspective.
-        if let Some(p) = Option::take(&mut self.perspective) {
-            self.phead = None;
-            let segment = storage.write(p)?;
-            self.heads
-                .insert(segment.head_id(), segment.head_location()?);
-        }
+        // Persist any in-flight perspective into self.heads.
+        self.flush(provider)?;
```
jdygert-spok: Take the storage instead of provider as param, will let you avoid the re-fetching of the storage.

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -86,97 +116,61 @@ impl<SP: StorageProvider, PS: PolicyStore> Transaction<SP, PS> {
...
+        // Rebuild the fact cache for this head set.
+        let head_locs: heapless::Vec<Location, MAX_HEADS> =
+            head_set.iter().map(LocatedAddress::location).collect();
+        let fact_cache = if head_locs.len() == 1 {
+            // Single head: reuse that head's fact index.
+            storage.get_segment(head_locs[0])?.facts()?
+        } else {
+            // Multi-head: braid the heads into a merged fact index.
+            // TODO(multi-policy): if heads can have differing policies, fold choose_policy
+            // over the head set instead of taking head 0's policy.
```
jdygert-spok, line 157: This is sort of a regression but probably OK since we don't really support it yet.

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -86,97 +116,61 @@ impl<SP: StorageProvider, PS: PolicyStore> Transaction<SP, PS> {
...
-                let mut perspective = storage.new_merge_perspective(
-                    left_loc,
-                    right_loc,
-                    last_common_ancestor,
-                    policy_id,
-                    braid,
-                )?;
-                perspective.add_command(&command)?;
+        // Build the committed head set from the live tips.
+        let mut head_set = HeadSet::default();
```
jdygert-spok, line 140: Seems like we could avoid constructing this here by using iterators in more places.

crates/aranya-runtime/src/storage/head_set.rs
```diff
@@ -0,0 +1,141 @@
...
+    pub fn push(&mut self, head: LocatedAddress) -> Result<(), StorageError> {
+        // Use partition_point on the deref'd slice to find insertion position.
+        let idx = self.heads.partition_point(|h| h < &head);
+        // Skip duplicate.
+        if self.heads.get(idx).is_some_and(|h| h == &head) {
+            return Ok(());
+        }
+        self.heads
+            .insert(idx, head)
+            .map_err(|_| StorageError::HeadSetFull(MAX_HEADS))?;
```
jdygert-spok: We should think about how to handle a full head set.

crates/aranya-runtime/src/storage/linear/testing.rs
```diff
@@ -54,7 +54,8 @@ struct Shared {
 }
 
 pub struct Writer {
-    head: Mutex<Option<Location>>,
+    heads: Mutex<Option<HeadSet>>,
+    fact_cache: Mutex<Option<io::FactCacheOffset>>,
```
jdygert-spok: nit: this should probably be `Mutex<Option<(HeadSet, FactCacheOffset)>>` (mayybe with a struct instead of tuple for readability)

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -519,6 +521,80 @@ where
...
+{
+    let mut q: VecDeque<(CmdId, Location)> =
+        heads.iter().map(|la| (la.id, la.location())).collect();
+
+    // The multi-head state being collapsed was produced by a `commit` that
+    // already braided these same heads and emitted their effects in converged
+    // order. Re-braiding the identical head set yields no new effects (braid
+    // order is a stable global sort by `(priority, id)`), so emitting again
+    // would only duplicate what `commit` delivered.
+    let mut null = NullSink;
```
jdygert-spok, line 559: Worth noting this is a change in behavior. We will no longer observe the effects of intermediate merges. I think that's probably a good thing anyway?

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -519,6 +521,80 @@ where
...
+    while q.len() > 1 {
+        let (left_id, mut left_loc) = q.pop_front().assume("len > 1")?;
+        let (right_id, mut right_loc) = q.pop_front().assume("len > 1")?;
```
jdygert-spok: nit: avoid these `assume`s

```rs
while let Some((left_id, mut left_loc)) = heads.pop_front() {
    let Some((right_id, mut right_loc)) = heads.pop_front() else {
        return Ok(left_loc);
    };
    
    ...
}
bug!("head set was empty");
```

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -519,6 +521,80 @@ where
...
+        let merge_ids = MergeIds::new(
+            Address {
+                id: left_id,
+                max_cut: left_loc.max_cut,
+            },
+            Address {
+                id: right_id,
+                max_cut: right_loc.max_cut,
+            },
+        )
```
jdygert-spok: Not relevant to this PR but noticed this and the queue could probably benefit from using `LocatedAddress` to clean this function up a bit.

crates/aranya-runtime/src/client/braiding.rs
```diff
@@ -48,6 +48,23 @@ pub(super) fn last_common_ancestor<S: Storage>(
...
+pub(super) fn last_common_ancestor<S: Storage>(
+    storage: &mut S,
+    heads: &[Location],
+) -> Result<Location, ClientError> {
+    let (first, rest) = heads.split_first().assume("braid heads non-empty")?;
+    let mut lca = *first;
+    for &h in rest {
+        lca = lca_pair(storage, lca, h)?;
+    }
+    Ok(lca)
```
jdygert-spok: nit: use `try_fold`

crates/aranya-runtime/src/storage/mod.rs
```diff
@@ -651,7 +667,14 @@ pub trait Storage {
...
         buffer: &mut TraversalBuffer,
     ) -> Result<Option<Location>, StorageError> {
-        self.get_location_from(self.get_head()?, address, buffer)
+        // The graph may be multi-head (lazy merges). Search from each head until
+        // the command is found.
+        for head in self.get_heads()?.iter() {
+            if let Some(found) = self.get_location_from(head.location(), address, buffer)? {
+                return Ok(Some(found));
+            }
+        }
```
jdygert-spok, line 731: Should we make this a single search where the queue starts as the head set?

crates/aranya-runtime/src/client/transaction.rs
jdygert-spok: I guess not the fact cache offset, but I think the head set offset would work.

crates/aranya-runtime/src/client/transaction.rs
mooreoak: Also I think we will need to reconsider how this should work. The last finalized policy should alway be used so we really want to move the policy on to the perspective or something. (This might be one policy at the origen perspective and a different in the braid thou)

crates/aranya-runtime/src/client/transaction.rs
mooreoak: Yes I think it is probably good. I can see cases where you want to be able to know intermediate states other clients were in but that is an edge case we may never actually need so let's do it like this for now and if we ever need it we can figure out a different API for that kind of use.

crates/aranya-runtime/src/client/braiding.rs
benz-spideroak: Done.

crates/aranya-runtime/src/client/transaction.rs
benz-spideroak: Done.

crates/aranya-runtime/src/storage/linear/testing.rs
benz-spideroak: Done.

crates/aranya-runtime/src/client/transaction.rs
benz-spideroak: Done.

crates/aranya-runtime/src/storage/head_set.rs
benz-spideroak: Updated to use binary_search.

crates/aranya-runtime/src/storage/head_set.rs
benz-spideroak: This method isn't actually needed. I deleted it. The reason for sorting by `id` is that it's consistent across peers. `collapse_heads` uses the sorted order to pair up merges. This ensures the same set of commands produces the same merges.  This might not be the case if we sorted by `Location`.

crates/aranya-runtime/src/client/transaction.rs
benz-spideroak: Before I did it this way you'd see some effects multiple times which makes the tests tricky. I believe you should still see the same effects in the same order, but only once.

crates/aranya-runtime/src/client/transaction.rs
benz-spideroak: I used the offset.

crates/aranya-runtime/src/storage/head_set.rs
benz-spideroak: Switched to an alloc:Vec to follow the other similar changes.

crates/aranya-runtime/src/storage/head_set.rs
benz-spideroak: This can't happen anymore.

crates/aranya-runtime/src/client/transaction.rs
benz-spideroak: Why are you wanting to avoid constructing it?

crates/aranya-runtime/src/storage/mod.rs
benz-spideroak: Done

crates/aranya-runtime/src/client/transaction.rs
jdygert-spok: Just avoiding some unnecessary work and previously stack space + copying it around.

crates/aranya-runtime/src/testing/dsl.rs
```diff
@@ -1419,14 +1538,21 @@ fn sync<SP: StorageProvider>(
...
+        request_trx.flush(storage)?;
+
+        // Advance the requester's have-set from the transaction's accumulated
+        // frontier. `PeerCache::add_command` prunes ancestors via `is_ancestor`
+        // over the now-written segments. This replaces the old `update_heads`
+        // (which only saw committed state) and is what lets progressive fetch
+        // advance across exchanges while the transaction stays open.
+        for head in request_trx.in_flight_heads() {
+            request_cache.add_command(storage, head, &mut rt_buffers.traversal.primary)?;
+        }
```
jdygert-spok: This seems funky to me. We are recording in our cache the location of commands that we haven't yet committed. If we abort the transaction we will end up with "stale" locations.

benz-spideroak: I updated the API so only committed heads can be added to the cache.

crates/aranya-runtime/src/client/transaction.rs
benz-spideroak: It doesn't seem worth it anymore now that it's less stack space.

mooreoak (approved):
I aprouve conditional on Digert's aproval.

jdygert-spok (approved):

crates/aranya-runtime/src/client.rs
```diff
@@ -222,16 +209,76 @@ where
...
+    /// A single-head graph advertises its head. A multi-head graph advertises
+    /// the deterministic merge its head set would collapse to, computed
+    /// without writing anything to storage: a peer holding the same head set
+    /// computes the same address, and a peer that already collapsed these
+    /// heads (by committing a command on top) has this exact command in its
+    /// graph. See [`Self::should_sync_on_hello`] for the receiving side.
+    pub fn hello_head(&mut self, graph_id: GraphId) -> Result<Address, ClientError> {
+        let storage = &*self.provider.get_storage(graph_id)?;
+        let heads = storage.get_heads()?;
+        transaction::synthetic_head(storage, &self.policy_store, heads)
```
jdygert-spok, line 223: I just noticed we compute the synthetic head on request, wouldn't we rather just store it? We already do the work of braiding to create the lazy fact cache.

PR #717, opened 2026-07-02
Optimize fsync

benz-spideroak:
Optimize fsync.


crates/aranya-libc/src/sys/vxworks.rs
```diff
@@ -142,3 +142,10 @@ pub fn fsync(fd: BorrowedFd<'_>) -> Result<(), Errno> {
     let ret = unsafe { libc::fsync(fd.as_raw()) };
     if ret < 0 { Err(errno()) } else { Ok(()) }
 }
+
```
mooreoak, line 145: No really and thing to do in this PR but I wonder if we should retire VxWorks. Do we even have tests for it?

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -269,6 +271,57 @@ impl<SP: StorageProvider, PS: PolicyStore> Transaction<SP, PS> {
...
+        // Run the action, reverting its partial effects on failure so
+        // one bad action doesn't poison the rest of the batch.
+        sink.begin();
+        let checkpoint = perspective.checkpoint();
+        if let Err(e) = policy.call_action(action, perspective, sink, ActionPlacement::OnGraph) {
+            perspective.revert(checkpoint)?;
+            sink.rollback();
+            return Err(e.into());
+        }
+        sink.commit();
```
mooreoak: should there be a checkpoint commit before this line or is that handled by the call_action?

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -163,42 +175,84 @@ const ROOT_B: i64 = PAGE * 2;
...
 
+/// Returns the other root slot, for ping-ponging between the two.
+fn other_root(slot: i64) -> i64 {
+    if slot == ROOT_A { ROOT_B } else { ROOT_A }
+}
+
+/// Granularity by which the file is grown ahead of the write
+/// frontier. Preallocating in large chunks keeps the file size
+/// stable across appends so `fdatasync` avoids the extra
+/// inode-metadata journal commit that a growing file forces.
```
mooreoak, line 187: It also improves locality on disk which makes our IO more efficient.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -163,42 +175,84 @@ const ROOT_B: i64 = PAGE * 2;
...
+        if end <= self.alloc_end {
+            return Ok(());
+        }
+        let mut new_end = self.alloc_end;
+        while new_end < end {
+            new_end = new_end
+                .checked_add(PREALLOC_CHUNK)
+                .assume("preallocation size fits in `i64`")?;
+        }
+        self.file.fallocate(0, new_end)?;
```
mooreoak, line 256: if we don't do a synce() after the falloacte we are not garenteed that our full allocated region will be preserved. If we crash after part was written with fdadasync, the written pages will be durable but the extra allocations may be lost.

If we want to get even a little more performance we could fallocate the whole DB only at creation and then manually zero fill it. This removes the need for persisting at in the file system the meta data about the pages going from unwritten and zero backed, to written and disk backed. I don't fully understand the A/B design you have here so I don't know how this approach would interact with that.

mooreoak (commented):
Ok so I think this design works like this. We have two roots for the DB so we can update the bounds of the DB in a single atomic write. This assumes the root pointer is less that or equal to the size of a single atomic write to the underlying storage. (This seems a reasonable assumption.) This seems like an ok design. Another way to do this is to not assume atomic writes and just have the A/B roots, but include a sequence number and checksum on the roots. At boot up you take whichever root has both a valid checksum and the largest sequence number. Not requesting a change just providing this as information incase it becomes useful.

Dose this design only support one transaction at a time? Dose a transaction block other writes? One of the goals of the original design was that slow syncs would not block local actions. Is that still true?

crates/aranya-libc/src/sys/vxworks.rs
jdygert-spok: We dropped our tests for it when migrating back to public github and it has since rotted at least missing some of the modifications we've made in aranya-libc.

jdygert-spok:
> just have the A/B roots, but include a sequence number and checksum on the roots. At boot up you take whichever root has both a valid checksum and the largest sequence number

If I'm understanding you right, this is already how it works before/after this PR.

jdygert-spok:
> Dose this design only support one transaction at a time? Dose a transaction block other writes? One of the goals of the original design was that slow syncs would not block local actions. Is that still true?

The current design only supports one transaction at a time.

crates/aranya-libc/src/sys/vxworks.rs
mooreoak: We should just remove it for sure then.

crates/aranya-runtime/src/client/transaction.rs
benz-spideroak: It's handled by call_action.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
benz-spideroak: Added a sync

benz-spideroak:
> > Dose this design only support one transaction at a time? Dose a transaction block other writes? One of the goals of the original design was that slow syncs would not block local actions. Is that still true?
> 
> The current design only supports one transaction at a time.

I added the test test_action_aborts_open_transaction to confirm the behavior. A transaction does not block other writes, but because another commit advances the head a prior transaction will fail if you try to commit. We don't enforce it, but you probably want to only have one transaction at a time. This isn't a change in this PR. This is what the behavior has been. You could take the actions from the sync and retry them in a new transaction after the first one failed.

mooreoak:
> > > Dose this design only support one transaction at a time? Dose a transaction block other writes? One of the goals of the original design was that slow syncs would not block local actions. Is that still true?
> > 
> > 
> > The current design only supports one transaction at a time.
> 
> I added the test test_action_aborts_open_transaction to confirm the behavior. A transaction does not block other writes, but because another commit advances the head a prior transaction will fail if you try to commit. We don't enforce it, but you probably want to only have one transaction at a time. This isn't a change in this PR. This is what the behavior has been. You could take the actions from the sync and retry them in a new transaction after the first one failed.

If I recall correctly if the head changed in the original design for transactions we  we and a new merge head and recommit. I think something like that could be done her. The initial motivation for the transaction was to allow a slow stink to happen while local actions still happen. There is probably also value it allowing more then one sync to happen at a time but I am happy to skip that for now as that has to handle the case if the same command being received in more than one transaction.

(Hmm I suppose with lazy merges we open the issue that we might create a merge on demand that has also been accepted in a transaction. We should think about that case.)

benz-spideroak:
> If I recall correctly if the head changed in the original design for transactions we we and a new merge head and recommit. I think something like that could be done her. The initial motivation for the transaction was to allow a slow stink to happen while local actions still happen. There is probably also value it allowing more then one sync to happen at a time but I am happy to skip that for now as that has to handle the case if the same command being received in more than one transaction.
> 
> (Hmm I suppose with lazy merges we open the issue that we might create a merge on demand that has also been accepted in a transaction. We should think about that case.)

I believe that will work if this code is removed. https://github.com/aranya-project/aranya-core/blob/ebdea6bb2137a44c5df1c74ac2ef5550291aae26/crates/aranya-runtime/src/client/transaction.rs#L86-L91

@jdygert-spok You added that check. Do we need it?

jdygert-spok:
I added that check to prevent committing simultaneous syncs, since we haven't tested or thought through what happens when a command is stored in multiple locations. If we modify this check to allow the case of performing an action, that's probably OK. In theory you could still end up with the same command stored twice that way, if you have a "forgetful" actor, but that's probably not a big issue?

crates/aranya-libc/src/sys/macos.rs
```diff
@@ -14,6 +14,13 @@ pub fn fallocate(_fd: BorrowedFd<'_>, _mode: c_int, _off: i64, _len: i64) -> Res
     Ok(())
 }
 
+/// See `fdatasync(2)`.
+///
+/// macOS has no distinct `fdatasync`, so fall back to `fsync`.
+pub fn fdatasync(fd: BorrowedFd<'_>) -> Result<(), Errno> {
+    super::unix::fsync(fd)
```
jdygert-spok: triggering unnecessary qualification lint:

```suggestion
    fsync(fd)
```

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -269,6 +271,57 @@ impl<SP: StorageProvider, PS: PolicyStore> Transaction<SP, PS> {
...
+        }
+
+        // Ensure there's an open perspective, anchored at the current
+        // transaction head, to run the action against.
+        if self.perspective.is_none() {
+            let head = storage.get_head()?;
+            self.phead = Some(storage.get_head_address()?.id);
+            self.perspective = Some(storage.get_linear_perspective(head)?);
+        }
+        let perspective = self.perspective.as_mut().assume("trx has perspective")?;
```
jdygert-spok: nit: use `Option::insert` to avoid `assume`

```suggestion
        let perspective = match &mut self.perspective {
            Some(p) => p,
            None => {
                let head = storage.get_head()?;
                self.phead = Some(storage.get_head_address()?.id);
                self.perspective
                    .insert(storage.get_linear_perspective(head)?)
            }
        };
```

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -269,6 +271,57 @@ impl<SP: StorageProvider, PS: PolicyStore> Transaction<SP, PS> {
...
+            perspective.revert(checkpoint)?;
+            sink.rollback();
+            return Err(e.into());
+        }
+        sink.commit();
+
+        // Advance the perspective head to the action's last command.
+        if let Prior::Single(addr) = perspective.head_address()? {
+            self.phead = Some(addr.id);
+        }
```
jdygert-spok: Let's handle the case where this isn't single. Probably with a `bug!`.

mooreoak (dismissed):

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -163,42 +175,84 @@ const ROOT_B: i64 = PAGE * 2;
...
         let file = File { fd: Arc::new(fd) };
-        // Preallocate so we can start appending from FREE_START
-        // forward.
-        file.fallocate(0, FREE_START)?;
+        // Preallocate the control region plus a first data chunk so
+        // we can start appending from FREE_START forward without
+        // extending the file size on every append.
+        let alloc_end = FREE_START
+            .checked_add(PREALLOC_CHUNK)
+            .assume("initial preallocation fits in `i64`")?;
```
jdygert-spok: nit: move arithmetic check to compile time.

```suggestion
        let alloc_end = const { (FREE_START + PREALLOC_CHUNK) as i64 };
```

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -149,6 +149,18 @@ impl IoManager for FileManager {
...
     file: File,
     root: Root,
+    /// End of the region preallocated (and size-extended) via
+    /// `fallocate`. Appends stay within this bound so `fdatasync`
+    /// doesn't have to flush a file-size change on every commit.
+    alloc_end: i64,
+    /// Root slot (`ROOT_A`/`ROOT_B`) to write on the next commit.
+    /// We ping-pong between the two so the previously committed root
+    /// stays intact until the new one is durable.
+    next_root: i64,
```
jdygert-spok, line 160: I feel like there was a reason I decided to write both roots rather than just ping ponging, but now I don't know what it could have been :thinking:

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -163,42 +175,84 @@ const ROOT_B: i64 = PAGE * 2;
...
             (Err(e), Err(_)) => return Err(e),
         };
 
-        // Write other side if needed (corrupted or outdated)
-        if let Some(offset) = overwrite {
-            file.dump(offset, &root)?;
-        }
+        // Everything up to the write frontier is known to be
+        // allocated; `ensure_capacity` grows from here as needed.
+        let alloc_end = root.free_offset;
```
jdygert-spok, line 232: Should we use the file size for this?

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -163,42 +175,84 @@ const ROOT_B: i64 = PAGE * 2;
...
+    /// this bound so their `fdatasync` doesn't flush a size change.
+    fn ensure_capacity(&mut self, end: i64) -> Result<(), StorageError> {
+        if end <= self.alloc_end {
+            return Ok(());
+        }
+        let mut new_end = self.alloc_end;
+        while new_end < end {
+            new_end = new_end
+                .checked_add(PREALLOC_CHUNK)
+                .assume("preallocation size fits in `i64`")?;
```
jdygert-spok, line 254: Thoughts on doubling every time instead?

crates/aranya-libc/src/sys/macos.rs
benz-spideroak: Done

crates/aranya-runtime/src/client/transaction.rs
benz-spideroak: Updated.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
benz-spideroak: Done.

crates/aranya-runtime/src/client/transaction.rs
benz-spideroak: Done.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
benz-spideroak: I looked into what other dbs do and fixed size segments seem pretty standard. This is already a small amount of the run time so I doubt it would make much of a performance difference. We could go with bigger chunks at some points if necessary.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
benz-spideroak: Seems unnecessary. I don't think we currently have a way to get file size on the `File` wrapper. Doing that would potentially save one fallocate. Not much of a difference.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
jdygert-spok: It would also let us decrease the initial size. This PR makes the file start out at 4MB, but looking at a graph from a demo, it is around 40KB, 100x less.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -247,15 +301,40 @@ impl Write for Writer {
...
     }
 
     fn commit(&mut self, head: Location) -> Result<(), StorageError> {
+        // Barrier 1: ensure the appended data is durable before the
+        // root that references it, so a crash can't leave the head
+        // pointing at data that never reached disk.
+        if self.data_dirty {
+            self.file.sync()?;
+            self.data_dirty = false;
+        }
```
jdygert-spok, line 370: It's not a big deal to keep it but I noticed the dirty flag isn't really necessary. `commit` is only called when there's a new segment to advance the head to, so this will always be dirty.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
mooreoak: If we want it to be more efficient on disk we can just make the chunks smaller; but any growing is actually kind of against the core idea that you commit to how much max storage can be used and just allocate that all upfront when you create the DB.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
mooreoak: It would handle corruption better but that might not be the reason.

mooreoak (dismissed):

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -163,42 +175,82 @@ const ROOT_B: i64 = PAGE * 2;
...
     fn create(fd: OwnedFd) -> Result<Self, StorageError> {
         let file = File { fd: Arc::new(fd) };
-        // Preallocate so we can start appending from FREE_START
-        // forward.
-        file.fallocate(0, FREE_START)?;
+        // Preallocate the control region plus a first data chunk so
+        // we can start appending from FREE_START forward without
+        // extending the file size on every append.
+        let alloc_end = const { FREE_START + PREALLOC_CHUNK };
+        file.fallocate(0, alloc_end)?;
```
djohnson-spideroak, line 201: How does this interact with existing files?

jdygert-spok (dismissed):

djohnson-spideroak (dismissed):

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -265,6 +265,95 @@ impl<SP: StorageProvider, PS: PolicyStore> Transaction<SP, PS> {
...
+            self.original_heads_offset = Some(storage.heads_offset()?);
+        }
+
+        // Ensure there's an open perspective to run the action against.
+        // A new command needs a single parent, so anchor at the
+        // collapse of the transaction's tips, which may materialize a
+        // merge (as `ClientState::action` does for the committed set).
+        let perspective = match &mut self.perspective {
+            Some(p) => p,
+            None => {
```
jdygert-spok: This seems odd, if we do an action on a transaction after a sync, it will be a child of the last synced command rather than the merge of all heads.

mooreoak: I agree that that is not what we want

benz-spideroak: Good catch. Fixed.

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -265,6 +265,103 @@ impl<SP: StorageProvider, PS: PolicyStore> Transaction<SP, PS> {
...
+                let mut tips = HeadSet::default();
+                for (id, loc) in &self.heads {
+                    tips.push(LocatedAddress {
+                        id: *id,
+                        segment: loc.segment,
+                        max_cut: loc.max_cut,
+                    });
+                }
+                let head = collapse_heads::<_, PS, F, _>(
```
jdygert-spok: `collapse_heads` uses a `NullSink` because it assumes we already saw the effects during a commit, but I don't believe that's true here.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -253,6 +254,49 @@ impl Writer {
...
+        let item = builder(off);
+        let bytes = postcard::to_allocvec(&item).map_err(|err| {
+            error!(?err, "append");
+            StorageError::IoError
+        })?;
+        // Ensure the file is grown ahead of this write so appending
+        // it doesn't change the file size (keeping `fdatasync` cheap).
+        let len = i64::try_from(bytes.len()).assume("serialized len fits in `i64`")?;
+        let end = offset
+            .checked_add(4)
```
djohnson-spideroak: Why 4 bytes?

crates/aranya-runtime/src/client/transaction.rs
benz-spideroak: I added a fix for this. It does feel complicated though. I wish there were a simpler way to handle this. Maybe get rid of the NullSink and give test effects a unique ID so we can ignore duplicates in tests.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
benz-spideroak: I added a const to make it clear.

crates/aranya-libc/src/sys/linux.rs
```diff
@@ -16,6 +16,13 @@ pub fn fallocate(fd: BorrowedFd<'_>, mode: c_int, off: i64, len: i64) -> Result<
     if ret < 0 { Err(errno()) } else { Ok(()) }
 }
 
+/// See `fdatasync(2)`.
+pub fn fdatasync(fd: BorrowedFd<'_>) -> Result<(), Errno> {
+    // SAFETY: FFI call, no invariants.
+    let ret = unsafe { libc::fdatasync(fd.fd) };
+    if ret < 0 { Err(errno()) } else { Ok(()) }
```
mooreoak, line 23: I don't want to block on this comment but it seems like `errno()` should be `errno(ret)` so we don't swallow what the acutal error was.

mooreoak:
I am confused why there is all this stuff about actions and braiding in the fsync PR. Are these issues brought up by implementing fsync or are they unrelated work that got tacked on to this PR?

benz-spideroak:
> I am confused why there is all this stuff about actions and braiding in the fsync PR. Are these issues brought up by implementing fsync or are they unrelated work that got tacked on to this PR?

Previously we had transactions for commands added through sync, but when adding commands with actions each one was written individually. This PR allows you to add actions to transactions. Bundling the writes improves performance. This gets complicated if you have multiple heads. Specifically around the interaction between multiple heads and Sinks. We were previously able to know that collapse_heads didn't need to emit effects. That's not always true anymore with the changes in this PR.

mooreoak:
> Previously we had transactions for commands added through sync, but when adding commands with actions each one was written individually. This PR allows you to add actions to transactions. Bundling the writes improves performance.

It seems like this PR should be renamed or split in two. This sounds like this is updating actions to allow actions in transactions, materialization of merge heads on demand, and adding fdatasync etc. It feels like this is two or three PRs. At least the name has become misleading.

benz-spideroak:
> > Previously we had transactions for commands added through sync, but when adding commands with actions each one was written individually. This PR allows you to add actions to transactions. Bundling the writes improves performance.
> 
> It seems like this PR should be renamed or split in two. This sounds like this is updating actions to allow actions in transactions, materialization of merge heads on demand, and adding fdatasync etc. It feels like this is two or three PRs. At least the name has become misleading.

I removed the actions in transactions. The PR is now much simpler. And I believe the existing code has previously been reviewed.

mooreoak (approved):
It would be better to return errors than use assume but the cases where I see it seem like they should not happen in practice; it just makes it harder to review the code for panic safety.

crates/aranya-libc/src/sys/linux.rs
jdygert-spok: "On error, -1 is returned, and errno is set appropriately."

jdygert-spok (approved):

crates/aranya-libc/src/sys/linux.rs
mooreoak: Ok so we would have to read errno? it still seems better to capture the value here are return it then to expose that interface to the caller.

jdygert-spok: `libc::fdatasync` sets errno to the error value, then we read it with `errno()` and return it to the caller in the `Result`.

mooreoak: O I thought errorno() was creating a type. Ok that looks good I misunderstood. Thanks for the explaining.

PR #722, opened 2026-07-07
runtime: fix convergence map braid drop

jdygert-spok:


crates/aranya-runtime/src/client/braiding.rs
```diff
@@ -260,6 +260,11 @@ where
                 );
                 continue 'location;
             }
+            // Convergence check (incremental BFS, O(1) amortized).
```
mooreoak, line 263: Dos this work here? I thought it had to be on line 260 in the check `location.max_cut <= lca.max_cut`

mooreoak (approved):

djohnson-spideroak (approved):

PR #726, opened 2026-07-07
Remove vm span dependency

chip-so:
This is currently just removing the dependency. It replaces maps/sets with vecs in a number of places to simply things. I'm not super pleased with some of it. The things in `interface.rs` feel like they should be in the compiler, but `NamedMap` is in the module crate. I suppose that means `NamedMap` should probably be moved into the compiler or AST.

Next steps are to make something like `NamedMap` for the module/VM side that doesn't embed Span information, and replace every `BTreeMap<Identifier, ...>` with that.


mooreoak (dismissed):
What I learned in this review is that we have a lot of tests.

chip-so:
So there's three main things happening in this PR:

- Data structures have been split into versions with spans (mostly unchanged things from the AST, and a few new things in `aranya_policy_module::interface` to support the interface generator) and those without (the `FooDef` structs in `aranya_policy_module`).
- The module format itself has been simplified to remove `NamedMap` and most `BTreeMap`s in favor of `Vec`s of spanless module types.
- The VM now uses a new `AutoMap` type, which is designed to work like `NamedMap` but with a generic key type (and it's a `BTreeMap` inside). I think `NamedMap` can be subsumed by it if I add serialization support, but I haven't done that here.

mooreoak (approved):

mooreoak (approved):

mooreoak (dismissed):

apetkov-so (dismissed):

crates/aranya-policy-module/src/automap.rs
jdygert-spok, line 1: We could change the named map to just use `&str` for the name and continue to use that instead.

IDK whether switching from insertion-order-preserving to sorted-order was intentional or not.

chip-so: > We could change the named map to just use &str for the name and continue to use that instead.

Kind of. The problem is the `AlreadyExists` error provides the `Ident` captured by the `Named` trait to the compiler's error reporting routines, so they can locate the error.

> IDK whether switching from insertion-order-preserving to sorted-order was intentional or not.

Not an intentional difference, but I did intentionally use the built-in `BTreeMap` because AFAIK nothing in the VM depends on order.

mooreoak (dismissed):

jdygert-spok:
Could wait for a followup but we can move `AutoMap` to the vm and `NamedMap` to the compiler.

jdygert-spok (dismissed):
I don't love that a few more places have regressed to linear lookups, but we already have a bunch and they shouldn't be large in practice :shrug:

mooreoak:
> I don't love that a few more places have regressed to linear lookups, but we already have a bunch and they shouldn't be large in practice 🤷

sometimes linear is faster that binary for small scans, but it really depends on your hardware.

jdygert-spok (approved):

mooreoak (approved):

PR #730, opened 2026-07-15
test: add linear storage tests

djohnson-spideroak:


crates/aranya-runtime/src/storage/linear/libc/imp.rs
```diff
@@ -410,3 +410,92 @@ impl File {
...
+    }
+
+    #[test]
+    fn test_head_before_commit_is_bug() {
+        let dir = tempfile::tempdir().unwrap();
+        let mut fm = FileManager::new(dir.path()).unwrap();
+        let writer = fm.create(graph_id("uninit")).unwrap();
+        let res = std::panic::catch_unwind(core::panic::AssertUnwindSafe(|| writer.head()));
+        assert!(matches!(res, Err(_) | Ok(Err(StorageError::Bug(_)))));
+    }
```
djohnson-spideroak, line 486: We probably want to update the code for "fn head" here (in the libc writer). It returns a bug if head is not initialized, but this is a path you could easily hit during normal usage. We should return a normal error.

crates/aranya-runtime/src/storage/linear/mod.rs
```diff
@@ -1281,4 +1284,338 @@ mod test {
...
+    fn test_linear_perspective_mid_init_segment() {
+        let mut provider = new_provider();
+        let mut p = provider.new_perspective(PolicyId::new(0));
+        let c0 = cmd("mid", Prior::None);
+        p.add_command(&c0).unwrap();
+        // Updates carried by the second command: insert then delete the same
+        // fact, plus a delete for a name that never existed.
+        p.insert("n".into(), key("k"), val(b"v")).unwrap();
+        p.delete("n".into(), key("k")).unwrap();
+        p.delete("m".into(), key("k")).unwrap();
```
mooreoak, line 1400: I thought this was should to fail. Maybe not at this level of abstraction?

crates/aranya-runtime/src/storage/linear/mod.rs
```diff
@@ -1281,4 +1284,338 @@ mod test {
...
+        let head = storage.get_head().unwrap();
+        let head_addr = storage.get_head_address().unwrap();
+
+        // Left branch keeps the graph's policy (0).
+        let mut pa = storage.get_linear_perspective(head).unwrap();
+        let ca = cmd("merge.a", Prior::Single(head_addr));
+        pa.add_command(&ca).unwrap();
+        let seg_a = storage.write(pa).unwrap();
+
+        // Right branch is hand-built with a different policy (1).
```
mooreoak, line 1448: This is a reasonable test of an out dated API we should only allow policy changes at finalized  commands. Not requesting a change to this test but it's something we should track as needing a fix in the future.

crates/aranya-runtime/src/storage/linear/mod.rs
```diff
@@ -1281,4 +1284,338 @@ mod test {
...
+        assert_eq!(mp.policy(), PolicyId::new(1));
+    }
+
+    #[test]
+    fn test_query_prefix_skips_tombstones() {
+        let mut provider = new_provider();
+        let mut p = provider.new_perspective(PolicyId::new(0));
+        let c0 = cmd("tomb", Prior::None);
+        p.add_command(&c0).unwrap();
+        p.insert("n".into(), key("k1"), val(b"v1")).unwrap();
```
mooreoak, line 1485: I am confused by all this inserting in to the fact DB is the emulating the execution of policy?

crates/aranya-runtime/src/storage/linear/mod.rs
```diff
@@ -1281,4 +1284,338 @@ mod test {
...
+        assert_eq!(
+            p.head_address().unwrap(),
+            Prior::Single(c0.address().unwrap())
+        );
+
+        // A checkpoint beyond the command history is a bug.
+        let res = std::panic::catch_unwind(core::panic::AssertUnwindSafe(|| {
+            p.revert(Checkpoint { index: 99 })
+        }));
+        assert!(matches!(res, Err(_) | Ok(Err(StorageError::Bug(_)))));
```
mooreoak, line 1588: This is a bug and not an error because it uses a forged checkpoint?

I guess this is not a question about the test and this PR but about the code under test.

mooreoak (changes requested):
Over all looks good. It could use some more comments both explaining the goal of a test and the theory of how the tests work. 

Ben should really review the TerversalQueue stuff as I don't really know what to expect therte.

crates/aranya-runtime/src/storage/linear/mod.rs
djohnson-spideroak: Yeah, this is the relevant code:

```
        if checkpoint.index > self.commands.len() {
            bug!(
                "A checkpoint's index should always be less than or equal to the length of a perspective's command history!"
            );
        }
```

It seems like a fair place to hit Bug. The only way to hit it would be to use the checkpoint incorrectly.

crates/aranya-runtime/src/storage/linear/mod.rs
djohnson-spideroak: Yes, it is setting up the fact DB to test specific branches of the perspective.

crates/aranya-runtime/src/storage/linear/libc/imp.rs
jdygert-spok: Without directly using the file manager, I think the only way to trigger this via storage would be to manually craft a file that has a root with both generation 0 and a valid checksum.

IMO that puts this in the same class as the `bug!` in `revert`. It's something called by us but not by our users.

PR #731, opened 2026-07-16
Avoid exposing AST VType to client

apetkov-so:
Use associated ActionReturn associated type instead.

### DIFF
--- a/crates/aranya-runtime/src/client.rs
+++ b/crates/aranya-runtime/src/client.rs
@@ -247,7 +247,7 @@ where
         graph_id: GraphId,
         sink: &mut impl Sink<PS::Effect>,
         action: <PS::Policy as Policy>::Action<'_>,
-    ) -> Result<(), ClientError> {
+    ) -> Result<<PS::Policy as Policy>::ActionReturn, ClientError> {
         let storage = self.provider.get_storage(graph_id)?;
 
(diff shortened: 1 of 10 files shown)


crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -490,6 +490,7 @@ impl<CE> VmPolicy<CE> {
 
 impl<CE: aranya_crypto::Engine> Policy for VmPolicy<CE> {
     type Action<'a> = VmAction<'a>;
+    type ActionReturn = ();
```
jdygert-spok, line 493: This change doesn't do anything if we're just using `()`.

crates/aranya-runtime/src/client.rs
```diff
@@ -262,11 +262,11 @@ where
 
         sink.begin();
         match policy.call_action(action, &mut perspective, sink, ActionPlacement::OnGraph) {
-            Ok(()) => {
+            Ok(ret) => {
```
jdygert-spok, line 265: The issue I noticed with my result approach in #695 was that I couldn't match on it here to determine whether to commit or roll back. I updated that to use `Result<Result<(), Self::ActionError>, PolicyError>` instead. It does work, it just feels a bit clunky.

PR #733, opened 2026-07-22
Add a recall block so the example compiles

chip-so:
A little thing I found while testing. Compiler got stricter, and this got broken. :)

### DIFF
--- a/crates/aranya-policy-runner/examples/policy.md
+++ b/crates/aranya-policy-runner/examples/policy.md
@@ -170,6 +170,9 @@ command AddUser {
             }
         }
     }
+
+    recall test_fail(err string) {
+    }
 }
 ```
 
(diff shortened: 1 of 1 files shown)


apetkov-so (dismissed):

crates/aranya-policy-runner/examples/policy.md
jdygert-spok, line 1: Looks like this was accidentally changed to `recall test_fail` instead of `test_fail`. It might be better to make this use recall instead but if so we should change the command below it as well.

apetkov-so: Yeah, could drop `recall` instead. (And the new recall block)
Also, seems like the message should be something like "device already added".

chip-so: That makes sense. I thought `test_fail` was only supposed to work in debug mode?

chip-so: Well anyway, it works. Updated.

apetkov-so (approved):

PR #736, opened 2026-07-27
707 remove unwrap

### LINKED ISSUE #707: Remove `unwrap`
The `unwrap` expression panics if the value is `None`. That's not ideal, and we can unwrap safely by matching instead:

```
action do_something() result[unit, Error] {
    let fact = match query Foo[] => {} {
        Some(f) => f
        None => return Err(Error::not_found)
    }
    ...
    return Ok(unit)
}
```

So we don't need `unwrap` anymore. We already removed `check_unwrap` (in #706) for similar reasons - matching provides an alternative. (Also, check_unwrap didn't allow calling a recall block.)

We lose the ability to assert though. I'm not sure if that's an issue now that we can return errors from actions.

apetkov-so:
Close #707: Remove `unwrap`, `check_unwrap`. Better error handling, without panics/recalls. We can use optional matching or the coalescing `or` operator instead.


apetkov-so:
@jdygert-spok Where did we land on the seal/open panics? Do we want to keep `unwrap`, only in seal/open blocks?

jdygert-spok:
#750 makes the crypto ffi take an option key input.

crates/aranya-policy-lang/src/lang/parse/tests.rs
```diff
@@ -89,15 +89,15 @@ fn parse_atom_fn() -> Result<(), PestError<Rule>> {
 fn parse_expression() -> Result<(), PestError<Rule>> {
     let mut pairs = PolicyParser::parse(
         Rule::expression,
-        r#"unwrap call(unwrap add(3, 7), saturating_sub(0, b), "foo\x7b")"#,
+        r#"!call(!add(3, 7), saturating_sub(0, b), "foo\x7b")"#,
```
chip-so: I understand not wanting to substantially rewrite the test, but this is kind of weird.

crates/aranya-policy-lang/tests/data/tictactoe.md
```diff
@@ -111,12 +111,12 @@ command Move {
         let player = envelope::author_id(envelope)
         // the query expression searches the fact database for facts which
         // match the signature, returning an Optional containing either all
-        // values marked with ?, or None. The unwrap expression returns the
-        // value inside an Optional or terminates rule execution.
-        let result = unwrap query PlayerProfile[gameID: gameID]=>{x: ?, o: ?}
+        // values marked with ?, or None. Matching on the result binds the
+        // value inside an Optional; the None arm recalls to reject the command.
```
chip-so: This description seems erroneous. There is no matching going on here; this should describe the `or` operator.

crates/aranya-policy-lang/src/lang/parse/tests.rs
apetkov-so: My bad. Claude got creative, and I missed it. I remember fixing this by hand, but I guess it got re-added at some point.

chip-so (dismissed):

chip-so (dismissed):

crates/aranya-policy-lang/src/lang/parse/error/rendering.rs
```diff
@@ -36,21 +36,15 @@ impl Report for InvalidOperator {
                 .patch(Patch::new(rhs.end()..rhs.end(), ")"))
         };
 
-        let elements = if input[op.start()..op.end()] == *"+" {
-            [
-                add_patch("saturating_add(", source.clone()),
-                add_patch("check_unwrap add(", source),
-            ]
```
jdygert-spok, line 43: I would like to continue to suggest `add` and `sub`. Maybe `add(x, y) or todo("handle overflow")`

crates/aranya-policy-lang/tests/data/tictactoe.snap
```diff
@@ -1,2529 +1,8 @@
...
-}
+error: Reserved identifier
+    |
+117 |         let result = query PlayerProfile[gameID: gameID]=>{x: ?, o: ?} or todo()
+    |             ^^^^^^ found reserved identifier: `result`
```
jdygert-spok: fix this error

crates/aranya-model/src/tests/basic-policy.md
jdygert-spok, line 1: It seems like a lot of the unwrap->todo changes across the files should use test_fail instead.

crates/aranya-policy-vm/tests/vm.rs
```diff
@@ -1379,7 +1376,7 @@ fn test_check_errors() -> anyhow::Result<()> {
 }
 
 #[test]
-fn test_check_unwrap() -> anyhow::Result<()> {
+fn test_query_match() -> anyhow::Result<()> {
```
jdygert-spok: This test doesn't make much sense as-is, but we can probably remove it anyways if it was just here to test `check_unwrap` before.

crates/aranya-policy-lang/src/lang/parse/error/rendering.rs
```diff
@@ -38,13 +38,13 @@ impl Report for InvalidOperator {
 
         let elements = if input[op.start()..op.end()] == *"+" {
             [
-                add_patch("saturating_add(", source.clone()),
-                add_patch("check_unwrap add(", source),
+                add_patch("add(", source.clone()),
```
jdygert-spok: nit: we were using unwrap in the suggestion before so we probably want to use `or todo()` so we aren't changing the type.

crates/aranya-policy-lang/tests/data/tictactoe.md
```diff
@@ -111,12 +111,13 @@ command Move {
...
         // the query expression searches the fact database for facts which
         // match the signature, returning an Optional containing either all
-        // values marked with ?, or None. The unwrap expression returns the
-        // value inside an Optional or terminates rule execution.
-        let result = unwrap query PlayerProfile[gameID: gameID]=>{x: ?, o: ?}
-        let playerX = result.x
-        let playerO = result.o
-        let p = unwrap query NextPlayer[gameID: gameID]=>{p: ?}
+        // values marked with ?, or None. The `or` operator unwraps the result
+        // of the query, if it is `Some`, or invokes the todo() expression,
```
jdygert-spok:
```suggestion
        // of the query, if it is `Some`, or invokes the test_fail() expression,
```

crates/aranya-policy-lang/src/lang/parse/tests.rs
```diff
@@ -169,7 +166,7 @@ fn parse_coalesce_operator() -> Result<(), PestError<Rule>> {
 #[test]
 fn parse_expression_pratt() -> Result<(), ParseError> {
     let source = r#"
-        unwrap call(unwrap add(3, 7), saturating_sub(0, b), "foo\x7b")
+        !call(!add(3, 7), saturating_sub(0, b), "foo\x7b")
```
jdygert-spok: This one also uses the `!`. These two expression tests are kinda odd anyway, not sure what we want to do with them. They don't really cover as much at this point :shrug:

apetkov-so: Yeah, it's not useful anymore. Removed.

crates/aranya-policy-lang/src/lang/parse/error/rendering.rs
apetkov-so: Adding a suffix is a bigger change to the add_patch closure... How about just "saturating_add", "add" patches?

jdygert-spok: It's not a very involved change, it just needs another parameter in place of the `")"`.

crates/aranya-policy-lang/src/lang/parse/snapshots/aranya_policy_lang__lang__parse__tests__parse_expression_pratt.snap
jdygert-spok, line 1: rm this file since it is now unreferenced

jdygert-spok (dismissed):
LGTM after one more comment and fix conflicts

crates/aranya-policy-vm/tests/vm.rs
```diff
@@ -1164,10 +1164,7 @@ fn test_match_optional_binding() -> anyhow::Result<()> {
             policy {}
         }
         action foo(o option[int]) {
-            let y = match o {
-                Some(n) => n
-                None => 0
-            }
```
jdygert-spok, line 1170: Revert, just noticed is meant to test `match` not `or`.

jdygert-spok (approved):

supply-chain/imports.lock
chip-so, line 1: I keep seeing changes to the quotes here. This looks like a regression to me but I don't know which is right.

chip-so (approved):

supply-chain/imports.lock
apetkov-so: That file is not in the diff anymore. `cargo make fmt` causes it to change - I guess my local version of cargo-fmt is wrong.

PR #738, opened 2026-07-29
vm-policy: bypass open in braid

jdygert-spok:
This turned out to have a fairly simple solution. However, it does make the assumption that seal uses `serialize(this)` as the envelope payload. But that's an assumption we seemed OK to make since we agreed we could remove custom serialization.


mooreoak (commented):
Dose this mean that we still have serialize in the policy that could be different than it is the CommandPlacement::OnGraphInBraid match arm?

jdygert-spok:
> Dose this mean that we still have serialize in the policy that could be different than it is the CommandPlacement::OnGraphInBraid match arm?

Yes, someone could potentially use an ffi for serialization instead of the `serialize` function, then we'd fail to deserialize during the braid.

If we want to prevent that we can change seal/open to just take in the bytes.

mooreoak:
> Yes, someone could potentially use an ffi for serialization instead of the `serialize` function, then we'd fail to deserialize during the braid.
> 
> If we want to prevent that we can change seal/open to just take in the bytes.

Ya I feel like it would be better to just take the serialization out of the open/seal completely.

mooreoak (commented):
This looks reasonable. It seems like we just track the payload separately from the envelope. 

I want to check that the command id still covers the signature. I did not see any change it that but it was hard to be sure from the diff.

jdygert-spok:
> I want to check that the command id still covers the signature. I did not see any change it that but it was hard to be sure from the diff.

This did not change. Note that the ID computation is not directly tied to the envelope but instead done over specifically chosen info. Here's the process for reference.

https://github.com/aranya-project/aranya-core/blob/54ab63caf71037169c1d0407fe3220d701faa053/crates/aranya-crypto-ffi/src/ffi.rs#L127-L131

https://github.com/aranya-project/aranya-core/blob/54ab63caf71037169c1d0407fe3220d701faa053/crates/aranya-crypto/src/aranya.rs#L298-L303

https://github.com/aranya-project/aranya-core/blob/54ab63caf71037169c1d0407fe3220d701faa053/crates/aranya-crypto/src/policy.rs#L70-L97

https://github.com/aranya-project/aranya-core/blob/54ab63caf71037169c1d0407fe3220d701faa053/crates/aranya-crypto/src/policy.rs#L29-L42

mooreoak:
> This did not change. Note that the ID computation is not directly tied to the envelope but instead done over specifically chosen info. Here's the process for reference.

This looks reasonable.

mooreoak (dismissed):

crates/aranya-policy-ast/src/ast.rs
```diff
@@ -539,10 +539,6 @@ pub enum InternalFunction {
     FactCount(FactCountType, IntLiteral, FactLiteral),
     /// An `if` expression
     If(Box<Expression>, Box<Expression>, Box<Expression>),
-    /// Serialize function
-    Serialize(Box<Expression>),
-    /// Deserialize function
-    Deserialize(Box<Expression>),
```
chip-so, line 545: We will need to remove this from the documentation as well.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -1359,7 +1358,7 @@ impl<'a> CompileState<'a> {
         span: Span,
     ) -> Result<(), CompileError> {
         // fake a function def for the seal block
-        let args = &[param::this(command.identifier.clone())];
+        let args = &[param::this(command.identifier.clone()), param::payload()];
```
chip-so, line 1361: ... and update the documentation about the arguments to `seal` and `open`.

chip-so (approved):
This looks fine to me, but just to check my understanding:

- `deserialize` and `serialize` are now gone; this happens in the runtime
- `payload` in `Envelope` is gone
- `open` doesn't return validated data, it just succeeds if it's valid
- And thus the arguments for `seal` and `open` have changed to include `payload` separately

jdygert-spok:
> This looks fine to me, but just to check my understanding:

Yes to all, and `open` now also includes `this` whereas before that would be accessed as `deserialize(envelope::payload(envelope))`.

mooreoak (approved):

PR #740, opened 2026-08-03
runtime: fix trusting peer-reported max cut

djohnson-spideroak:


crates/aranya-runtime/src/command.rs
```diff
@@ -142,9 +141,13 @@ pub struct Address {
...
 impl Prior<Address> {
     /// Returns the max cut for the command that is after this prior.
+    ///
+    /// This is the only authoritative source of a command's max cut; see
+    /// [`Command::max_cut`].
     pub fn next_max_cut(&self) -> Result<MaxCut, Bug> {
         Ok(match self {
-            Self::None => MaxCut::new(1),
+            // No prior means the init command, at the origin of the graph.
+            Self::None => MaxCut::new(0),
```
djohnson-spideroak, line 150: Something to note is that we had this value mismatched before, with one location using 0 for the init command max cut, and another using 1.

crates/aranya-runtime/src/client.rs
```diff
@@ -45,6 +45,13 @@ pub enum ClientError {
     ParallelFinalize,
     #[error("concurrent transaction usage")]
     ConcurrentTransaction,
+    /// A command reported a max cut which does not match its parents.
+    ///
+    /// A command's max cut is determined by its parents. A command which claims
+    /// otherwise would be stored at the wrong location in the graph, so it is
+    /// rejected rather than added.
```
djohnson-spideroak: trim

crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -225,14 +225,26 @@ impl<SP: StorageProvider, PS: PolicyStore> Transaction<SP, PS> {
...
+            // command: a forged max cut would evade the duplicate check below
+            // and let the command be added twice.
+            let parent = command.parent();
+            let address = Address {
+                id: command.id(),
+                max_cut: parent.next_max_cut()?,
+            };
+            if command.max_cut()? != address.max_cut {
+                return Err(ClientError::MaxCutMismatch);
+            }
```
jdygert-spok: Doesn't this move the issue to trusting the max cut of the parents?

jdygert-spok: Yes, although the parents get validated at least eventually when we search for them to add this command. Need to verify there isn't an ordering issue of trusting before validating.

crates/aranya-runtime/src/command.rs
jdygert-spok: Removing in #741.

crates/aranya-runtime/src/sync/requester.rs
```diff
@@ -249,13 +249,19 @@ impl SyncRequester {
                     let payload = &remaining[start..end];
                     start = end;
 
+                    // The parents determine the max cut, so a differing wire
+                    // value means tampering; reject it rather than silently
+                    // normalizing.
+                    if meta.max_cut != meta.parent.next_max_cut()? {
+                        return Err(SyncError::MaxCutMismatch);
+                    }
```
jdygert-spok: Let's just remove `max_cut` from `CommandMeta`.

crates/aranya-runtime/src/client/transaction.rs
jdygert-spok: Testing this here is a bit weird. I guess it is defending against bad downstream implementations of `Command`?

djohnson-spideroak: I think it also helps to discover broken peers/graphs.

crates/aranya-runtime/src/sync/requester.rs
djohnson-spideroak: Thoughts on doing that in this PR vs a followup? I think it could fit in this PR, so long as the change isn't massive.

jdygert-spok: Its a small change and reverts adding the error variant

crates/aranya-runtime/src/client/transaction.rs
djohnson-spideroak: The check was removed now that we derive directly from the parent.

benz-spideroak (approved):

jdygert-spok:
Closing in favor of #742.

PR #742, opened 2026-08-03
runtime: derive max cut from parents always

jdygert-spok:
Move `max_cut` and `address` out of `trait Command` so they cannot be implemented. `CommandExt` provides the methods but does not allow overriding them. The max cut is now always computed from the parents.


crates/aranya-runtime/src/command.rs
```diff
@@ -101,13 +80,36 @@ impl<C: Command> Command for &C {
     fn bytes(&self) -> &[u8] {
         (*self).bytes()
     }
+}
 
+pub trait CommandExt: Command {
+    /// Return this command's max cut.
+    ///
+    /// Max cut is the maximum distance to the init command.
+    fn max_cut(&self) -> Result<MaxCut, Bug>;
```
jdygert-spok, line 89: This should not use `Bug` since one could send a fake command over the wire with a parent with max cut `usize::MAX`

djohnson-spideroak: Agreed, I think we can probably push to a followup.

djohnson-spideroak (dismissed):

benz-spideroak (dismissed):

djohnson-spideroak (dismissed):

benz-spideroak (approved):

djohnson-spideroak (approved):

PR #743, opened 2026-08-04
Fact db convergence

benz-spideroak:
This adds the dsl machinery to allow testing fact db convergence. There will be follow up PRs with the issues I've uncovered and fixes.


jdygert-spok (approved):

crates/aranya-runtime/src/testing/dsl.rs
```diff
@@ -449,6 +469,19 @@ pub enum TestRule {
         #[serde(default)]
         sync_client_zero: bool,
         sync_method: SyncMethod,
+        /// Percent (0-100) of generated commands that delete a fact.
+        #[serde(default)]
+        delete_chance: u64,
+        /// Percent (0-100) of generated commands that are no-ops (no fact
+        /// mutation). Produces segments with empty fact maps.
+        #[serde(default)]
+        noop_chance: u64,
```
mooreoak, line 478: I think this would be better named `noop_proportion` and `delete_proportion` given the way you used them in the generation loop.

crates/aranya-runtime/src/testing/dsl.rs
mooreoak, line 821: We should control or report the the seed for the Rng  so that tests are repeatable.

mooreoak (approved):
A couple of nits but they can be addressed in a fallow up PR if you want.

crates/aranya-runtime/src/testing/dsl.rs
```diff
@@ -700,6 +767,52 @@ pub trait StorageBackend {
...
+    max_priority: u32,
+) -> TestRule {
+    let key = rng.gen_range(0..key_range);
+    let priority = if max_priority == 0 {
+        0
+    } else {
+        rng.gen_range(0..=max_priority)
+    };
+    let roll = rng.gen_range(0..100);
+    if noop_chance > 0 && roll < noop_chance {
```
chip-so, line 790: Isn't `noop_chance > 0` redundant here? It can't be below zero because it's unsigned and if it is zero, `roll < noop_chance` will always be false anyway.

Also not sure why these "chance" variables are `u64` when even a `u8` would work fine for the 0-100 range. If you just want them to be machine word-sized, `usize` would make more sense.

chip-so (approved):

PR #746, opened 2026-08-04
Check FFI structure when loading module

chip-so:
This is more or less a retread of #670, based on the VM span changes in #726. It creates a new module format (V1) containing "Contract" information, and uses this to validate the FFI API shape of the compiled module against configured FFIs when the runtime (`VmPolicy`) is created.

There is not really a reason to create a new module format, but I thought it was worth exercising the version mechanism to see how painful it is to create a new module version. I'd like to deprecate and remove V0 after this.


crates/aranya-policy-module/src/module/contract.rs
```diff
@@ -0,0 +1,418 @@
...
+use core::fmt;
+
+use aranya_policy_ast::Identifier;
+use serde::{Deserialize, Serialize};
+
+use crate::ffi;
+
+/// An error when validating the contract against expectations
+#[derive(Debug, thiserror::Error)]
+pub struct ContractValidationError(/* TODO(chip): Add detail */ pub String);
```
mooreoak, line 13: What is this TODO doing here? Should it be implemented now, stay a comment or become a ticket?

mooreoak (dismissed):
Did a quick-ish review and the only question I have is about the TODO comment.

crates/aranya-policy-module/src/module/contract.rs
chip-so: It's a wish and a hope, I guess. I'll make a ticket for it.

chip-so: #754

chip-so:
Rebased this on the merged #726.

mooreoak (dismissed):

crates/aranya-policy-module/src/module/contract.rs
```diff
@@ -0,0 +1,421 @@
...
```
jdygert-spok, line 13: Let's make the field private if we want to expand on it later.

We can derive the display impl since we're already using thiserror:

```rs
#[error("Contract Validation Error: {0}")]
pub struct ContractValidationError(String);
```

crates/aranya-policy-module/src/module/contract.rs
```diff
@@ -0,0 +1,421 @@
...
+            return Err(ContractValidationError(format!(
+                "`struct {}` has {} fields but VM expects {}",
+                self.name,
+                self.fields.len(),
+                other.fields.len()
+            )));
+        }
+        for (f1, f2) in self.fields.iter().zip(other.fields.iter()) {
+            f1.validate(f2)
+                .prepend(format!("`struct {}` field `{}`,", self.name, f1.name))?;
```
jdygert-spok, line 243: These prepended `format!`s are performed eagerly even in the happy path. Simple fix is to swap to `format_args!` instead. That just constructs a `core::fmt::Arguments<'_>` which I think is cheap enough.

crates/aranya-policy-module/src/module/contract.rs
chip-so: I did make this private, but I still had to put an accessor on it to support the test routines. :/

jdygert-spok: Could just use the display impl via `to_string()` :shrug:

chip-so: Oh I guess that would work, yeah.

crates/aranya-policy-module/src/module.rs
```diff
@@ -120,6 +129,58 @@ pub struct ModuleV0 {
...
+    pub struct_defs: Vec<StructDef>,
+    /// Enum definitions
+    pub enum_defs: Vec<EnumDef>,
+    /// Code map
+    pub codemap: Option<CodeMap>,
+    /// Global static data
+    pub globals: BTreeMap<Identifier, ConstValue>,
+    /// Module contract metadata
+    pub contract: ModuleContract,
+}
```
jdygert-spok, line 166: Should we make these all the same collection type instead of mixing btreemap, vec, and boxed slice?

crates/aranya-policy-compiler/src/validate.rs
```diff
@@ -12,14 +12,17 @@ use crate::{
 /// - function code paths return values
 /// - commands enter a finish block
 pub fn validate(module: &Module) -> bool {
-    let ModuleData::V0(ref m) = module.data;
+    let m = match &module.data {
+        ModuleData::V0(m) => m.clone(),
+        ModuleData::V1(m) => m.clone().into(),
+    };
```
jdygert-spok, line 18: Bleh, cloning.

crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -205,6 +205,23 @@ impl<CE> VmPolicy<CE> {
...
+                    ContractValidationError::new(format_args!(
+                        "Module has {} FFI modules but VM expects {}",
+                        contract.ffis.len(),
+                        ffis.len()
+                    )),
+                ));
+            }
+            for (mod_ffi, vm_ffi) in contract.ffis.iter().zip(ffis.iter().map(|m| m.schema())) {
+                mod_ffi.validate(&vm_ffi)?;
+            }
```
jdygert-spok: nit: maybe move this into a function in `contract.rs`.

jdygert-spok (dismissed):

mooreoak (dismissed):

crates/aranya-policy-module/src/module.rs
chip-so: I suppose all the `Vec`s could become `Box<[]>`, but the `BTreeMap`s are just things that are simple enough it wasn't worth making a dedicated struct for it.

crates/aranya-runtime/src/vm_policy.rs
chip-so: Playing around with the alternatives for refactoring this into a new function, but they all have warts, too.

1. Make the module crate aware of `FfiCallable` somehow so I can pass `ffis` directly (did not seriously consider this one)
2. Collect all the schemas in a `Vec` and give that to the new function.
3. Don't allocate, but pass the length and an iterator to the new function.

I chose 2, but if the allocation bothers you, 3 preserves the behavior of the original code.

jdygert-spok: Oh, didn't notice you'd have to manipulate it since those are ffis instead of just schemas.

Could take in `impl ExactSizeIterator` so you get the length with the iterator without needing to allocate.

chip-so: Ah, okay. I'm gonna attempt that on the next PR so I don't have to rebase it.

crates/aranya-policy-module/src/module.rs
chip-so: This also gets totally rearranged in the next PR, so I'm gonna leave it as-is for this one and experiment there.

crates/aranya-policy-compiler/src/validate.rs
chip-so: I suppose I could avoid it by unpacking both module versions. The effort doesn't seem worth it in the compiler.

chip-so: Actually, no, TraceAnalyzer needs `ModuleV0`.

mooreoak (approved):

jdygert-spok (approved):

PR #749, opened 2026-08-05
core: fix gates for android builds

djohnson-spideroak:


jdygert-spok (approved):

crates/aranya-libc/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 //!
 //! # Operating System Support
 //!
+//! - Android
```
chip-so, line 5: Is it fair to say we support it if we're not testing it?

chip-so (approved):

crates/aranya-libc/src/lib.rs
mooreoak: I think it's fine to have this as tear 2/3 without testing.

mooreoak (approved):

PR #755, opened 2026-08-12
compiler: detect unused vars in typestack

jdygert-spok:


jdygert-spok:
Closing in favor of #757.

PR #757, opened 2026-08-13
Add error handling, and emit errors, listing all unused vars.

apetkov-so:
This breaks a lot of existing policies, so we have an allow_unused flag so we can apply it selectively.


crates/aranya-policy-compiler/src/compile/types.rs
```diff
@@ -27,12 +28,17 @@ pub(crate) enum UserType<'a> {
     FFIStruct(&'a ffi::Struct<'a>),
 }
 
+/// Synthesized vars for command blocks.
+const PREDEFINED: [&str; 3] = ["envelope", "this", "payload"];
```
jdygert-spok: I think these are only here for the purpose of our own tests omitting seal/open, but since we are allowing unused for the tests anyways we can probably remove this.

crates/aranya-policy-compiler/src/compile/lower.rs
```diff
@@ -1472,7 +1474,9 @@ impl CompileState<'_> {
                     let stmts = self.lower_statements(&arm.statements, Scope::Same)?;
 
                     // Exit the scope for this arm
-                    self.identifier_types.exit_block();
+                    self.identifier_types
+                    .exit_block()
+                    .map_err(|e| self.err(e))?;
```
jdygert-spok: needs a cargo fmt

crates/aranya-policy-compiler/src/compile/types.rs
```diff
@@ -144,12 +153,18 @@ impl IdentifierTypeStack {
...
             .collect();
 
         if !unused.is_empty() {
-            panic!("unused variable(s): {}", unused.join(", "));
+            if self.allow_unused {
+                let names: Vec<&str> = unused.iter().map(|n| n.as_str()).collect();
+                warn!("unused variable(s): {}", names.join(", "));
+            } else {
+                return Err(UnusedVariable { names: unused });
+            }
```
jdygert-spok, line 163:
```suggestion
            let err = UnusedVariable { names: unused };
            if self.allow_unused {
                warn!("{}", err.description());
            } else {
                return Err(err);
            }
```

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -2327,6 +2333,16 @@ impl<'a> Compiler<'a> {
...
 
+    /// Whether local variables that are never read are allowed.
+    ///
+    /// On by default, which logs unused variables as warnings.
+    /// Turn it off to reject them at compile time instead.
+    #[must_use]
+    pub fn allow_unused(mut self, flag: bool) -> Self {
+        self.allow_unused = flag;
+        self
+    }
```
jdygert-spok: We should think about enabling it by default but that could be done on another PR, particularly if it needs a bunch of changes for disabling it for our tests.

jdygert-spok (commented):
Looking good, thanks for taking this. I suggest to base the PR on main so we can just review and merge this on its own.

crates/aranya-policy-compiler/src/compile/types.rs
apetkov-so: I'm not sure who we can remove them; we need to exclude these idents from unused-tracking.

jdygert-spok: But why do we need to exclude them?

apetkov-so: Because things like compile_recall append `Get(ident!("this"))`. Though, maybe they shouldn't.

Edit: Ignore the last comment, these idents are needed for things like seal, open, policy. They are passed as args to `compile_function_like`, which iterates over them, and `Def`s each one. But since they are not user-defined, they don't need to be used.

crates/aranya-policy-compiler/src/compile.rs
apetkov-so: Yeah, I think we should do it in a followup PR. One annoying thing is that many `action`s in tests don't use their args. We could either remove those args, or maybe allow leading underscores in the names, and ignore them.

crates/aranya-policy-compiler/tests/data/unused/map_binding.fail.policy
```diff
@@ -0,0 +1,16 @@
+// A `map ... as` binding is checked like any other local.
+struct Envelope {}
+fact User[i int]=>{name string}
+command Foo {
+    seal { return Envelope {} }
+    open { return Unit }
+    policy {
+        finish {}
+    }
+}
```
jdygert-spok: nit: we could just remove the command since it isn't needed to test the map in the action.

jdygert-spok (dismissed):

crates/aranya-policy-compiler/tests/data/unused/map_binding.fail.policy
apetkov-so: Yeah, that was there because the tracer would fail the action for not publishing. But now that we don't use the tracer, it's not needed.

crates/aranya-policy-compiler/src/compile/types.rs
```diff
@@ -100,30 +116,53 @@ impl IdentifierTypeStack {
...
             .expect("no function scope")
             .pop()
             .expect("no block scope");
+
+        let unused: Vec<Ident> = scope
+            .into_iter()
+            .filter(|(_, local)| !local.used.load(Ordering::Relaxed))
+            .map(|(name, _)| name)
+            // Skip compiler-synthesized vars (like this, envelope, payload).
+            .filter(|name| !name.span().is_empty())
```
jdygert-spok, line 154: This is much better since it won't apply to vars that happen to be named envelope or payload.

jdygert-spok (approved):

crates/aranya-policy-compiler/src/compile/types.rs
```diff
@@ -27,18 +30,27 @@ pub(crate) enum UserType<'a> {
...
-    locals: Vec<Vec<HashMap<Ident, VType>>>,
+    locals: Vec<Vec<IndexMap<Ident, Local>>>,
+    /// When set, don't error on unused vars
+    allow_unused: bool,
+}
+
+#[derive(Debug)]
+struct Local {
+    ty: VType,
+    used: AtomicBool,
```
mooreoak, line 44: Why do we use `AtomicBool` dose this need to be `Sync` for some reason?

mooreoak (approved):
Looks good. I don't understand why we use AtomicBool but no reason to block the PR on that.

crates/aranya-policy-compiler/src/compile/types.rs
jdygert-spok: I suppose it could be `Cell<bool>` instead since we don't currently use multithreading in the compiler. I just usually default to making things `Sync` since it is "just" a potential perf hit rather than needing to go back and change it to work at all.

PR #758, opened 2026-08-13
runtime: support syncing very large commands

jdygert-spok:
Encountering a command which would (may) overflow the buffer will stop adding commands and just send out the existing commands as the sync response. Then the next response should start with that command, so it should be able to be up to `COMMAND_DATA_BUF_SIZE` large, which is around 190KB when not in low-mem mode.

### DIFF
--- a/crates/aranya-runtime/src/sync/responder.rs
+++ b/crates/aranya-runtime/src/sync/responder.rs
@@ -648,7 +648,7 @@ impl SyncResponder {
     ) -> Result<
         (
             Vec<CommandMeta, COMMAND_RESPONSE_MAX>,
-            Vec<u8, MAX_SYNC_MESSAGE_SIZE>,
+            Vec<u8, COMMAND_DATA_BUF_SIZE>,
             usize,
         ),
         SyncError,
@@ -665,9 +665,9 @@ impl SyncResponder {
             }
         };
         let mut commands: Vec<CommandMeta, COMMAND_RESPONSE_MAX> = Vec::new();
-        let mut command_data: Vec<u8, MAX_SYNC_MESSAGE_SIZE> = Vec::new();
+        let mut command_data: Vec<u8, COMMAND_DATA_BUF_SIZE> = Vec::new();
         let mut index = self.next_send;
-        for i in self.next_send..self.to_send.len() {
+        'outer: for i in self.next_send..self.to_send.len() {
             if commands.is_full() {
                 break;
             }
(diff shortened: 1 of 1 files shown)


crates/aranya-runtime/src/sync/responder.rs
```diff
@@ -754,6 +731,39 @@ impl SyncResponder {
...
+
+/// Size of the buffer for storing the command data.
+///
+/// This is smaller than the max sync message size to account for the serialized
+/// `SyncResponseMessage::SyncResponse` to ensure the response will always fit
+/// within a supplied buffer of size `MAX_SYNC_MESSAGE_SIZE`.
+///
+/// The difference is a generous estimate of the size of serialized
+/// `SyncResponseMessage::SyncResponse`.
+const COMMAND_DATA_BUF_SIZE: usize = MAX_SYNC_MESSAGE_SIZE - (50 + COMMAND_RESPONSE_MAX * 150);
```
mooreoak: Even with the comment this line is quite opaque. It should describe `(50 + COMMAND_RESPONSE_MAX * 150)` and 50 and 150 should probably be constants instead of magic numbers.

mooreoak (commented):
I think what this dose allow one command to take up the whole sync response message?

Over all looks good I just think we need more description of the calculation of COMMAND_DATA_BUF_SIZE.

crates/aranya-runtime/src/sync/responder.rs
jdygert-spok: Changed to compute it explicitly

jdygert-spok:
> I think what this dose allow one command to take up the whole sync response message?

Yes, updated PR description with this and some more info.

mooreoak (approved):
This looks good but I do wonder if we should even use `COMMAND_RESPONSE_MAX` in this design. Why not just keep adding commands until it is filled. I think it used to exist to bound the heapless array. It's not part of the protocol as the requester dose not set it.

jdygert-spok:
> This looks good but I do wonder if we should even use `COMMAND_RESPONSE_MAX` in this design. Why not just keep adding commands until it is filled. I think it used to exist to bound the heapless array. It's not part of the protocol as the requester dose not set it.

I agree, I'd rather do that but it is a more complex change so I wanted to put that off for now. #475 would remove `COMMAND_RESPONSE_MAX`.

benz-spideroak (approved):

PR #759, opened 2026-08-13
Module format V1

### LINKED ISSUE #756: Implement Module format
Implement the module format from #696.

chip-so:
This updates the ModuleData::V1 format to match [the spec](https://github.com/aranya-project/aranya-docs/pull/202) and implements read/write to slice routines.

Closes #756


chip-so:
I still have to implement something for AST signatures, but I think this is worth looking at.

crates/aranya-policy-module/src/module/io.rs
```diff
@@ -0,0 +1,338 @@
...
+impl Header {
+    const LEN: usize = size_of::<Self>();
+
+    fn write_to_slice(&self, s: &mut [u8]) -> Result<(), ModuleIoError> {
+        if s.len() < Self::LEN {
+            return Err(ModuleIoError::SliceTooSmall);
+        }
+        let mut b = &mut s[..];
+        b.put_slice(&self.magic);
+        b.put_u32(self.version);
```
mooreoak: these should be put_u32_le() to match the common cpu archetecture. (why is this not the default in rust?) I think you should do a larger audit for check for if you are using little endien encoding and decoding.

crates/aranya-policy-module/src/module/io.rs
```diff
@@ -0,0 +1,338 @@
...
+            return Err(ModuleIoError::SliceTooSmall);
+        }
+        let after_header = &mut s[Header::LEN..];
+        let module_data = match &self.data {
+            ModuleData::V0(m) => postcard::to_allocvec(&m)?,
+            ModuleData::V1(m) => postcard::to_allocvec(&m)?,
+        };
+        let compressed_size = zrip::compress_into(&module_data, after_header, 4)?;
+        let header_data_len = compressed_size.saturating_add(Header::LEN);
+        let header_data_checksum_len = header_data_len.saturating_add(32);
```
mooreoak: What is 32?

Assuming it is the size of the hash you could make it automatic:
```rust
use sha2::Sha256;
use sha2::digest::{OutputSizeUser, typenum::Unsigned};

const HASH_LEN: usize = <Sha256 as OutputSizeUser>::OutputSize::USIZE;

```

crates/aranya-policy-module/src/module/io.rs
```diff
@@ -0,0 +1,338 @@
...
+        if s.len() < Header::LEN {
+            return Err(ModuleIoError::SliceTooSmall);
+        }
+        let after_header = &mut s[Header::LEN..];
+        let module_data = match &self.data {
+            ModuleData::V0(m) => postcard::to_allocvec(&m)?,
+            ModuleData::V1(m) => postcard::to_allocvec(&m)?,
+        };
+        let compressed_size = zrip::compress_into(&module_data, after_header, 4)?;
+        let header_data_len = compressed_size.saturating_add(Header::LEN);
```
mooreoak, line 120: We should have some safety argument why saturating add is safe here. I think it is fine becouse max len of an slice should be usize and I assume these are usize values but we should say someting like that.

crates/aranya-policy-module/src/module/io.rs
```diff
@@ -0,0 +1,338 @@
...
+    ///
+    /// The slice does not have to be exactly sized; the module header will read only the bytes
+    /// necessary. If there are not enough bytes in the slice according to the header, this will
+    /// return [`ModuleIoError::SliceTooSmall`].
+    pub fn read_from_slice(s: &[u8]) -> Result<Self, ModuleIoError> {
+        let h = Header::read_from_slice(s)?;
+        if h.magic != MODULE_MAGIC {
+            return Err(ModuleIoError::BadMagic);
+        }
+        let header_data_checksum_len = h.size as usize;
```
mooreoak: we need a compile time assert that `usize::MAX >= u32::MAX` or we need to use `try_into`

mooreoak (changes requested):

crates/aranya-policy-module/src/module/io.rs
jdygert-spok: I recommend to switch to using `zerocopy::byteorder::little_endian::U32` for the types in the header, and adding the appropriate `zerocopy` trait derives to the header. Then these two methods can just be `IntoBytes::write_to_prefix` and `FromBytes::ref_from_prefix` (or `TryFromBytes` if you use an enum to encode the magic value and version).

crates/aranya-policy-module/src/module/io.rs
chip-so: Yes, I should have known there was some way to know this statically. Looks like there's a slightly shorter way to write it through a convenient type alias: `OutputSize::<Sha256>::USIZE`

crates/aranya-policy-module/src/module/io.rs
chip-so: `usize` is defined to be the same size as the pointer size on the platform, so this could only be erroneous on platforms with 16-bit pointers. So it can't be true on anything we currently support, and is exceedingly unlikely to be true on anything we support in the future (as much as I would love to port this to AVR). But I have swapped these out for  `try_into()`s.

crates/aranya-policy-module/src/module/io.rs
chip-so: Added.

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -404,6 +440,8 @@ impl<'a> CompileState<'a> {
             }
         }
 
+        hash_parts!(self.signature_hasher => Enum &enum_def.identifier, &enum_def.variants);
```
chip-so: I wanted to do a head check before I go hog wild with this. Does this seem like a reasonable approach for capturing the logical form of the source code, given that we don't have a good way to iterate over the AST? The major downside is it would be our responsibility to make sure anything relevant gets hashed, but it does mean we can easily ignore anything that isn't relevant.

chip-so (commented):
Updated based on feedback, and I have a question below about my approach for AST signature hashing.

crates/aranya-policy-compiler/src/compile.rs
mooreoak: It seems reasonable to me. I think the a hash or hash tree is a great way to give things compact names/types names

crates/aranya-policy-compiler/src/compile.rs
```diff
@@ -124,6 +125,39 @@ macro_rules! typekind {
     };
 }
 
+/// Hashes data from an AST representation and updates a [`Digest`](sha2::Digest) with the type, identifier.
+///
+/// ```
+/// hash_parts!(self.signature_hasher => Enum &enum_def.identifier, &enum_def.variants);
+/// ```
+macro_rules! hash_parts {
```
mooreoak: I might call this `update_hash_parts` or something. `hash_parts` sounds like it should return a value.

mooreoak (dismissed):

crates/aranya-policy-module/src/module/io.rs
```diff
@@ -0,0 +1,348 @@
...
+            return Err(ModuleIoError::SliceTooSmall);
+        }
+        let header_data_len = header_data_checksum_len.saturating_sub(CHECKSUM_HASH_LEN);
+        let checksum = &s[header_data_len..header_data_checksum_len];
+        let computed_checksum: [u8; CHECKSUM_HASH_LEN] =
+            Sha256::digest(&s[..header_data_len]).into();
+        if checksum != computed_checksum {
+            return Err(ModuleIoError::ChecksumMismatch);
+        }
+        let compressed_data = &s[Header::LEN..header_data_len];
```
benz-spideroak, line 165: This will panic if size is too small. You can end up something like `&s[20..16]` if `header_data_len < Header::LEN`.

crates/aranya-policy-module/src/module/contract.rs
```diff
@@ -0,0 +1,402 @@
...
+
+impl FunctionContract {
+    fn validate(&self, other: &ffi::Func<'_>) -> Result<(), ContractValidationError> {
+        if self.name != other.name {
+            return Err(ContractValidationError(format!(
+                "function `{}`, VM expected `{}`",
+                self.name, other.name
+            )));
+        }
+        for (a1, a2) in self.args.iter().zip(other.args.iter()) {
```
benz-spideroak, line 191: I believe you need a check that self and other have the same number of args.

crates/aranya-policy-compiler/src/compile.rs
benz-spideroak: Is hash collision a security issue or just for convenience? It looks like you can get collisions because of the way `\x1F` is used as a separator. `f(x: A) -> BC` and `f(x: AB) -> C` both hash to `x␟ABC`

mooreoak: @benz-spideroak your saying that becouse the return value has no identifier it is functionally appended to the last argument in the hash?  

I admit that don't fully understand the macro_rules syntax and may be misunderstanding how it works.

chip-so: I suppose it could be a security issue, but I don't see it right now. The goal is that policies with equivalent source code (that is, they do the same thing but might have different text) have the same hash. What I'm realizing from your example is I should be hashing the argument types in addition to the names. Ideally this elides the names and only considers the functionality, but that's harder so I'm being a little stricter than necessary.

Edit: Ah, I am hashing the types.

crates/aranya-policy-module/src/module/io.rs
chip-so: I think that could only happen if the module is intentionally malformed, but that's something we should be defending against.

If the length is >= `Header::LEN + CHECKSUM_HASH_LEN`, then this slice is 0 or more bytes. If it's less, the checksum overlaps with the header. I suppose you could still have a valid header and checksum if you corrupt `Header.uncompressed_size`, which isn't checked at all. So you could make a malformed module that causes a crash here.

I've added a check that `header_data_checksum_len >= Header::LEN + CHECKSUM_HASH_LEN`.

crates/aranya-policy-module/src/module/v1.rs
```diff
@@ -0,0 +1,64 @@
...
+    PartialEq,
+    Serialize,
+    Deserialize,
+    rkyv::Archive,
+    rkyv::Deserialize,
+    rkyv::Serialize,
+)]
+pub struct Contract {
+    /// A SHA-256 hash over the normalized AST
+    pub signature: [u8; CHECKSUM_HASH_LEN],
```
jdygert-spok, line 51: What's the intended use for this signature? I would guess to be used as the policy ID for the policy store.

crates/aranya-policy-module/src/module/io.rs
```diff
@@ -0,0 +1,374 @@
...
+#[derive(
+    Debug,
+    PartialEq,
+    Eq,
+    zerocopy::IntoBytes,
+    zerocopy::FromBytes,
+    zerocopy::Immutable,
+    zerocopy::KnownLayout,
+)]
+struct Header {
```
jdygert-spok, line 79: Need `#[repr(C)]` on this for consistent layout. I'm surprised zerocopy doesn't complain.

jdygert-spok: Oh, it isn't strictly required if you just want to convert to/from bytes within a certain compilation of a rust program. So zerocopy only needs to check the particular layout selected by the compiler this time (for the `IntoBytes` impl which prohibits padding).

crates/aranya-policy-module/src/module/v1.rs
mooreoak: I think the policyID would have to include the FFIs as well but given that we `use` them in the policy we might already have that?

mooreoak (dismissed):
Looks reasonable to me but I think @jdygert-spok is right about the repr(C) commnet.

crates/aranya-policy-module/src/module/v1.rs
chip-so: The signature does include `use` statements. It hashes everything in the AST.

jdygert-spok: Based on the work in #783 it seems to make more sense just hashing the particular serialized module bytes since that's what we're passing around, so I'm not sure what hashing the ast currently gives us.

crates/aranya-policy-module/src/module/io.rs
```diff
@@ -0,0 +1,374 @@
...
+    const LEN: usize = size_of::<Self>();
+}
+
+impl Module {
+    /// Write a module to a mutable slice.
+    ///
+    /// The slice need to be large enough to contain the serialized module, or it will
+    /// return [`ModuleIoError::SliceTooSmall`]. But since this process involves serialization and
+    /// compression, it is not generally possible to know that size ahead of time.
+    pub fn write_to_slice(&self, s: &mut [u8]) -> Result<usize, ModuleIoError> {
```
jdygert-spok, line 107: IMO for this we can just return a vec instead of writing to a slice.

benz-spideroak (dismissed):

crates/aranya-policy-module/src/module/v1.rs
mooreoak: It provides normalization over white space right?

mooreoak: I would not get rid of this hashing as I think it will generally be a useful way to compair source even if we decided to go with @jdygert-spok suggestion of the policy id is the hash of the module. The advantage of the hash of the module as the id is that one can check it before parsing anything so an attacker who is not authorized to update the policy can't attack the module parsing.

crates/aranya-policy-module/src/module/io.rs
mooreoak: I think it is better to not do the allocation in the write function but it's not a hill I will die on.

crates/aranya-policy-compiler/src/compile/ast_hash.rs
```diff
@@ -0,0 +1,574 @@
...
+
+/// Hashes data from an AST representation and updates a [`Digest`](sha2::Digest) with the type, identifier.
+///
+/// ```ignore
+/// update_hash_parts!(self.signature_hasher => Enum &enum_def.identifier, &enum_def.variants);
+/// ```
+macro_rules! update_hash_parts {
+    // A type with no name uses _, which elides the name from the output
+    ($hasher:expr => $ident:ident _, $($tail:tt)* ) => {
+        $hasher.update(&[0x01]); // ASCII Start of Heading
```
apetkov-so, line 21: I would move these magic values to an `enum`.

crates/aranya-policy-module/Cargo.toml
```diff
@@ -13,9 +13,13 @@ aranya-policy-ast = { version = "0.16.0", path = "../aranya-policy-ast" }
 
 fnv = { workspace = true, default-features = false }
 indexmap = { workspace = true, default-features = false, features = ["serde"] }
+postcard = { workspace = true, features = ["alloc"] }
 rkyv = { workspace = true, features = ["alloc", "indexmap-2"] }
 serde = { workspace = true, default-features = false, features = ["derive"] }
+sha2 = { workspace = true }
 thiserror = { workspace = true }
+zerocopy = { workspace = true, features = ["derive"] }
+zrip = { version = "0.8.5", default-features = false, features = ["alloc", "paranoid"] }
```
jdygert-spok, line 22: These two crates both have over 1000x as many downloads and seem to support no-std, maybe we should use one of them instead:

- https://crates.io/crates/miniz_oxide
- https://crates.io/crates/brotli

chip-so: I don't have strong feelings about what compressor we use, but FWIW I did audit `zrip` for safety and it looks fine (in as much as any compression library "looks fine", it's a ton of math and array accesses). `unsafe` is used purely for speed and is completely disabled by the `paranoid` feature, which I'm using.

crates/aranya-policy-module/src/module/v1.rs
chip-so: > It provides normalization over white space right?

It's AST-level. It doesn't consider white space at all.

mooreoak: That's what I assumed I just didn't know if it was preserving white space for error reporting or something.

apetkov-so (dismissed):

chip-so:
On the demo call we decided to get rid of the AST hashing stuff. I'll get to that on Monday.

chip-so:
Actually, I've got some more work to do on this, so hold off on review for now.

crates/aranya-policy-module/src/module/contract.rs
```diff
@@ -188,6 +188,14 @@ impl FunctionContract {
...
             )));
         }
+        if self.args.len() != other.args.len() {
+            return Err(ContractValidationError::new(format_args!(
+                "function `{}` has {} arguments but expected {}",
+                self.name,
+                other.args.len(),
+                self.args.len()
+            )));
+        }
```
benz-spideroak, line 198: This appears to be backwards. It's arguments are in the opposite order from the check at line 249.

crates/aranya-policy-compiler/src/compile/ast_hash.rs
```diff
@@ -0,0 +1,574 @@
...
+        $hasher.update(stringify!($kind).as_bytes());
```
benz-spideroak: This appears to be a typo. You bind $ident, but output $kind which is unbound.

crates/aranya-policy-compiler/src/compile/ast_hash.rs
```diff
@@ -0,0 +1,574 @@
...
+}
+
+impl AstHash for StmtKind {
+    fn ast_hash(&self, hasher: &mut impl Update) {
+        match self {
+            Self::Let(let_stmt) => let_stmt.ast_hash(hasher),
+            Self::Check(check_stmt) => check_stmt.ast_hash(hasher),
+            Self::Match(match_stmt) => match_stmt.ast_hash(hasher),
+            Self::If(if_stmt) => if_stmt.ast_hash(hasher),
+            Self::Finish(statements) => statements.ast_hash(hasher),
```
benz-spideroak: Does this need to call update_hash_parts? It appears it will hash without the finish tag.

crates/aranya-policy-compiler/src/compile/ast_hash.rs
chip-so: Yep, that's a mistake. All of this is going away, anyway.

crates/aranya-policy-module/src/module/io.rs
chip-so: I've added this, but I wonder where the inconsistency would come from. Everything is already aligned to `u32`. Does Rust sometimes pad a `u32` to 64-bit alignment?

crates/aranya-policy-module/src/module/contract.rs
chip-so: Good catch.

crates/aranya-policy-module/src/module/io.rs
mooreoak: I think Rust could even reorder fields between versions. Rust dose not have a stable ABI so it can kinda do what it wants.

jdygert-spok: This particular struct would _probably_ not be changed. But perhaps the compiler could notice that one of the fields is accessed much more often and move it to the front to minimize pointer arithmetic.

Generally, `repr(Rust)` has no guarantees about field order and padding so that the compiler can try to optimize the layout, especially for more complex cases of generics, niches, and alignment.

crates/aranya-policy-module/src/module/v1.rs
chip-so: Signature has been removed.

crates/aranya-policy-compiler/src/compile/ast_hash.rs
chip-so: No longer valid since AST hashing was removed.

benz-spideroak (approved):

jdygert-spok (approved):

PR #760, opened 2026-08-18
fix(runtime): clear pending fact writes in LinearPerspective::revert

benz-spideroak:
`LinearPerspective::revert` early-returned when `checkpoint.index == self.commands.len()`, treating "no commands added since the checkpoint" as "nothing to revert". But fact writes accumulate in the perspective's overlay map and `current_updates` buffer *while a rule executes*, and are only attributed to a command by `add_command` afterwards — so the perspective can be dirty at equal index. `Transaction::add_single` hits exactly that state on every rule failure (checkpoint before the rule, revert on error, `add_command` never runs), meaning the early return made `revert` a no-op on the one path that depends on it.

The consequence of a rule that writes facts and then fails — the shape a VM policy produces whenever a `finish` block errors partway, since `VmPolicyIO` writes into the perspective unbuffered per instruction — is twofold:

1. **Phantom facts:** the rejected command's writes stay visible to every subsequent policy evaluation in the same transaction, so commands that should be rejected can be accepted.
2. **Silent fact divergence:** the next successful `add_command` does `mem::take(&mut current_updates)`, absorbing the leaked writes into *its own* persisted `CommandData.updates`. A peer reconstructing facts mid-segment from stored updates then computes different fact state than a peer that evaluated the same commands in a clean transaction — at an identical graph head, so graph-equality convergence checks stay green while fact state forks permanently.

The fix drops the early return: the equal-index case now falls through to the same rebuild the truncating path already used (clear the overlay and `current_updates`, replay the remaining commands' updates). `revert` is only called on failure paths, so the replay adds no hot-path cost. `SessionPerspective::revert` keeps its analogous-looking early return deliberately — its checkpoint indexes the fact log itself (write granularity), so equal index there genuinely means nothing was written.

Reproduced at two levels, committed failing-first so each is verified against the broken behavior:

- `test_revert_clears_writes_made_after_checkpoint` — minimal `Revertable` contract test mirroring the `add_single` call pattern (checkpoint → insert → revert, no `add_command`).
- `rejected_command_fact_leak` DSL vector — end-to-end divergence: adds a `Poison` test command whose rule writes its fact and then rejects, plus an `IngestPoisonThenSet` DSL rule that ingests it, asserts the rejection, then commits a valid command in the same transaction (modeling a caller that tolerates per-command errors, which the API permits). Before the fix, the ingesting peer committed the rejected command's write while a cleanly-synced peer at the identical head did not, failing `CompareGraphs`' fact-cache check with `payload[9]=99` vs. absent.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/aranya-runtime/src/storage/linear/mod.rs
+++ b/crates/aranya-runtime/src/storage/linear/mod.rs
@@ -1128,11 +1128,16 @@ impl<R: Read> Revertable for LinearPerspective<R> {
     }
 
     fn revert(&mut self, checkpoint: Checkpoint) -> Result<(), StorageError> {
-        // No early return when `checkpoint.index == self.commands.len()`:
-        // fact writes accumulate in `facts`/`current_updates` between
-        // `add_command` calls, so the perspective can be dirty even when no
-        // command was added since the checkpoint (e.g. a rule that wrote
-        // facts and then failed). Those writes must be cleared too.
+        // Equal command count alone does not mean clean: a rule that wrote
+        // facts and then failed leaves its writes pending in
+        // `facts`/`current_updates` without having added a command. But
+        // every fact write pushes onto `current_updates`, so an empty
+        // buffer at equal command count means the fact overlay is untouched
+        // since the checkpoint and there is nothing to rebuild.
+        if checkpoint.index == self.commands.len() && self.current_updates.is_empty() {
+            return Ok(());
+        }
+
         if checkpoint.index > self.commands.len() {
             bug!(
                 "A checkpoint's index should always be less than or equal to the length of a perspective's command history!"
--- a/crates/aranya-runtime/src/storage/linear/mod.rs
+++ b/crates/aranya-runtime/src/storage/linear/mod.rs
@@ -1128,10 +1128,11 @@ impl<R: Read> Revertable for LinearPerspective<R> {
     }
 
     fn revert(&mut self, checkpoint: Checkpoint) -> Result<(), StorageError> {
-        if checkpoint.index == self.commands.len() {
-            return Ok(());
-        }
-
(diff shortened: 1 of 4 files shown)


crates/aranya-runtime/src/storage/linear/mod.rs
```diff
@@ -1128,10 +1128,11 @@ impl<R: Read> Revertable for LinearPerspective<R> {
     }
 
     fn revert(&mut self, checkpoint: Checkpoint) -> Result<(), StorageError> {
-        if checkpoint.index == self.commands.len() {
-            return Ok(());
-        }
```
jdygert-spok, line 1133: I believe we could keep this by also checking for fact changes:

```rs
checkpoint.index == self.commands.len() && self.current_updates.is_empty()
```

It seems worthwhile since commands are usually expected to fail before the finish block, so this should be true most of the time that a revert happens.

benz-spideroak: Done in 915a7f61. Verified the guard is sufficient: every fact write goes through `LinearPerspective::insert`/`delete`, which unconditionally push onto `current_updates`, and the buffer is only drained by `add_command` (which grows `commands`, failing the index check) — so empty buffer at equal command count proves the overlay is untouched. Full runtime suite passes.

jdygert-spok (dismissed):
Needs a fmt but lgtm.

crates/aranya-runtime/src/storage/linear/mod.rs
benz-spideroak: Done

jdygert-spok (approved):

mooreoak (approved):

PR #765, opened 2026-08-24
fix(runtime): Always calculate priority

benz-spideroak:
A command's Priority travels as unauthenticated sync metadata (CommandMeta) and is used verbatim as the braid strand-ordering key. Evaluated commands are already checked against the policy's priority map in vm_policy, but merges are never evaluated, so a byzantine or buggy relay could alter one merge's priority and make downstream peers braid the same DAG in a different total order: an identical head set (graph-equality passes) with permanently divergent facts -- or, with Priority::Finalize, a crash when the lone merge strand is emitted into the braid for evaluation.

This PR removes priority from the data send when syncing and calculates priorities.


mooreoak:
"A command's Priority travels as unauthenticated sync metadata (CommandMeta) and is used verbatim as the braid strand-ordering key."

Why are we sending the priority over the network we have all the information locally to compute it and we should never trust a priority that was sent us.

This could be a historical artifact but it is a bug we need to fix if we are trusting the remote assertion of priorities.

benz-spideroak:
> "A command's Priority travels as unauthenticated sync metadata (CommandMeta) and is used verbatim as the braid strand-ordering key."
> 
> Why are we sending the priority over the network we have all the information locally to compute it and we should never trust a priority that was sent us.
> 
> This could be a historical artifact but it is a bug we need to fix if we are trusting the remote assertion of priorities.

I updated the code to no longer send priorities. They're only computed locally.

crates/aranya-runtime/src/client/session.rs
```diff
@@ -562,7 +560,15 @@ mod test {
                     .insert("seq".into(), Keys::default(), data.into())
                     .unwrap();
             }
-            Ok(())
+            Ok(match command.parent() {
+                Prior::None => Priority::Init,
+                // Use the last byte of the ID as priority, just so we can
```
mooreoak: Why do we need to see braiding effects in session?

mooreoak (dismissed):
I really like how this cleans up priorities.

crates/aranya-runtime/src/storage/mod.rs
```diff
@@ -915,6 +915,13 @@ pub trait Segment {
     /// Returns the command at the given location.
     fn get_command(&self, location: Location) -> Option<Self::Command<'_>>;
 
+    /// Returns the priority of the command at the given location.
+    ///
+    /// Priorities are assigned at ingest (structurally for merge and init
+    /// commands, by the policy for evaluated commands) and persisted with
+    /// the command.
+    fn get_priority(&self, location: Location) -> Option<Priority>;
```
jdygert-spok, line 923: I don't love making this a separate method in storage. We could add a `trait Prioritized` that `Storage::Command` must implement, but IDK that's particularly worth doing.

crates/aranya-runtime/src/client/session.rs
```diff
@@ -426,7 +419,13 @@ where
...
     }
 
-    fn add_command(&mut self, command: &impl Command) -> Result<usize, StorageError> {
+    fn add_command(
+        &mut self,
+        command: &impl Command,
+        _priority: Priority,
+    ) -> Result<usize, StorageError> {
+        // Session commands are ephemeral and never braided, so the priority
+        // is not persisted.
```
jdygert-spok, line 428: Could check that it is basic priority like before.

crates/aranya-runtime/src/storage/linear/mod.rs
```diff
@@ -1160,14 +1166,30 @@ impl<R: Read> Perspective for LinearPerspective<R> {
...
+        // contradicts the command's structure would poison braid ordering, so
+        // catch caller bugs before the value becomes durable.
+        debug_assert!(
+            match command.parent() {
+                Prior::Merge(..) => priority == Priority::Merge,
+                Prior::None => priority == Priority::Init,
+                Prior::Single(..) => matches!(priority, Priority::Basic(_) | Priority::Finalize),
+            },
+            "priority must match command structure"
+        );
```
jdygert-spok: Let's make this a proper error or at least a `bug!`.

crates/aranya-runtime/src/testing/protocol.rs
```diff
@@ -67,18 +67,20 @@ pub struct TestProtocol<'a> {
...
-            WireProtocol::Init(_) => Priority::Init,
-            WireProtocol::Merge(_) => Priority::Merge,
-            WireProtocol::Basic(m) => Priority::Basic(m.prority),
-            WireProtocol::Delete(m) => Priority::Basic(m.prority),
-            WireProtocol::NoOp(m) => Priority::Basic(m.prority),
-            WireProtocol::Poison(m) => Priority::Basic(m.prority),
-        }
+/// Derives a command's priority from its (id-covered) serialized body,
+/// mirroring how a real policy derives priority from the command kind.
+fn wire_priority(command: &WireProtocol) -> Priority {
```
jdygert-spok: nit: would be nicer as a method IMO

jdygert-spok (commented):
Good fix

crates/aranya-runtime/src/client/session.rs
benz-spideroak: Add the check

crates/aranya-runtime/src/storage/linear/mod.rs
benz-spideroak: Done

crates/aranya-runtime/src/testing/protocol.rs
benz-spideroak: changed

jdygert-spok (dismissed):

mooreoak (approved):

jdygert-spok (approved):

PR #768, opened 2026-08-26
fix(runtime): validate merge command ID from sync

jdygert-spok:


crates/aranya-runtime/src/client/transaction.rs
```diff
@@ -792,6 +794,11 @@ mod test {
 
             Ok(SeqCommand::new(id, Prior::Merge(left, right)))
         }
+
+        fn validate_merge(&self, _command: &impl Command) -> Result<(), PolicyError> {
```
apetkov-so, line 798: This isn't a trait method, so why implement a no-op function?

jdygert-spok: It is a method in `trait Policy`, it just has a default implementation.

crates/aranya-runtime/src/policy.rs
```diff
@@ -178,6 +187,21 @@ pub trait Policy {
...
         ids: MergeIds,
     ) -> Result<Self::Command<'a>, PolicyError>;
+
+    /// Validate a merge command.
+    fn validate_merge(&self, command: &impl Command) -> Result<(), PolicyError> {
+        let ids = match command.parent() {
+            crate::Prior::Merge(left, right) => MergeIds::from_ordered(left, right),
+            _ => None,
+        }
+        .ok_or(PolicyError::Panic)?;
```
apetkov-so, line 197: We should probably log these with `error!` so we can diagnose. I'm not a fan of generic panic errors, but I guess it's fine.

apetkov-so (approved):

benz-spideroak (approved):

PR #769, opened 2026-08-27
fix(runtime): Convergence map livelock

benz-spideroak:
ConvergenceMap::should_continue could loop forever: probing a spilled block installed it into memory, which re-spilled an evicted block back onto the root index, so a lookup covered by every block's max_cut range but present in none never ran out of root entries to scan. Each iteration also wrote a block to the spill file, growing it without bound. Any peer braiding a graph with this shape hangs permanently.

The fix probes a copy of the spilled block and installs it only on a hit, so the scan index always advances past misses. Misses no longer cost a spill write.

Includes a regression test that builds the poisoned state through the map's public API and fails fast via a spill-write tripwire if the livelock returns (verified failing on the unfixed code).

### DIFF
--- a/crates/aranya-runtime/src/client/convergence_map.rs
+++ b/crates/aranya-runtime/src/client/convergence_map.rs
@@ -533,21 +533,25 @@ mod livelock_tests {
         Location::new(SegmentIndex::new(segment), MaxCut::new(K))
     }
 
-    /// Build a `ConvergenceMap` whose state is:
-    /// - one spilled block on disk (segments 1..=256, all at `max_cut = K`),
-    /// - all three in-memory blocks non-empty, all at `max_cut = K`
-    ///   (segments 257..=512, 513..=768, 769..=1024),
+    /// Build the state the livelock needed, then run one `lookup(query)`
+    /// against it and return the result.
     ///
-    /// then call `lookup` for `query` and return its result. Every block's
-    /// `[min_max_cut, max_max_cut]` range is `[K, K]`, so any query at
-    /// `max_cut = K` is "covered" by every block.
+    /// That state is: all three in-memory blocks full (segments 257..=512,
+    /// 513..=768, 769..=1024) and one block spilled to disk (segments
+    /// 1..=256). Every entry sits at the same `max_cut = K`, so every
+    /// block's `[min, max]` range is exactly `[K, K]` — meaning any
+    /// depth-K query looks like it could be in *any* of the four blocks.
+    /// A depth-K query for a segment that was never inserted is therefore
+    /// "covered everywhere, present nowhere": the shape that used to loop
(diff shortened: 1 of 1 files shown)


crates/aranya-runtime/src/client/convergence_map.rs
```diff
@@ -386,14 +385,20 @@ impl<'a, F: Spill> ConvergenceMap<'a, F> {
...
+        // Advance BFS to cover the query location.
+        self.advance_to(storage, location.max_cut)?;
+        self.lookup(location)
+    }
+
+    /// Look up `location` in memory and on disk, consuming an entry on
+    /// a hit. The BFS must already cover `location.max_cut` (see
+    /// [`Self::should_continue`], which advances it first).
+    fn lookup(&mut self, location: Location) -> Result<bool, ClientError> {
         self.access_counter = self
```
mooreoak, line 397: Is it ok that this updates the counter even if we return and error?

mooreoak (commented):
This looks good but I want to make sure that the incrementing the visited count even when error is returned is ok.

jdygert-spok (dismissed):
I don't know the convergence map particularly well but these changes seem reasonable.

crates/aranya-runtime/src/client/convergence_map.rs
jdygert-spok: Seems fine, we are just using it to mark blocks with a last-accessed time for LRU eviction, so it should be ok to go up more than "necessary".

crates/aranya-runtime/src/client/convergence_map.rs
```diff
@@ -405,15 +410,21 @@ impl<'a, F: Spill> ConvergenceMap<'a, F> {
...
-                    // Load block into memory (removes root[ri] via swap_remove).
-                    let bi = self.load_block_from_disk(ri)?;
-                    if let Some(ei) = self.storage.blocks[bi].find(location) {
+                    // Probe a copy without touching the root index or
+                    // evicting an in-memory block, so `ri` advances past
+                    // every miss and the scan terminates even when all
+                    // block ranges cover `location.max_cut`. Installing
+                    // on a miss would re-append the evicted block to the
+                    // root and the scan would never run out of entries.
+                    let probed = self.read_block_from_disk(ri)?;
```
mooreoak: I know these are not new symbols but the ri, ei, bi, names make it hard do understand the code. I don't know wha they represent.

benz-spideroak: I added more descriptive names.

mooreoak (dismissed):

jdygert-spok (approved):

djohnson-spideroak (approved):

Issue #607, opened 2026-03-09
policy-lang: Disallow function recursion


chip-so:
We changed function compilation to remove ordering dependency, but now that functions can call any other function, they can recurse in violation of [the spec](https://aranya-project.github.io/policy-book/reference/top-level/functions.html). We should have some way of detecting and disallowing this in the compiler.

cc @apetkov-so @djohnson-spideroak @gknopf-aranya @jdygert-spok @ssweetney-spok

PR #772, opened 2026-08-28
Ensure all `action` paths `publish` or return an error.

apetkov-so:


crates/aranya-policy-compiler/src/tests.rs
```diff
@@ -248,3 +248,142 @@ fn test_validate_publish() {
...
+        // A nested action's `publish` does not satisfy the caller.
+        concat(
+            r#"
+            action h() {
+                publish Foo { a: 0 }
+            }
+
+            action i() {
+                action h() // fail
+            }
```
jdygert-spok, line 375: Why do we want this to fail? I could see exposing an action that wraps another action with some default values or to shape the API in a certain way or something.

apetkov-so: Following `publish` across calls seemed too hard for the benefit. :) The tracer visits each action separately, so if an inner one doesn't `publish`, its difficult to determine if it's part of a path that does publish. Also, all actions are callable from the app so the non-publishing one is a legitimate fail.

jdygert-spok: We can simply assume that calling another action does publish.

This isn't strictly true because we haven't disallowed cyclic recursion yet ala #607. But otherwise you're just calling another action which we would separately check to ensure that it publishes.

apetkov-so: I can see the value in being able to delegate publishing to another action, but the word "assume" makes me uneasy. It's probably not a big deal, though - the worst case is a badly-written policy that doesn't publish anything.
@chip-so

jdygert-spok: To be clear, if

1. cyclic recursion is not allowed
2. for each action, every path either publishes directly or calls another action

then it is guaranteed that each action publishes on every path.

PR #774, opened 2026-09-01
afc: enhance replay protection

djohnson-spideroak:
- add replay protection for control command create channel
- add epoch to policy and bind channels to epoch in FFI
- use command ID as nonce for control command
- cap number of channels (number of nonces) per epoch to 16k

### DIFF
--- a/crates/aranya-afc-util/src/ffi.rs
+++ b/crates/aranya-afc-util/src/ffi.rs
@@ -67,6 +67,7 @@ function create_uni_channel(
     seal_id id,
     open_id id,
     label_id id,
+    epoch int,
 ) struct AfcUniChannel
 "#)]
     pub(crate) fn create_uni_channel<E: Engine>(
@@ -79,7 +80,9 @@ function create_uni_channel(
         seal_id: DeviceId,
         open_id: DeviceId,
         label_id: LabelId,
+        epoch: i64,
     ) -> Result<AfcUniChannel, FfiError> {
+        let epoch = u64::try_from(epoch).map_err(|_| FfiError::InvalidEpoch)?;
         let our_sk = &self
             .store
             .lock()
@@ -94,6 +97,7 @@ function create_uni_channel(
             seal_id,
             open_id,
             label_id,
+            epoch,
         };
         let UniSecrets { author, peer } = UniSecrets::new(eng, &ch)?;
 
@@ -133,6 +137,9 @@ pub(crate) enum FfiError {
     /// The keystore failed.
     #[error("keystore failure")]
     KeyStore,
+    /// The epoch is negative.
+    #[error("epoch must be non-negative")]
+    InvalidEpoch,
     /// Bug
     #[error("bug: {0}")]
     Bug(Bug),
--- a/crates/aranya-afc-util/src/handler.rs
+++ b/crates/aranya-afc-util/src/handler.rs
@@ -8,7 +8,11 @@ use aranya_crypto::{
 use aranya_fast_channels::Directed;
 use serde::{Deserialize, Serialize};
 
-use crate::{shared::decode_enc_pk, transform::Transform};
+use crate::{
+    replay::{ReplayStore, Verdict},
+    shared::decode_enc_pk,
+    transform::Transform,
+};
 
 /// Wraps `tracing::error` to always use the `afc-handler`
 /// target.
@@ -71,6 +75,7 @@ impl<S: KeyStore> Handler<S> {
             our_sk,
             their_pk,
             label_id: effect.label_id,
+            epoch: effect.epoch,
         };
 
         UniKey::new(&ch, secret, UniKey::SealOnly)
@@ -78,18 +83,40 @@ impl<S: KeyStore> Handler<S> {
 
     /// Converts a [`UniPeerEncap`] into a key suitable for
     /// [`AranyaState`][aranya_fast_channels::AranyaState].
-    pub fn uni_channel_received<E, SK, OK>(
+    ///
+    /// The control message is first checked for freshness via
+    /// [`ReplayStore::accept`]: anything other than
+    /// [`Verdict::Fresh`] fails with [`Error::Replay`] before any
+    /// key material is derived. The message is recorded durably
+    /// before this returns, so a crash before the key is
+    /// installed loses the channel (the sender creates a new
+    /// one), but the message can never be accepted twice.
+    pub fn uni_channel_received<E, R, SK, OK>(
         &mut self,
         eng: &E,
+        replay: &mut R,
+        graph: BaseId,
         effect: &UniChannelReceived<'_>,
     ) -> Result<UniKey<SK, OK>, Error>
     where
         E: Engine,
+        R: ReplayStore,
         SK: for<'a> Transform<(&'a UniChannel<'a, E::CS>, UniPeerEncap<E::CS>)>,
         OK: for<'a> Transform<(&'a UniChannel<'a, E::CS>, UniPeerEncap<E::CS>)>,
     {
         if effect.seal_id == self.device_id {
             return Err(Error::AuthorMustBeSealer);
+        }
+
+        // Freshness, before any key material exists.
+        match replay
+            .accept(graph, effect.seal_id, effect.epoch, effect.cmd_id)
+            .map_err(|err| {
+                error!("replay store failed: {err}");
+                Error::ReplayStore
+            })? {
+            Verdict::Fresh => {}
+            verdict => return Err(Error::Replay(verdict)),
         }
 
         let encap =
@@ -111,6 +138,7 @@ impl<S: KeyStore> Handler<S> {
             our_sk,
             their_pk,
             label_id: effect.label_id,
+            epoch: effect.epoch,
         };
 
         UniKey::new(&ch, encap, UniKey::OpenOnly)
@@ -132,6 +160,8 @@ pub struct UniChannelCreated<'a> {
     pub label_id: LabelId,
     /// The unique key identifier for the [`UniAuthorSecret`].
     pub key_id: UniKeyId,
+    /// The channel author's epoch at creation time.
+    pub epoch: u64,
 }
 
 /// Data from the `AfcUniChannelReceived` effect.
@@ -149,6 +179,12 @@ pub struct UniChannelReceived<'a> {
     pub label_id: LabelId,
     /// The peer's encapsulation.
     pub encap: &'a [u8],
+    /// The channel author's epoch at creation time.
+    pub epoch: u64,
+    /// The ID of the control command that created the channel.
+    ///
+    /// This is the replay-protection nonce.
+    pub cmd_id: CmdId,
 }
 
 /// Uniquely identifies a unirectional channel.
@@ -221,4 +257,10 @@ pub enum Error {
     /// A `crypto` crate error.
     #[error(transparent)]
     Crypto(#[from] aranya_crypto::Error),
+    /// The control message was rejected by the [`ReplayStore`].
+    #[error("control message rejected: {0}")]
+    Replay(Verdict),
+    /// The [`ReplayStore`] failed.
+    #[error("replay store failure")]
+    ReplayStore,
 }
--- a/crates/aranya-afc-util/src/handler.rs
+++ b/crates/aranya-afc-util/src/handler.rs
@@ -32,6 +32,34 @@ impl<S> Handler<S> {
     pub const fn new(device_id: DeviceId, store: S) -> Self {
         Self { device_id, store }
     }
+
+    /// Handles the `AfcEpochRotated` effect by raising the
+    /// replay floor for the rotated device via
+    /// [`ReplayStore::raise_floor`], forgetting the nonces
+    /// recorded for its older epochs.
+    ///
+    /// A device's own rotation is a no-op: replay state is only
+    /// kept for messages from *other* devices (see
+    /// [`Error::AuthorMustBeSealer`]).
+    pub fn epoch_rotated<R>(
+        &mut self,
+        replay: &mut R,
+        graph: BaseId,
+        effect: &EpochRotated,
+    ) -> Result<(), Error>
+    where
+        R: ReplayStore,
+    {
+        if effect.device_id == self.device_id {
+            return Ok(());
+        }
+        replay
+            .raise_floor(graph, effect.device_id, effect.epoch)
+            .map_err(|err| {
+                error!("replay store failed: {err}");
+                Error::ReplayStore
+            })
+    }
 }
 
 // Uni impl.
@@ -185,6 +213,15 @@ pub struct UniChannelReceived<'a> {
     ///
     /// This is the replay-protection nonce.
     pub cmd_id: CmdId,
+}
+
+/// Data from the `AfcEpochRotated` effect.
+#[derive(Copy, Clone, Debug, Serialize, Deserialize)]
+pub struct EpochRotated {
+    /// The device whose epoch was rotated.
+    pub device_id: DeviceId,
+    /// The new epoch.
+    pub epoch: u64,
 }
 
 /// Uniquely identifies a unirectional channel.
--- a/crates/aranya-afc-util/src/replay.rs
+++ b/crates/aranya-afc-util/src/replay.rs
@@ -1,0 +1,419 @@
+//! Receiver-side replay protection for AFC control messages.
+
+use core::fmt;
+
+use aranya_crypto::{BaseId, DeviceId, policy::CmdId};
+
+/// The default per-sender, per-epoch nonce cap.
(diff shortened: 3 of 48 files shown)


mooreoak:
> cap number of channels (number of nonces) per epoch to 16k

As this just goes in the control message why not a u32 worth?

crates/aranya-afc-util/src/ffi.rs
```diff
@@ -79,7 +80,9 @@ function create_uni_channel(
         seal_id: DeviceId,
         open_id: DeviceId,
         label_id: LabelId,
+        epoch: i64,
```
mooreoak, line 83: why i64 and not u64?

djohnson-spideroak:
> > cap number of channels (number of nonces) per epoch to 16k
> 
> As this just goes in the control message why not a u32 worth?

Its a storage tradeoff, we can bump to a u32 worth if we want. On the wire, postcard uses the variable length integers so we dont save anything on the wire either way.

mooreoak:
> > > cap number of channels (number of nonces) per epoch to 16k
> > 
> > 
> > As this just goes in the control message why not a u32 worth?
> 
> Its a storage tradeoff, we can bump to a u32 worth if we want. On the wire, postcard uses the variable length integers so we dont save anything on the wire either way.

The storage costs are really on the receiver. I see your point. I guess it's ok.

crates/aranya-afc-util/src/handler.rs
```diff
@@ -149,6 +207,21 @@ pub struct UniChannelReceived<'a> {
     pub label_id: LabelId,
     /// The peer's encapsulation.
     pub encap: &'a [u8],
+    /// The channel author's epoch at creation time.
+    pub epoch: u64,
+    /// The ID of the control command that created the channel.
```
mooreoak, line 212: I don't understand why this comment has two descriptions of the field and the blank line.

crates/aranya-afc-util/src/replay.rs
```diff
@@ -0,0 +1,419 @@
...
+    /// `>= epoch`.
+    fn raise_floor(
+        &mut self,
+        graph: BaseId,
+        sender: DeviceId,
+        epoch: u64,
+    ) -> Result<(), Self::Error>;
+}
+
+impl<T: ReplayStore + ?Sized> ReplayStore for &mut T {
```
mooreoak, line 53: We don't get this automatically? if `T` `impl`s `ReplayStore`?

crates/aranya-afc-util/src/replay.rs
```diff
@@ -0,0 +1,419 @@
...
+            graph: BaseId,
+            sender: DeviceId,
+            epoch: u64,
+            nonce: CmdId,
+        ) -> Result<Verdict, Self::Error> {
+            let mut inner = self.inner.lock();
+            let cap = inner.cap;
+            let rec = inner.records.entry((graph, sender)).or_default();
+
+            if epoch > rec.epoch_max {
```
mooreoak, line 222: It seems like you should call self.raise_floor() here to ensure that we have one implementation of the algrythem.

crates/aranya-afc-util/src/testing.rs
```diff
@@ -559,25 +773,477 @@ where
...
+        )
+        .expect_err("negative epoch should be rejected");
+    assert!(matches!(err, FfiError::InvalidEpoch), "{err}");
+}
+
+/// Delivering the same control message twice installs exactly
+/// one key: the second delivery is rejected as a replay.
+pub fn test_receive_replayed_uni_channel<T: TestImpl>()
+where
+    <T::Aranya as AranyaState>::SealKey: for<'a> Transform<(
```
mooreoak, line 841: That is quite the type. It is a test so it dose not matter so much but dose this really need to be gernic over T or could we just use a concreate type here?

crates/aranya-afc-util/src/replay.rs
```diff
@@ -0,0 +1,419 @@
...
+
+use aranya_crypto::{BaseId, DeviceId, policy::CmdId};
+
+/// The default per-sender, per-epoch nonce cap.
+pub const DEFAULT_NONCE_CAP: usize = 16_384;
+
+/// Durable, receiver-side memory of accepted control messages.
+///
+/// Records are keyed by `(graph, sender)`. `graph` is the graph's
+/// ID as a [`BaseId`].
```
jdygert-spok, line 13: Why `BaseId` instead of `GraphId`?

jdygert-spok: Oh, this doesn't depend on the runtime. I think we could move the graph ID definition to crypto and re-export it from runtime.

crates/aranya-crypto/src/afc/uni.rs
```diff
@@ -178,6 +186,8 @@ pub(crate) struct Info {
     seal_id: DeviceId,
     open_id: DeviceId,
     label_id: LabelId,
+    /// Big-endian `u64`.
+    epoch: [u8; 8],
```
jdygert-spok, line 190: Let's use little endian and use the zerocopy wrapper type: `zerocopy::little_endian::U64`.

crates/aranya-fast-channels/src/replay.rs
```diff
@@ -0,0 +1,441 @@
...
+/// authenticated, so that a forged frame cannot move the window.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ReplayWindow {
+    /// Highest authenticated `seq`; `None` until the first
+    /// frame.
+    high: Option<u64>,
+    /// Bit `i` set ⇔ `high − i` has been accepted (`i < size`).
+    bitmap: [u64; WORDS],
+    /// `W`, in `1..=MAX_REPLAY_WINDOW`.
+    size: u16,
```
jdygert-spok, line 59: So size is an artificial limit, making it behave as if we had `W` bits of storage instead? Is there any reason to limit the window size other than space, which this doesn't reduce?

crates/aranya-fast-channels/src/replay.rs
```diff
@@ -0,0 +1,441 @@
...
+        let word = i / (u64::BITS as usize);
+        let bit = i % (u64::BITS as usize);
+        if let Some(w) = self.bitmap.get_mut(word) {
+            *w |= 1u64 << bit;
+        }
+    }
+
+    /// Shifts the bitmap up by `n` positions (bit `i` becomes bit
+    /// `i + n`), discarding bits that fall off the end.
+    fn shift(&mut self, n: u64) {
```
jdygert-spok, line 190: Instead of physically shifting the bits, we should treat it like a ring buffer wrapping around. Then we only need to clear the "new" bits.

crates/aranya-fast-channels/src/replay.rs
jdygert-spok, line 1: AFAICT this is completely orthogonal to the epoch replay stuff so it might be nice to move it to another PR for smaller PR reviewing.

crates/aranya-afc-util/src/replay.rs
```diff
@@ -0,0 +1,419 @@
...
+    /// for a daemon's implementation. It is **not durable** and
+    /// must not be used in production.
+    ///
+    /// Cloning a `MemStore` yields a handle to the same
+    /// underlying state.
+    #[derive(Clone, Debug)]
+    #[cfg_attr(docsrs, doc(cfg(feature = "testing")))]
+    pub struct MemStore {
+        inner: Arc<Mutex<Inner>>,
+    }
```
jdygert-spok, line 162: Why are we cloning this and using interior mutability?

crates/aranya-afc-util/src/testing.rs
```diff
@@ -230,21 +242,165 @@ impl<T: TestImpl> Device<T> {
...
+    }
+
+    /// Handles the `AfcUniChannelReceived` effect for `ch` with
+    /// `replay`, without installing the key.
+    #[allow(clippy::type_complexity)]
+    fn receive<R: ReplayStore>(
+        &mut self,
+        author: &Self,
+        ch: &CreatedChannel,
+        replay: &mut R,
```
jdygert-spok, line 337: Why pass in the replay when we hold it in `self.replay`?

crates/aranya-fast-channels/src/client.rs
```diff
@@ -23,18 +24,73 @@ use crate::{
...
     }
 
     /// Returns the current state.
     pub fn state(&self) -> &S {
         &self.state
     }
+
+    /// Returns the [`ReplayWindow`] size used for new
+    /// [`OpenCtx`]s.
+    pub const fn replay_window(&self) -> u16 {
```
mooreoak, line 63: this should be called `replay_window_size` for clearty.

crates/aranya-fast-channels/src/client.rs
```diff
@@ -23,18 +24,73 @@ use crate::{
...
+/// [`Client::setup_open_ctx`].
+#[derive(Debug)]
+pub struct OpenCtx<C> {
+    inner: C,
+    window: ReplayWindow,
+}
+
+impl<C> OpenCtx<C> {
+    /// Returns the underlying [`AfcState::OpenCtx`].
+    pub const fn inner(&self) -> &C {
```
mooreoak, line 81: I am suprised we are declaring these const. Is that actually useful?

crates/aranya-fast-channels/src/replay.rs
```diff
@@ -0,0 +1,441 @@
...
+/// RFC 4303 recommends a window of at least 64 for IPsec.
+pub const DEFAULT_REPLAY_WINDOW: u16 = 64;
+
+/// The maximum [`ReplayWindow`] size, in frames.
+pub const MAX_REPLAY_WINDOW: u16 = 1024;
+
+/// The number of `u64` words in the bitmap.
+const WORDS: usize = (MAX_REPLAY_WINDOW as usize) / u64::BITS as usize;
+
+const _: () = assert!(WORDS * (u64::BITS as usize) == MAX_REPLAY_WINDOW as usize);
```
mooreoak, line 26: A comment explaining this would be good.

crates/aranya-fast-channels/src/replay.rs
```diff
@@ -0,0 +1,441 @@
...
+    }
+
+    /// Reports whether `seq` would be accepted, without
+    /// mutating the window.
+    ///
+    /// Call this before performing any cryptographic work on the
+    /// frame.
+    pub fn check(&self, seq: Seq) -> Result<(), Error> {
+        let seq = seq.to_u64();
+        let Some(high) = self.high else {
```
mooreoak, line 120: We should consider a `cold_path()` annotation here. As it can only happen on the first packet. 

https://doc.rust-lang.org/std/hint/fn.cold_path.html

crates/aranya-fast-channels/src/replay.rs
```diff
@@ -0,0 +1,441 @@
...
+            return Ok(());
+        };
+        if seq > high {
```
mooreoak, line 123: We should put the rest of the function in the else and use `cold_path()` here. 

https://doc.rust-lang.org/std/hint/fn.cold_path.html

crates/aranya-fast-channels/src/replay.rs
```diff
@@ -0,0 +1,441 @@
...
+            return Err(Error::ReplayedSeq { seq: Seq::new(seq) });
+        }
+        Ok(())
+    }
+
+    /// Records `seq` as accepted.
+    ///
+    /// Only call this after the frame has been authenticated;
+    /// [`check`][Self::check] must have returned `Ok` for `seq`.
+    pub fn commit(&mut self, seq: Seq) {
```
mooreoak, line 138: We should use `cold_path()` in all but the `seq > high` arm.

 https://doc.rust-lang.org/std/hint/fn.cold_path.html

crates/aranya-fast-channels/src/replay.rs
```diff
@@ -0,0 +1,441 @@
...
+    }
+
+    /// Reports whether bit `i` is set.
+    ///
+    /// Bits outside the bitmap are clear.
+    fn bit(&self, i: u64) -> bool {
+        let Ok(i) = usize::try_from(i) else {
+            return false;
+        };
+        let word = i / (u64::BITS as usize);
```
mooreoak, line 169: As bit's will be a power of 2 we can use a shift operator here.  (the compiler might figure that out for us but we might as well do it)

crates/aranya-fast-channels/src/replay.rs
```diff
@@ -0,0 +1,441 @@
...
+        let bit = i % (u64::BITS as usize);
```
mooreoak, line 170: I think this can be written as `(i & (u64::BITS as usize - 1))`

crates/aranya-fast-channels/src/replay.rs
```diff
@@ -0,0 +1,441 @@
...
+                }
+            }
+        }
+    }
+
+    /// Reports whether bit `i` is set.
+    ///
+    /// Bits outside the bitmap are clear.
+    fn bit(&self, i: u64) -> bool {
+        let Ok(i) = usize::try_from(i) else {
```
mooreoak, line 166: more cold path.

crates/aranya-fast-channels/src/replay.rs
```diff
@@ -0,0 +1,441 @@
...
+        };
+        let word = i / (u64::BITS as usize);
+        let bit = i % (u64::BITS as usize);
+        self.bitmap.get(word).is_some_and(|w| (w >> bit) & 1 == 1)
+    }
+
+    /// Sets bit `i`.
+    ///
+    /// Bits outside the bitmap are ignored.
+    fn set_bit(&mut self, i: u64) {
```
mooreoak, line 177: Same optimization as in `bit()` apply.

crates/aranya-fast-channels/src/replay.rs
```diff
@@ -0,0 +1,441 @@
...
+        let bit = i % (u64::BITS as usize);
+        if let Some(w) = self.bitmap.get_mut(word) {
+            *w |= 1u64 << bit;
+        }
+    }
+
+    /// Shifts the bitmap up by `n` positions (bit `i` becomes bit
+    /// `i + n`), discarding bits that fall off the end.
+    fn shift(&mut self, n: u64) {
+        let Ok(n) = usize::try_from(n) else {
```
mooreoak, line 191: This seems wrong can you explain why reseting the map is the right thing to do here?

crates/aranya-fast-channels/src/replay.rs
```diff
@@ -0,0 +1,441 @@
...
+//! tolerating reordering.
+
+use aranya_crypto::afc::Seq;
+
+use crate::error::Error;
+
+/// The default [`ReplayWindow`] size, in frames.
+///
+/// RFC 4303 recommends a window of at least 64 for IPsec.
+pub const DEFAULT_REPLAY_WINDOW: u16 = 64;
```
mooreoak, line 18: I think we should just set the replay window to the word size so that we can alway use fast bitwise math. This is in the hot path so it is worth doing I think.

crates/aranya-afc-util/src/replay.rs
```diff
@@ -0,0 +1,419 @@
...
+
+/// Durable, receiver-side memory of accepted control messages.
+///
+/// Records are keyed by `(graph, sender)`. `graph` is the graph's
+/// ID as a [`BaseId`].
+///
+/// Per `(graph, sender)` the store conceptually holds
+/// `Record { epoch_max: u64, seen: Set<CmdId> }`, with an absent
+/// record meaning `{ 0, ∅ }`.
+pub trait ReplayStore {
```
mooreoak, line 18: I don't think replay store should be responsible for the ratcheting logic. It should just be the storage backend for the state. We don't want each impl making the acptance ratcheting choices.

mooreoak (changes requested):
This is in the right direction. Across my comments I think one theme is that the epoch check and the per packet sequance numbers are in the hot path and we will want to make sure they are fast. I might have done some premature optimization in the comments but it is at least worth benchmarking it.

crates/aranya-afc-util/src/replay.rs
djohnson-spideroak: As in, the store should just be for storing the nonces? So the API would be more like basic CRUD for the nonces? Something like: `register_nonce`, `check_nonce`, `clear`?

crates/aranya-afc-util/src/testing.rs
djohnson-spideroak: We were using this pattern previously so that we could put it into the test_all macro for use with the different engines, etc.

crates/aranya-afc-util/src/ffi.rs
djohnson-spideroak: Policy `int` data type is i64. We could write a conversion in the VM but its getting converted somewhere.

PR #777, opened 2026-09-01
policy-text: manual proptest impls

jdygert-spok:
Remove `policy-text` dependency on `std`.


benz-spideroak (approved):

apetkov-so (approved):
Should add a comment to the PR to explain why we

jdygert-spok:
> Should add a comment to the PR to explain why we

?

apetkov-so:
> > Should add a comment to the PR to explain why we
> 
> ?

Never mind, I figured out why we're doing this. It's to remove dependency on std.

PR #778, opened 2026-09-01
runtime: expose `MemSpill`

jdygert-spok:


crates/aranya-runtime/src/storage/spill.rs
```diff
@@ -107,23 +106,24 @@ impl Spill for LibcSpill {
     }
 }
 
-// --- testing (in-memory) backend ---
+/// Convenience function for passing as `make_spill` argument.
+pub const fn mem_spill() -> Result<MemSpill, StorageError> {
+    Ok(MemSpill::new())
```
chip-so, line 111: The fact that this is always `Ok` makes me question the need for the result in the first place. I understand this fits into a general allocator callback, but does it need to consider errors?

jdygert-spok: The result is for e.g. `LibcSpill::new` which creates a new file so can actually error reasonably.

chip-so (approved):

djohnson-spideroak (approved):

PR #269, opened 2025-06-10
crypto: add `PolicyId` contextual binding

elagergren-spideroak:
Fixes #23

Signed-off-by: Eric Lagergren <elagergren@spideroak.com>

PR #783, opened 2026-09-03
MVP load policy module into store

jdygert-spok:


crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -190,11 +191,124 @@ macro_rules! vm_effect {
...
+                }
+            };
+            let ffis = self.ffis.select(ffi_names).expect("TODO");
+            let machine = Machine::from_module(module).expect("TODO");
+            let policy = VmPolicy::new(machine, self.engine.clone(), ffis).expect("TODO");
+            e.insert(policy);
+        }
+
+        Ok(id)
+    }
```
jdygert-spok, line 245: Here's the heart of the loading.

crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -626,7 +742,7 @@ impl<CE: aranya_crypto::Engine> Policy for VmPolicy<CE> {
         sink: &mut impl Sink<Self::Effect>,
         action_placement: ActionPlacement,
     ) -> Result<(), PolicyError> {
-        let VmAction { name, args } = action;
+        let VmAction { name, args, policy } = action;
```
jdygert-spok, line 745: The policy bytes are fed into the init command via the action but this is kinda awkward and the policy bytes are not yet included when hashing the init command.

jdygert-spok: I think #269 shows about half off the solution toward including the policy ID in the command ID derivation.

mooreoak:
So this is for use with eventual loading of policies on policy upgrade?

jdygert-spok:
Just realized this doesn't handle restarts at all. If we plan to keep the modules in the commands then we can trigger a reload from the source command somehow. If we move them out then the policy store needs to be an actual on disk store.

PR #791, opened 2026-09-09
Move seal/open from policy lang to runtime traits

jdygert-spok:


crates/aranya-crypto-ffi/src/lib.rs
jdygert-spok, line 1: The crypto ffi is no more.

crates/aranya-envelope-ffi/src/ffi.rs
jdygert-spok, line 1: envelope ffi still provides author_id, command_id, parent_id, and signature to the policy and recall blocks, but we might want to figure out a different way instead.

crates/aranya-model/src/tests/mod.rs
```diff
@@ -174,6 +184,40 @@ impl ClientFactory for FfiClientFactory {
...
+                .find(|(name, _)| name.as_str() == "sign_pk")
+                .ok_or(OpenError::BadStruct)?
+                .1;
+            let Value::Bytes(key) = key else {
+                return Err(OpenError::BadStruct);
+            };
+            let key: aranya_crypto::VerifyingKey<CE::CS> =
+                postcard::from_bytes(key).map_err(OpenError::BadKey)?;
+            aranya_runtime::vm_policy::open_with_key(command_struct, payload, envelope, key)
+        } else {
```
jdygert-spok, line 209: Had to customize model's opener because it has this `AddDeviceKeys` command which lets a user add themself to the graph, and the command itself holds the keys.

crates/aranya-perspective-ffi/src/ffi.rs
```diff
@@ -7,19 +7,13 @@ use aranya_policy_vm::{CommandContext, MachineError, MachineErrorType, ffi::ffi}
...
-/// }
-///
 /// action do_something() {
 ///   let head_id = perspective::head_id()
 /// }```
 pub struct FfiPerspective;
 #[ffi(module = "perspective")]
 impl FfiPerspective {
-    /// Returns the ID of the command at the head of the perspective. Only valid for `Seal` and `Action` contexts.
+    /// Returns the ID of the command at the head of the perspective. Only valid for `Action` context.
```
jdygert-spok, line 16: Feels very extra to have perspective ffi just to get the head in actions. Although this one would be extra tricky to remove since the head changes after publish in the actions.

crates/aranya-policy-runner/src/policy.rs
```diff
@@ -39,14 +39,20 @@ pub fn create_vmpolicy<CE: aranya_crypto::Engine>(
...
-        Box::from(IdamFfi::new(keystore)),
+        Box::from(IdamFfi::new(keystore.try_clone()?)),
         Box::from(PerspectiveFfi),
     ];
 
+    let seal = Arc::new(StandardSeal::new(
+        device_id,
+        SigningKeyId::default(), // hmmm
+        keystore,
+    ));
```
jdygert-spok, line 52: Policy runner doesn't (can't?) know the signing ID here.

jdygert-spok: We could have the sealer look it up based on the command or facts. This is how I originally had the sealers working before I realized you should generally just know the key you will use. We'd have to add back facts to the seal trait but that would be fine. But instead maybe the policy runner just needs to change to be more in line with the fact we aren't seal/opening in policy anymore.

crates/aranya-policy-vm/src/context.rs
```diff
@@ -46,10 +30,6 @@ pub struct PolicyContext {
 pub enum CommandContext {
     /// Action
     Action(ActionContext),
-    /// Seal operation
-    Seal(SealContext),
-    /// Open operation
-    Open(OpenContext),
```
jdygert-spok, line 52: No more seal/open context in vm of course.

crates/aranya-policy-vm/src/machine.rs
```diff
@@ -973,49 +973,6 @@ where
                     }
                 }
             }
-            Instruction::Serialize => {
```
jdygert-spok, line 976: I had forgotten to remove the serialize and deserialize instructions in #738, so need to remove them here since they use seal/open context.

crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -196,6 +195,8 @@ pub struct VmPolicy<CE> {
     engine: CE,
     ffis: Vec<Box<dyn FfiCallable<CE> + Send + 'static>>,
     priority_map: BTreeMap<Identifier, VmPriority>,
+    seal: Arc<dyn Seal<CE>>,
+    open: Arc<dyn Open<CE>>,
```
jdygert-spok, line 199: I think cloneable trait objects are the way to go for the sealer and opener stored in the vm policy instances. I think we'd probably store them in the `VmPolicyStore` to be fetched when a policy is loaded, but I'm not sure how we'd make it so you can select different ones if that is needed. It's roughly like the `FfiSet` in #783, but I don't know what we would "key" based on.

crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -370,42 +375,43 @@ impl<CE: aranya_crypto::Engine> VmPolicy<CE> {
...
+        facts: &mut impl QueryValue,
+    ) -> Result<(), PolicyError> {
+        self.open
+            .open_command(command_struct, payload, envelope, facts)
+            .map_err(|error| {
+                tracing::warn!(%error, "could not open command");
+                PolicyError::Panic
+            })?;
+
+        Ok(())
```
jdygert-spok, line 414: The `VmPolicy`'s seal/open basically just call out to the trait objects.

crates/aranya-runtime/src/vm_policy.rs
```diff
@@ -685,44 +686,21 @@ impl<CE: aranya_crypto::Engine> Policy for VmPolicy<CE> {
...
+                        let command_name = command_struct.name.clone();
 
                         // The parent of a basic command should be the command that was added to the perspective on the previous
                         // iteration of the loop
                         let parent = rs.io.facts.head_address()?;
+
+                        let parent_id = match parent {
+                            Prior::None => CmdId::default(),
+                            Prior::Single(x) => x.id,
+                            Prior::Merge(_, _) => todo!(),
```
jdygert-spok: fixme

crates/aranya-runtime/src/vm_policy/testing.rs
```diff
@@ -1,82 +1,62 @@
...
 
-use crate::testing::hash_for_testing_only;
+use crate::{
+    command::CmdId,
+    vm_policy::{Envelope, Open, OpenError, QueryValue, Seal, SealError},
+};
 
-pub struct TestFfiEnvelope {
-    pub device: DeviceId,
-}
```
jdygert-spok, line 14: `TestFfiEnvelope` is gone since it was just used for a fake seal open. There's a few spots where this was the only ffi used for some tests. I _think_ that was of necessity rather than the purpose of testing that FFIs work but this did reduce the amount of tests using FFI.

crates/aranya-runtime/src/vm_policy/testing.rs
```diff
@@ -1,82 +1,62 @@
...
     }
 }
+
+fn cmd_id<CS: CipherSuite>(parent_id: CmdId, author_id: DeviceId, payload: &[u8]) -> CmdId {
+    use aranya_crypto::id::IdExt as _;
+    CmdId::new::<CS>(
+        b"TestSealCommandId",
+        [parent_id.as_bytes(), author_id.as_bytes(), payload],
+    )
+}
```
jdygert-spok, line 62: New `TestSeal`/`TestOpen` replaces the `TestFfiEnvelope`. They still don't actually sign but just create and check command ID. (Test open previously didn't check the ID. No changes in tests from that.)

crates/aranya-runtime/src/vm_policy/seal_open.rs
jdygert-spok, line 1: This is the main addition, the seal/open traits and some implementations.

crates/aranya-runtime/src/vm_policy/seal_open.rs
```diff
@@ -0,0 +1,225 @@
...
+        Err(SealError::NotSupported)
+    }
+}
+
+/// Implements the standard command seal, using the provided key.
+pub struct StandardSeal<KS> {
+    device_id: DeviceId,
+    sign_key_id: SigningKeyId,
+    keystore: KS,
+}
```
jdygert-spok, line 87: Standard seal takes in the author id and the signing key (via ID along with keystore), and just uses those directly to perform the seal.

crates/aranya-runtime/src/vm_policy/seal_open.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+}
+
+/// Implements the standard command open.
+///
+/// The init command will be opened using the key bytes at `this.sign_pk`.
+///
+/// Other commands will be opened with the key stored at
+/// `fact DeviceSignPubKey[device_id id]=>{key bytes, ...}`.
+pub struct StandardOpen;
```
jdygert-spok, line 148: This is not _great_. `StandardOpen` expects a particular field for init commands and a particular fact otherwise. This works OK, as long as we are willing to say you just have to implement your own if you don't follow this. 

The more I think about this, I'm not particularly sure there is value in configuring the open implementation itself, but instead only in configuring how to fetch the key. The key fetching could either be via a trait in rust, or something in the policy lang.

crates/aranya-runtime/src/vm_policy/seal_open.rs
```diff
@@ -0,0 +1,225 @@
...
+            .ok_or(OpenError::BadFact)?
+            .value;
+        let Value::Bytes(key) = key else {
+            return Err(OpenError::BadFact);
+        };
+        Cow::Owned(key)
+    };
+
+    postcard::from_bytes(&key).map_err(OpenError::BadKey)
+}
```
jdygert-spok, line 225: This is what it looks like to fetch the verifying key in Rust.

jdygert-spok:
Replaced by #804.

PR #793, opened 2026-09-11
policy: remove unused instructions

jdygert-spok:


jdygert-spok:
Note this disturbs the value of existing instructions as far as serialization is concerned. If we want to keep them the same, we can explicitly set the discriminant values. IDK that really matters yet?

chip-so (approved):

chip-so:
> Note this disturbs the value of existing instructions as far as serialization is concerned. If we want to keep them the same, we can explicitly set the discriminant values. IDK that really matters yet?

AFAIK it shouldn't matter in anything we're currently shipping. I think it might break the embedded project if you don't rebuild the intermediate module, but I think it might do that anyway.

apetkov-so (approved):

PR #795, opened 2026-09-11
chore: silence cargo fmt unstable config warnings

jdygert-spok:
For context, the script applies the config values via the command line since that allows us to apply unstable config values. We now prevent rustfmt from reading the config since it would warn about the config holding unstable config values. I previously missed that this was possible since I was looking for an option specifically to ignore config instead of just pointing to an empty one.


chip-so (approved):

apetkov-so:
Why does skipping a config file silence those errors?

apetkov-so (approved):

jdygert-spok:
> Why does skipping a config file silence those errors?

When using stable rustfmt, if your config has an unstable option, you will get a warning and that option will not be applied. But if you set the same option via CLI, it is applied and no warning given.

PR #798, opened 2026-09-14
Naive braid and sync

benz-spideroak:
Adds `aranya_runtime::testing::braid` (compiled under the `testing` feature):

- `naive`: a reference braid that walks the whole graph backward, taking the
  frontier command with the lowest (priority, id), then reverses. Plain data,
  whole-slice scans, no code shared with production. Its output is also checked
  against the weave requirements in graph.md.
- `shapes`: enumerates every graph shape a client can hold, up to isomorphism,
  from one definition of the legal growth moves (single child, or merge of an
  unmerged concurrent pair, at most `PEERS` chains).
- `policy`: a probe policy whose only rule appends each command's id to a `seq`
  fact, so the committed fact is production's evaluation order.
- `harness`: builds each shape across real clients, one per chain, reconciles
  them through the real sync protocol, and diffs every client's `seq` against
  the oracle.

Exhaustive for all 849 shapes with up to 8 commands, plus a seeded random sweep
at 10 to 30 commands. A mismatch implicates the braid walk, its incremental
composition across commits, the sync protocol, or the oracle's reading of the
spec. No production code changes.


mooreoak (dismissed):
I did a mid effort review on this and it looks reasonable. I will not say that I understand everything but for test code I am comfortable with it.

jdygert-spok (approved):

chip-so (approved):
I don't totally understand everything going on here, but I don't see any obvious mistakes.

PR #803, opened 2026-09-15
Close #800: `action` calls are now expressions, passing along the inner action's result.

### LINKED ISSUE #800: `action` needs an expression form
Since actions can now return, the action-calling statement (`action foo()`) needs to be changed to an expression so it can propagate the result value back to the caller.

apetkov-so:
- `action` is now an expression that evaluates to the return type of the called action, or to `Never`, if the action doesn't return.
- There's an `Expression` statement kind that allows expressions (specifically `action`) to be used in statement form; this is to allow calling non-returning actions. In future we could use it for `if`, `match` statements too, so we don't have to have two parsing/lowering paths for them.


crates/aranya-policy-compiler/src/compile/lower.rs
```diff
@@ -1854,64 +1931,21 @@ impl CompileState<'_> {
...
-
-                    if action_def.arguments.len() != fc.arguments.len() {
-                        let note = format!(
-                            "call to `{}` has {} arguments, but it should have {}",
-                            fc.identifier.inner,
-                            fc.arguments.len(),
-                            action_def.arguments.len()
-                        );
-                        return Err(self.err(BadArgument(note, statement.span)));
+                    // Only `Never` expressions can be used as statements
```
chip-so, line 1943: Doesn't this make the `return` statement redundant as well?

chip-so (approved):

crates/aranya-policy-compiler/src/compile/lower.rs
```diff
@@ -1032,12 +1036,85 @@ impl CompileState<'_> {
...
+
+        // A non-returning action leaves nothing behind, so its call has type
+        // `never`. A fallible one evaluates to its declared `result`.
+        let return_type = match &action_def.return_type.inner {
+            TypeKind::Unit => VType {
+                inner: TypeKind::Never,
+                span,
+            },
+            _ => action_def.return_type.clone(),
+        };
```
jdygert-spok, line 1104: I get why you're doing this but it feels hacky to me. The action doesn't never return, it just returns without producing a value. So that doesn't match unit or never... Really the semantics we want are more like C's `void` return type. In a compiled language, unit and void can be the same thing fairly naturally, but for us I think it would take a lot of special casing...

I think this needs more thought and the simple thing to do would be to just duplicate the action statement and expression for now, like we did for other things.

crates/aranya-policy-compiler/src/compile/lower.rs
apetkov-so: Yes, `return`, `if`, and `match` statement kinds can all be removed and lowered to `Expression`s. But I would do that in a different ticket.

crates/aranya-policy-compiler/src/compile/lower.rs
apetkov-so: You mean drop `StmtKind::Expression` and duplicate the `ActionCall`? I was thinking we may want to expand Expression to de-duplicate the if, match, return statements which also have expression forms.

jdygert-spok: Yes. I agree that it would be nice to deduplicate those, but I think doing it "properly" will be complicated so we should just duplicate the two action forms for the purpose of solving #800 quickly, and work on the dedup separately.

PR #804, opened 2026-09-17
policy: replace seal/open with get_key

jdygert-spok:
Replaces policylang's seal/open blocks with a `get_key` block defined on base commands.

Policy writers can no longer customize seal and open within policy. Instead they decide how a command determines its verifying key. Init commands will generally take a key from the command fields, while other commands will take it from a fact. The signing key and author ID are now passed via a seal context which is currently selected by the policy store.

`get_key` blocks are defined on base commands instead of being repeated on each command. Base commands only support fields and the `get_key` block. A command will inherit the fields of its base command. Each command must define a base command, unless the `allow_baseless` flag is set on the compiler. This flag is mostly intended for internal use since many of our tests don't care about defining the `get_key` block.

The policy runner currently does not work.


crates/aranya-policy-compiler/src/validate.rs
```diff
@@ -24,12 +24,12 @@ pub fn validate(module: &Module) -> bool {
             LabelType::Action => {
                 tracer = tracer.add_analyzer(ActionAnalyzer::new());
             }
+            LabelType::GetKey => {
```
mooreoak: is this ok? We should at least describe why it is.

crates/aranya-policy-runner/src/lib.rs
```diff
@@ -345,6 +345,11 @@ impl PolicyRunner {
                         &mut perspective,
                         &mut sink,
                         ActionPlacement::OnGraph,
+                        // TODO
```
mooreoak: please explain todo

crates/aranya-runtime/src/testing/vm.rs
```diff
@@ -323,12 +343,39 @@ impl TestPolicyStore {
         let policy = VmPolicy::new(
             machine,
             eng,
-            vec![Box::from(TestFfiEnvelope {
-                device: DeviceId::random(Rng),
-            })],
+            // TODO(jdygert): Do we need to test an ffi here?
```
mooreoak: Another todo

mooreoak (changes requested):
Over all looks good but we need to replace TODO comments with either code or and explanation of why there is a TODO.

crates/aranya-runtime/src/testing/vm.rs
jdygert-spok: This was the only place in aranya-runtime we were testing ffi in `VmPolicy` but we also test it in the model and core example so I think it should be fine to not have any ffi here.

crates/aranya-policy-compiler/src/validate.rs
jdygert-spok: Using function analyzer now and added some tests.

crates/aranya-policy-runner/src/lib.rs
jdygert-spok: Expanded comment. Still need to figure out how to get the policy runner working again but it's not strictly required on this PR.

mooreoak (dismissed):

crates/aranya-policy-lang/src/lang/parse/policy.pest
```diff
@@ -399,7 +400,9 @@ struct_definition = { "struct" ~ identifier ~ struct_def }
 // enum Result { OK, Err }
 enum_definition = { "enum" ~ identifier ~ "{" ~ identifier ~ ("," ~ identifier)* ~ ","? ~ "}" }
 // command Foo { ... }
-command_definition = { ephemeral_modifier? ~ "command" ~ identifier ~ "{" ~ command_blocks+ ~ "}" }
+command_definition = { ephemeral_modifier? ~ "command" ~ identifier ~ base_usage? ~ "{" ~ command_blocks+ ~ "}" }
+// base command Foo { ... }
+base_command_definition = { "base" ~ "command" ~ identifier ~ "{" ~ base_command_blocks+ ~ "}" }
```
chip-so: So I hadn't noticed the `+` after `command_blocks` (and now `base_command_blocks`). Should that be there? Doesn't that mean a block could be defined twice as long as they're all in the right order?

crates/aranya-policy-lang/src/lang/parse/policy.pest
```diff
@@ -368,22 +368,23 @@ policy_block = { "policy" ~ statement_block }
...
 }
 
+get_key_block = { "get_key" ~ statement_block }
+// Sequence of base command blocks
+base_command_blocks = _{
+    fields_block?
+    ~ get_key_block
+}
+
+base_usage = { ":" ~ identifier }
```
chip-so: Not to bikeshed this, but for clarity I think I'd rather have a keyword here than the `:`. Only thing I can think of right now is `based_on`, though. :/

crates/aranya-policy-lang/src/lang/parse/policy.pest
jdygert-spok: Yeah looks like that was mistakenly kept when we changed it to a sequence.

crates/aranya-policy-lang/src/lang/parse/policy.pest
jdygert-spok: Using `with` instead based on decision from call.

mooreoak (approved):

crates/aranya-model/src/tests/basic-policy.md
```diff
@@ -17,6 +17,35 @@ some other mechanism.
...
+}
+
+base command BaseEphemeral {
+    get_key {
+        return match query Key[] {
+            Some(f) => Some(f.key)
+            None => None
+        }
+    }
+}
```
chip-so, line 47: Why are there two different base commands if the implementation is the same?

chip-so (approved):

crates/aranya-model/src/tests/basic-policy.md
jdygert-spok: For the future when the ephemeral one will be marked with a different "kind".

PR #805, opened 2026-09-17
crypto: use custom cache cell for thread safety

jdygert-spok:
The various key types in `aranya-crypto` used `core::cell::OnceCell` to cache computing their ID. That forced those keys to be `!Sync`. Replace this with a custom caching cell type that is `Sync`.

The cache cell is thread safe but never spins or blocks. Rather than waiting on another thread who is initializing the cache, we accept that we may call the initializer more than once. This is different from any cells provided by `std`, `oncecell`, or `spin`. I'm not aware of a popular crate's cell which has this behavior.


crates/aranya-crypto/src/util/cache.rs
```diff
@@ -0,0 +1,129 @@
...
+
+impl<T: Clone> CacheCell<T> {
+    /// Get the cached value or compute and potentially store it.
+    ///
+    /// Will call the init function "spuriously" on contention rather than waiting.
+    pub fn get_or_init(&self, init: impl FnOnce() -> T) -> T {
+        if let Some(val) = self.get() {
+            return val.clone();
+        }
+        self.init_cold(init)
```
mooreoak: I thought `cold(self.init_cold(init))` was the way this was supposed to be done. Did that change?

mooreoak (dismissed):

crates/aranya-crypto/src/util/cache.rs
jdygert-spok: The cold marker on `init_cold`'s definition handles this. But since `hint::cold_path()` is available now I am switching to that.

mooreoak (approved):

PR #807, opened 2026-09-18
policy-lang: support field init shorthand

jdygert-spok:
This lets you omit the expression when the field name and expression are the same identifier.


crates/aranya-policy-lang/src/lang/parse.rs
```diff
@@ -1093,7 +1093,13 @@ impl ChunkParser<'_> {
...
                 Rule::struct_literal_field => {
                     let identifier = pc.consume_ident(self)?;
-                    let expression = pc.consume_expression(self)?;
+                    let expression = match pc.consume_optional(Rule::expression) {
+                        Some(token) => self.parse_expression(token)?,
+                        None => Expression {
+                            inner: ExprKind::Identifier(identifier.clone()),
+                            span: identifier.span,
+                        },
+                    };
```
jdygert-spok, line 1102: This only needed a simple parser change by using the field identifier as the expression if the expression is missing.

mooreoak (dismissed):

apetkov-so (dismissed):
Huh... didn't even have to touch lower, or compile.

chip-so (approved):
Nice. Please also make a change in the policy book to describe this.

jdygert-spok:
> Nice. Please also make a change in the policy book to describe this.

https://github.com/aranya-project/aranya-docs/pull/214

mooreoak (approved):

PR #808, opened 2026-09-18
runtime: full size PolicyID and Arc FFI

jdygert-spok:
These are supporting changes needed for policy loading.

Use a full 32 byte ID for `PolicyId` and use `Arc<dyn FfiCallable>` so later they can be selected and cloned from a pool during loading based on schema.


mooreoak (dismissed):
It's hard to understand why this change is needed but I accept that it will be important.

apetkov-so (approved):

mooreoak (approved):

PR #810, opened 2026-09-18
policy-compiler: simplify error rendering

jdygert-spok:
Adds `trait SimpleError` which is just used to define error impls via a simpler API. These impls return an iterator of `(span, text)`, and the rendering is inferred by using the description or overloaded title, and the label iter as the snippets.


mooreoak (dismissed):
I don't see anything objectionable in this but I really don't understand this part of the code base so my review has low value.

mooreoak (approved):
This seems all reasonable but it is not part of the code I really know much about

PR #814, opened 2026-09-21
syncer: nostd syncer state machine v2

djohnson-spideroak:
Add a no_std compatible syncer state machine. This machine takes in time, storage on creation. During operation, the Syncer produces messages based on time passed into calls, and consumes syncer messages from peers. On consumption, it handles the lifecycle of peer subscriptions, in flight syncing, etc. 

Start with time.rs, config.rs, then move into slots.rs and schedule.rs, then the syncer mod.rs.

This also changes the behavior the the push sync SyncResponder and adds some functions to help the syncer drive it. Previously, each call to push even in the same session recomputed the commands to send. Now, repeatedly calling push on the same sync responder (while still in Send state) will attempt to work through the list of commands until up to date without recomputing each call. The behavior is unchanged if you are just constructing a SyncResponder, calling push once, then dropping.


crates/aranya-runtime/src/sync/syncer/config.rs
```diff
@@ -0,0 +1,231 @@
...
+        if remain_open.subsec_nanos() != 0 {
+            secs = secs.saturating_add(1);
+        }
+        secs.clamp(1, self.max_sub_duration.as_secs().max(1))
+    }
+
+    /// Delay before the next poll attempt after `failures` consecutive
+    /// failures (`failures >= 1`): exponential from `initial_backoff`,
+    /// capped at `max_backoff` and at the peer's own `interval`.
+    pub(crate) fn backoff(&self, failures: u8, interval: Option<Duration>) -> Duration {
```
djohnson-spideroak, line 127: We probably want a different backoff method, or configurable backoff via passed in function?

crates/aranya-runtime/src/sync/syncer/slots.rs
```diff
@@ -0,0 +1,438 @@
...
+            if min.is_none_or(|m| at < m) {
+                min = Some(at);
+            }
+        });
+        self.next_deadline = min;
+    }
+
+    /// Visits every armed timer as an `(instant, kind)` pair, in a fixed
+    /// role order so exact ties break deterministically.
+    pub(crate) fn for_each_deadline(&self, mut consider: impl FnMut(T, DueKind))
```
mooreoak, line 205: This should really document what `consider` is meant to do.

crates/aranya-runtime/src/sync/syncer/slots.rs
```diff
@@ -0,0 +1,438 @@
...
+        if let Some(at) = search_slots(&self.slots, graph_id, peer) {
+            return self.slots.get_mut(at).ok_or(OutOfSlots);
+        }
+        self.slots
+            .push(SyncSlot::new(graph_id, peer.clone()))
+            .map_err(|_| OutOfSlots)?;
+        self.slots.last_mut().ok_or(OutOfSlots)
+    }
+
+    fn remove(&mut self, graph_id: GraphId, peer: &A) {
```
mooreoak, line 353: Seems strange that this has no return value saying if it did it but I guess if it would not be used there is no reason to do that.

crates/aranya-runtime/src/sync/syncer/schedule.rs
```diff
@@ -0,0 +1,733 @@
...
+
+    /// Unregisters `peer` from polling. Returns whether it was registered.
+    pub(crate) fn remove_peer(&mut self, peer: &A, graph_id: GraphId) -> bool {
+        self.modify(graph_id, peer, |slot| slot.poll.take().is_some())
+            .unwrap_or(false)
+    }
+
+    /// Schedules an immediate poll, keeping any recurring interval. Returns
+    /// whether the peer is registered.
+    pub(crate) fn sync_now(&mut self, peer: &A, graph_id: GraphId, now: T) -> bool {
```
mooreoak, line 193: This seems to say sync now can only happen if you are already subscribed to that peer? I don't think that is correct if so. We should support a sync now that works even if you don't have that device as a current peer.

crates/aranya-runtime/src/sync/syncer/schedule.rs
```diff
@@ -0,0 +1,733 @@
...
+    /// rescheduled or cleared it. Timers whose subject died in the meantime
+    /// are cleared silently and the scan continues.
+    pub(crate) fn next_due(
+        &mut self,
+        now: T,
+        poll_gate: &Gate<A>,
+        push_gate: &Gate<A>,
+    ) -> Option<Due<A>> {
+        loop {
+            let (kind, graph_id, peer) = self.select(now, poll_gate, push_gate)?;
```
mooreoak, line 547: so the select returns `None` if there is nothing to select and this is another termination path? It's not easy to see how this loop terminates a `while let` loop on the select might read better.

crates/aranya-runtime/src/sync/syncer/mod.rs
```diff
@@ -0,0 +1,1311 @@
...
+}
+
+impl From<StorageError> for SyncerError {
+    fn from(err: StorageError) -> Self {
+        Self::Client(ClientError::StorageError(err))
+    }
+}
+
+/// Everything a call that touches storage needs, borrowed for the call.
+pub struct SyncContext<'a, PS, SP: StorageProvider, K, R, MS> {
```
mooreoak, line 215: if the things your are borrowing don't all have the same life time will only using `q'` be an issue? 

I feel like I have run in to that issue before but possibly it was a different situation or rust has been improved.

crates/aranya-runtime/src/sync/syncer/mod.rs
```diff
@@ -0,0 +1,1311 @@
...
+///         {
+///             transport.reply(peer, &out[..len])?;
+///         }
+///     }
+///
+///     // Outbound: send what is due; hand replies and failures back.
+///     while let Some(ob) = syncer.next_outbound(now, &mut cx, out)? {
+///         match transport.send(ob.peer(), &out[..ob.len()]) {
+///             Err(_) => syncer.complete(ob.token(), Outcome::Failed, now),
+///             Ok(()) if ob.expects_reply() => match transport.recv_reply(ob.peer(), reply) {
```
mooreoak, line 303: This is kind of confusing. I think what this means is that if we sent something successfully we will wait around for a reply before moving on to send more things?

crates/aranya-runtime/src/sync/syncer/mod.rs
```diff
@@ -0,0 +1,1311 @@
...
+///                 Ok(n) => { syncer.handle_reply(ob.token(), &reply[..n], now, &mut cx)?; }
```
mooreoak, line 304: Also not sure what this is doing.

crates/aranya-runtime/src/sync/syncer/mod.rs
```diff
@@ -0,0 +1,1311 @@
...
+///     while let Some((peer, frame)) = transport.try_recv() {
+///         if let Inbound::Reply { len } =
+///             syncer.handle_incoming(peer, frame, now, &mut cx, out)?
+///         {
+///             transport.reply(peer, &out[..len])?;
+///         }
+///     }
+///
+///     // Outbound: send what is due; hand replies and failures back.
+///     while let Some(ob) = syncer.next_outbound(now, &mut cx, out)? {
```
mooreoak, line 300: This loop and all the matches and sub matches are so compact it is hard to understand. I don't think it is a good teaching example.

crates/aranya-runtime/src/sync/syncer/mod.rs
```diff
@@ -0,0 +1,1311 @@
...
+        MS: Fn() -> Result<F, StorageError>,
+    {
+        self.settle(now, cx)?;
+        loop {
+            let (poll_gate, push_gate) = self.gates();
+            let Some(Due {
+                kind,
+                graph_id,
+                peer,
+            }) = self.schedule.next_due(now, &poll_gate, &push_gate)
```
mooreoak, line 635: it's not clear that there are not cases where this loops forever. If `next_due()`  were next due will return a stream of events forever? Some error case that keeps the protocol from advancing? It seems like it would be good to have a clear bound on this loop or it is hard to know when it terminates. At least a comment explaining why it terminates would be good.

crates/aranya-runtime/src/sync/syncer/mod.rs
```diff
@@ -0,0 +1,1311 @@
...
+        F: Spill,
+        MS: Fn() -> Result<F, StorageError>,
+    {
+        self.settle(now, cx)?;
+        match SyncIncoming::decode(data)? {
+            SyncIncoming::Poll(poll) => {
+                let graph_id = poll.graph_id();
+                let mut responder = SyncResponder::new();
+                responder.receive(poll)?;
+                let Self {
```
mooreoak, line 685: why not just access the fields with normal "." notation? Is this required for a partial barrow?

crates/aranya-runtime/src/sync/syncer/mod.rs
```diff
@@ -0,0 +1,1311 @@
...
+        out: &mut [u8],
+    ) -> Result<Inbound, SyncerError>
+    where
+        K: Sink<PS::Effect>,
+        R: Csprng,
+        F: Spill,
+        MS: Fn() -> Result<F, StorageError>,
+    {
+        self.settle(now, cx)?;
+        match SyncIncoming::decode(data)? {
```
mooreoak, line 680: this might be better as as dispatch in to a bunch of methods instead of huge match.

crates/aranya-runtime/src/sync/syncer/mod.rs
```diff
@@ -0,0 +1,1311 @@
...
+                let graph_id = push.graph_id();
+                if !self.schedule.has_active_push_request(graph_id, &peer) {
+                    return Ok(Inbound::Ignored(IgnoreReason::NotSubscribed));
+                }
+                if self
+                    .poll_session
+                    .as_ref()
+                    .is_some_and(|s| s.token.graph_id == graph_id)
+                {
+                    return Ok(Inbound::Ignored(IgnoreReason::Busy));
```
mooreoak, line 754: It feels like we should have a todo to queue this for later.

crates/aranya-runtime/src/sync/syncer/mod.rs
```diff
@@ -0,0 +1,1311 @@
...
+                }
+                Ok(Inbound::Handled)
+            }
+        }
+    }
+
+    /// Applies the reply to an outbound that
+    /// [`expects_reply`](Outbound::expects_reply).
+    ///
+    /// A poll reply's commands join the session's transaction; a full reply
```
mooreoak, line 821: We will want to consider what a "full reply" means with @jdygert-spok large message PR as full is not a count any more but "I could not fit the next message in the buffer" instead.

crates/aranya-runtime/src/sync/syncer/mod.rs
```diff
@@ -0,0 +1,1311 @@
...
+                Ok(match SubscribeResponse::decode(data)? {
+                    SubscribeResponse::Success => Inbound::Handled,
+                    SubscribeResponse::TooManySubscriptions => Inbound::SubscribeRejected,
+                })
+            }
+            OutboundKind::Unsubscribe
+            | OutboundKind::Push
+            | OutboundKind::HelloSubscribe
+            | OutboundKind::HelloUnsubscribe
+            | OutboundKind::Hello => Ok(Inbound::Ignored(IgnoreReason::StaleToken)),
```
mooreoak, line 873: This seems like the wrong reason? These are unexpected right?

crates/aranya-runtime/src/sync/syncer/mod.rs
```diff
@@ -0,0 +1,1311 @@
...
+            && !cmds.is_empty()
+        {
+            let added = cx.client.add_commands(
+                &mut session.trx,
+                cx.sink,
+                &cmds,
+                cx.buffers,
+                &cx.make_spill,
+            )?;
+            session.added = session.added.saturating_add(added);
```
mooreoak, line 1065: I fell like saturating add is not really correct here. The correct this is probably to have a type that supports a "over flow" state but that seems kind of like pointless work given that it really should never happen. We should probably just add a comment that it can't happen or use checked arithmetic and  an unreachable statment or return an error or something. It just feels wrong to hider the error.

crates/aranya-runtime/src/sync/syncer/mod.rs
```diff
@@ -0,0 +1,1311 @@
...
+            // Make the next round's advertised frontier include this one.
+            let storage = cx.client.provider().get_storage(graph_id)?;
+            session.trx.flush(storage)?;
```
mooreoak, line 1068: why do we need a flush here?

crates/aranya-runtime/src/sync/syncer/mod.rs
```diff
@@ -0,0 +1,1311 @@
...
+            follow_up = cmds.len() >= COMMAND_RESPONSE_MAX;
+        }
+        if follow_up {
+            self.schedule.poll_follow_up(graph_id, &peer, now);
+            return Ok(Inbound::Handled);
+        }
+
+        // Converged.
+        let Some(session) = self.poll_session.take() else {
+            return Ok(Inbound::Ignored(IgnoreReason::StaleToken));
```
mooreoak, line 1082: Is stale token really the right error here?

crates/aranya-runtime/src/sync/syncer/mod.rs
```diff
@@ -0,0 +1,1311 @@
...
+                self.schedule.poll_finished(graph_id, &peer, false, now);
+                self.finish_failed_session(session, now, cx)?;
+                Err(err.into())
+            }
+        }
+    }
+
+    /// Feeds one reply into the open session and decides whether it needs
+    /// another round or has converged.
+    fn apply_poll_reply<K, R, F, MS>(
```
mooreoak, line 1036: This function really should be broken up in to it's core flow putting the work in each branch in it's own helper functions. It is hard to fallow at it's current size.

crates/aranya-runtime/src/sync/syncer/mod.rs
```diff
@@ -0,0 +1,1311 @@
...
+            .push(out, cx.client.provider(), &mut cx.buffers.traversal)
+        {
+            Ok(len) => len,
+            Err(SyncError::Storage(StorageError::NoSuchStorage)) => 0,
+            Err(err) => return Err(err.into()),
+        };
+        if len == 0 {
+            return Ok(None);
+        }
+        let bytes = u64::try_from(len).unwrap_or(u64::MAX);
```
mooreoak, line 1178: It feels like this should be an error not u64::MAX

crates/aranya-runtime/src/sync/syncer/mod.rs
```diff
@@ -0,0 +1,1311 @@
...
+            // subscription instead of being sent.
+            self.schedule.remove_push_subscriber(&peer, graph_id);
+            return Ok(None);
+        }
+        self.schedule.record_push(graph_id, &peer, bytes);
+        // What was pushed counts as held by the peer from here on, so the
+        // next push (or the next session, when this one's segment list was
+        // capped) continues past it. A push that is lost stays hidden until
+        // the subscriber's next heads sample resets the cache.
+        if let Some(cache) = self.schedule.cache_mut(graph_id, &peer) {
```
mooreoak, line 1190: What happens in the else case here? We just don't store the cache? Dose that mean we will send the same commands agin later?

mooreoak (commented):
Over all looks reasonable. I do have some comments in line. I also think over all it would be better if functions were smaller and used fewer nested calls in matches ect with those pulled out to their own lines.

I also see a lot of stuff being put on the stack like `let len = SyncRequester::new(graph_id, cx.rng)...` How big is a SyncRequestor? I wonder if there is a stack analysis tool built in to LLVM that we could use to figure out if we are putting big objects on the stack.

Not your issue but this really shows that our current APIs are kind of inconsistent about how the return things. Some day it would be nice to document a standardize on a conventions. Ex "dose deleting something return a bool about weather it was present?" or the difference emit_push_unsubscribe and emit_hello_subscribe where one abstracts away the serialization and the other dose not.

